PR #56, opened 2019-02-13
AdminAuditLogConfig: Adding a new DSC Resource named AdminAuditLogConfig

RobBiddle:
Adding a new DSC Resource named **AdminAuditLogConfig** to allow for management of the Unified Audit Log Ingestion setting.

None

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/56)

### DIFF
--- a/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
+++ b/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
@@ -144,15 +144,15 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
             }
 
             It "Should throw error from the Get method" {
-                Get-TargetResource @testParams | Should Throw
+                Get-TargetResource @testParams | Should -Throw
             }
 
             It "Should throw error from the Sest method" {
-                Set-TargetResource @testParams | Should Throw
+                Set-TargetResource @testParams | Should -Throw
             }
 
             It "Should throw error from the Test method" {
-                Test-TargetResource @testParams | Should Throw
+                Test-TargetResource @testParams | Should -Throw
             }
         }
 
--- a/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
+++ b/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
@@ -128,7 +128,7 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
             }
 
         }
-
+<#
         Context -Name 'Test Fails When the Ensure Absent is specified' -Fixture {
             $testParams = @{
                 IsSingleInstance                = 'Yes'
@@ -155,7 +155,7 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
                 Test-TargetResource @testParams | Should -Throw
             }
         }
-
+#>
         Context -Name 'ReverseDSC Tests' -Fixture {
             $testParams = @{
                 IsSingleInstance                = 'Yes'
--- a/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
+++ b/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
@@ -144,15 +144,15 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
             }
 
             It 'Should throw error from the Get method' {
-                Get-TargetResource @testParams | Should Throw
+                { Get-TargetResource @testParams } | Should Throw "This resource cannot delete Managed Properties. Please make sure you set its Ensure value to Present."
             }
 
             It 'Should throw error from the Set method' {
-                Set-TargetResource @testParams | Should Throw
+                { Set-TargetResource @testParams } | Should Throw "This resource cannot delete Managed Properties. Please make sure you set its Ensure value to Present."
             }
 
             It 'Should throw error from the Test method' {
-                Test-TargetResource @testParams | Should Throw
+                { Test-TargetResource @testParams } | Should Throw "This resource cannot delete Managed Properties. Please make sure you set its Ensure value to Present."
             }
         }
 
--- a/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
+++ b/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
@@ -12,7 +12,7 @@ Import-Module -Name (Join-Path -Path $PSScriptRoot `
                                 -Resolve)
 
 $Global:DscHelper = New-O365DscUnitTestHelper -StubModule $CmdletModule `
-                                                -DscResource "AdminAuditLogConfig"
+                                                -DscResource "O365AdminAuditLogConfig"
 
 Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
     InModuleScope -ModuleName $Global:DscHelper.ModuleName -ScriptBlock {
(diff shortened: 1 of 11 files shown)


RobBiddle:
I fixed the Markdown failure by adding another blank line.  It looks like the only remaining issue is with the tests.  The test is unable to use **Get-AdminAuditLogConfig** which is part of the Exchange Online cmdlets.

RobBiddle:
Yay, passing all required tests!
Working on some of the Style Guideline warnings...

RobBiddle:
@NikCharlebois I think this PR is ready for review.  Anything else I need to do?

NikCharlebois:
No sir. Let me take it from here. I will do the code review later today. Thanks for your help

NikCharlebois:
@RobBiddle Great job with this PR. I did review the files. Only some very minor comments. Thanks

RobBiddle:
@NikCharlebois Thank you for the quick and thorough review!  I'll get to work on the requested changes.

In regards to a parameter named **Ensure**, I was originally going to include that, since it's a DSC staple, however it seemed confusing and superfluous since there is no way to delete the AdminAuditLogConfig configuration.

The options I considered are:
1. Include **Ensure** parameter, doing nothing if value is 'Absent'
2. Include **Ensure** parameter, set default values if value 'Absent' 
    - (which would be $false for UnifiedAuditLogIngestionEnabled)
3. Include **Ensure** parameter, and only allow a value of 'Present' (as you've suggested)
4. Not include the **Ensure** parameter

1&2 maintain consistency with built-in DSC Resources (the only advantage)
https://docs.microsoft.com/en-us/powershell/dsc/resources/authoringresourcemof

> Including a property called Ensure with values Present and Absent in your resource is recommended as a way to maintain a consistent style with built-in DSC resources.

1,2&3 would **_increase_** the number of parameters in a configuration, unless **Ensure** was used as the key in place of **IsSingleInstance**

Both 3 & 4 would result in an error if someone tried to generate a configuration using `Ensure = 'Absent'` and would not be consistent with built-in DSC resources.

Option 3 doesn't add any functionality.

4 would also result in an error if someone tried to generate a configuration using `Ensure = 'Present'`, and would not be consistent with built-in DSC resources.
_________________________

I'm of the opinion that 1,2 or 4 are the best options.  

One question I have; I don't know if deeper in either the DSC MOF generation code or the LCM code the parameter **IsSingleInstance** is given any special treatment, or if it's just a standard name given to a required and constrained key property in a lot of official resources.

The only reason to include **Ensure** is to maintain consistency, and if that is the purpose then I think option 2 is best.

Option 1 would be the _safest_ since no changes are made, however I also don't think it's good to have a configuration management system leave things in an ambiguous state.

Option 2 should make sense to authors of configurations (i.e. 'Absent' == 'Disabled'/$false).  So the only reason not to do it is because it increases the number of parameters in a configuration, which as stated above, could be remedied by replacing **IsSingleInstance** with **Ensure**.  Again this is under the assumption that **IsSingleInstance** holds no special status and that **Ensure** could be used sufficiently to prevent configuration conflicts (I'm guessing this is the case).

The antithesis to the entire discussion is that there are certainly examples of _**official**_ Microsoft DSC Resources (not sure about any examples in built-in resources) which do not include an **Ensure** parameter: https://github.com/PowerShell/ComputerManagementDsc/blob/dev/Modules/ComputerManagementDsc/DSCResources/MSFT_TimeZone/MSFT_TimeZone.schema.mof

/CanOfWorms

NikCharlebois:
IsSingleInstance is just a naming convention we established. Since you cannot have the same Resource/Key combination more than once in a config, it ensures an error is thrown the moment you have more than one instance of it when compiling the MOF.

In my opinion, Option 2 would not be viable because we make an assumption on behalf of the user. What happens if we eventually expand the scope of this resource to include other parameters (say number of audit logs retention days as an example). "Ensure" would then only default certain parameters.

I agree with you that option 1 but as you say, it would be ambiguous. The same thing would happen with no regards to Ensure being Absent or Present and we don't want that.

I am not sure I agree that Option 3 would add additional parameters to configs. Users can still omit it entirely and that would default to Present which is the standard.

RobBiddle:
Expanding the scope of the resource is always a _possibility_, however as of right now `Set-AdminAuditLogConfig` only has 4 parameters which apply to **Exchange Online**, and 3 of those are _**reserved for internal Microsoft use**_, so I'm not sure how much of a concern that is going to be. (unless you have some insider knowledge

If proceeding with Option 3, what are your thoughts on keeping **IsSingleInstance**? 
Am I correct in thinking that in a schema with a single key property, that property has to correspond to a mandatory parameter in the resource?  If that's not the case then it doesn't matter, but if it is then maybe I should make $UnifiedAuditLogIngestionEnabled the key property.

RobBiddle:
@NikCharlebois I think I need to make $UnifiedAuditLogIngestionEnabled a Mandatory Parameter as well, otherwise it's the same issue as $Ensure, making an assumption on behalf of the user.

NikCharlebois:
@RobBiddle I don't think we are making any additional assumption in that case. All resources that implement the IsSingleInstance pattern only define that one Key. Take a look at this example:
https://github.com/PowerShell/SharePointDsc/blob/dev/Modules/SharePointDsc/DSCResources/MSFT_SPFarm/MSFT_SPFarm.psm1

RobBiddle:
@NikCharlebois I think it does make assumptions, as-is.

If **UnifiedAuditLogIngestionEnabled** is not mandatory, and has a default value of 'Disabled' then it would be possible for someone to write a confusing configuration like this:

```PowerShell
        O365AdminAuditLogConfig EnableAuditLog {
            IsSingleInstance                = 'Yes'
            Ensure                          = 'Present'
            GlobalAdminAccount              = $GlobalAdminAccount
        }
```

The config author may think this is enabling the AdminAuditLog (Which, unlike on-prem Exchange, is _Not_ configurable in Exhange Online).  Instead, it would end up disabling Unified Audit Log Ingestion.

NikCharlebois:
I see what you mean, but it should be made mandatory, yes. Just not a key of the resource

RobBiddle:
@NikCharlebois 
It doesn't look like a Test Case for _**Ensure = 'Absent'**_ is going to work.

I have a test in place, expecting a `throw` from `Get-TargetResource`
https://github.com/RobBiddle/Office365DSC/blob/e2552248291d295281da29289328b246fe7405f2/Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1#L132#L157

I have the `throw` statement in `Get-TargetResource`
https://github.com/RobBiddle/Office365DSC/blob/e2552248291d295281da29289328b246fe7405f2/Modules/Office365DSC/DSCResources/MSFT_O365AdminAuditLogConfig/MSFT_O365AdminAuditLogConfig.psm1#L25#L28

However, it looks like the ValidateSet property ...
```PowerShell
        [Parameter()]
        [ValidateSet("Present")]
        [string]$Ensure = "Present",
```
is causing the function to exit prior to my `throw` statement:
https://ci.appveyor.com/project/NikCharlebois/office365dsc#L187

RobBiddle:
I could mock the calls to `Get-TargetResource`, but that kind of seems like it defeats the purpose of the test. :-p

NikCharlebois:
@RobBiddle for your commented out test, we don't have a choice but to Add the value Absent to the ValidateSet. I know it sounds counter intuitive, but those are the default values anyway. Let the Get then trap the error if Value is Absent

Tests/Unit/Office365DSC/Office365DSC.O365AdminAuditLogConfig.Tests.ps1
```diff
@@ -0,0 +1,180 @@
...
+            }
+
+            Mock -CommandName Get-AdminAuditLogConfig -MockWith {
+                return @{
+                    UnifiedAuditLogIngestionEnabled = $true
+                }
+            }
+
+            It 'Should throw error from the Get method' {
+                Get-TargetResource @testParams | Should -Throw
```
NikCharlebois: Use a syntax like the following instead:
{ Get-TargetResource @testParams } | Should Throw "This resource cannot delete Managed Properties. Please make sure you set its Ensure value to Present."

RobBiddle: Ok, I'll try that next

PR #65, opened 2019-03-07
Using ScriptsToProcess instead of RequiredModules in manifest.  Impor…

RobBiddle:
Imports required modules via **ScriptsToProcess** instead of **RequiredModules**.
This was done so that Verbose output could be suppressed during module import, which greatly reduces the noise seen during DSC LCM runs using the **-Verbose** switch.

None

### DIFF
--- a/Modules/Office365DSC/Office365DSCRequiredModules.ps1
+++ b/Modules/Office365DSC/Office365DSCRequiredModules.ps1
@@ -1,0 +1,5 @@
+Import-Module -Name 'Microsoft.Online.SharePoint.PowerShell' -Version '16.0.8316.0' -Force -Verbose:$false -Global
+Import-Module -Name 'MicrosoftTeams' -Version '0.9.6' -Force -Verbose:$false -Global
+Import-Module -Name 'MSOnline' -Version '1.1.183.17' -Force -Verbose:$false -Global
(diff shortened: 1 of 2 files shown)


Modules/Office365DSC/Office365DSCRequiredModules.ps1
```diff
@@ -0,0 +1,5 @@
+Import-Module -Name 'Microsoft.Online.SharePoint.PowerShell' -Version '16.0.8316.0' -Force -Verbose:$false -Global
```
NikCharlebois: Not sure what the advantage of this approach would be. By using RequiredModules instead, it forces PowerShellGet to build the dependencies tree and automatically download all dependencies from the gallery.

RobBiddle: Trying to avoid verbose messages from module imports from showing when running `Start-DscConfiguration -Verbose`

RobBiddle: I opened a UserVoice request for this: https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/37075921-requiredmodules-statement-in-module-manifests-shou

NikCharlebois:
@RobBiddle I will be closing this PR. I believe RequiredModules is still required. I do support the UserVoice issue you raised, and upvoted it.

PR #82, opened 2019-04-02
Additional properties for SPOSite

ThorstenLoeschmann:


ThorstenLoeschmann:
Found a few things I will have to fix. Will re-submit.

PR #89, opened 2019-04-12
Validating Examples

NikCharlebois:
Getting App Veyor to validate the "compilationess" of our examples

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/89)


NikCharlebois:
For the life of me, I can't figure out this error. Not sure how the validation is done. Even if I delete the first Example completely (e.g. EXOMailboxSettings\1-ConfigureMailboxSettings.ps1) then it's the second one that gets the error, which leads me to believe it is not related to the resource or the example itself. I will investigate further tomorow, but if you've seen this before I could use a hand.

NikCharlebois:
@ykuijs have you ever seen this before?

NikCharlebois:
@ykuijs  Since you are the mastermind behind this PR, you get the to be the one reviewing it :)

NikCharlebois:
Had to fix some merge issues, @ykuijs to review whenever you get a chance. Thanks

ykuijs:
Just a few minor comments

PR #105, opened 2019-05-10
Speed up OneDrive

### LINKED ISSUE #102: Speed up ODSettings
ODSettings Set-TargetResource method is extremely slow to complete. Instead of calling the Set-SPOTenant cmdlet multiple times, we should build an hash of parameters to set and only call it once.

N/A

Change logic to build a param hash and only call Set-SPOTenant once.

N/A

N/A

Alpha

desmay:
Reduced calls to Set-SPOTenant
    - Fixes #102

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/105)


NikCharlebois:
@desmay Please update your fork/branch with latest changes. It will help resolve some of the tests that fail.

PR #118, opened 2019-06-18
SCRetentionComplianceRule

NikCharlebois:
Initial commit for SCRetentionComplianceRule

N/A


NikCharlebois:
@ykuijs Need your eagle eyes once more. This once is in all aspect similar to the previous one, but with different parameters

Modules/Office365DSC/DSCResources/MSFT_SCRetentionComplianceRule/MSFT_SCRetentionComplianceRule.psm1
```diff
@@ -0,0 +1,283 @@
...
+            else
+            {
+                $result += @{
+                    $KeyName = $PSBoundParameters[$KeyName]
+                }
+            }
+        }
+
+        Write-Verbose -Message "Found RetentionComplianceRule $($Name)"
+        Write-Verbose -Message "Get-TargetResource Result: `n $(Convert-O365DscHashtableToString -Hashtable $result)"
```
ykuijs: The Test method also outputs this info. Since that also calls the Get method, this info is outputted to the screen twice. Might be good to call this just once.

Modules/Office365DSC/Examples/Resources/SCRetentionComplianceRule/1-AddingNewRetentionComplianceRule.ps1
```diff
@@ -0,0 +1,28 @@
...
+        $credsGlobalAdmin
+    )
+    Import-DscResource -ModuleName Office365DSC
+
+    node localhost
+    {
+        SCRetentionComplianceRule DemoRule
+        {
+            Name                      = "DemoRule2"
+            Policy                    = "Nik"
```
ykuijs: Maybe set this to Contoso, instead of Nik :-p

Modules/Office365DSC/DSCResources/MSFT_SCRetentionComplianceRule/MSFT_SCRetentionComplianceRule.psm1
NikCharlebois: Kept it only in the Get, this way we also target the ReverseDSC scenario.

PR #120, opened 2019-06-20
General fixes

### LINKED ISSUE #86: Need to Build a Robust Logging Engine
### Description
We need to ensure that the Extraction component (ReverseDSC) never stops whenever there are errors detected. We do need however to capture all of these errors, in a very structured way. Potential known errors should have an Error Code, and Stack Trace should be captured as much as possible. Upon finalizing the extraction, the error log should be stored wherever the client decides to store the extraction results and the client should be warned in the console that errors where encountered during the extraction.

We need to make the troubelshooting process as smooth as possible. A client does an extraction, something goes wrong, they get notified and they have a log file they can send us for further troubleshooting.

NikCharlebois:
Adds support for the new Log Engine.

- Fixes #86


Modules/Office365DSC/DSCResources/MSFT_O365User/MSFT_O365User.psm1
```diff
@@ -171,7 +171,7 @@ function Get-TargetResource
     }
     catch
     {
-        Write-Verbose -Message "The specified User doesn't already exist."
+        $Message = "The specified User {$UserPrincipalName} doesn't already exist."
```
ykuijs: Nothing is done with this Message variable.

Modules/Office365DSC/Modules/O365DSCLogEngine.psm1
```diff
@@ -0,0 +1,40 @@
...
+        $Error,
+
+        [Parameter()]
+        [System.String]
+        $Message
+    )
+
+    # Get the current folder we are executing the script from;
+    $ExecPath = $SCRIPT:MyInvocation.MyCommand.Path
+    $LastSlash = $ExecPath.ToString().LastIndexOf("\")
```
ykuijs: Isn't "Split-Path -Path $ExecPath -Parent" easier in this case?

Modules/Office365DSC/DSCResources/MSFT_O365User/MSFT_O365User.psm1
NikCharlebois: Good Catch. Fixed

Modules/Office365DSC/Modules/O365DSCLogEngine.psm1
NikCharlebois: You are right, and actually, because I am trying to save the file within the current execution folder, I don't even need to determine this. Out-File will save in current folder by default. So I completely removed these lines.

PR #139, opened 2019-07-09
Fixed ValueMap for Ensure

### LINKED ISSUE #94: Check the Ensure ValueMap for all resources
Most resources have the Ensure parameter. Some only allow the value Present, but the ValueMap still contains Absent. Should Absent be accepted if the description states that only Present is allowed?

NikCharlebois:
For resources that threw an error if Ensure = "Absent" was set, we simply removed Absent from the value map.

- Fixes #94

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/139)


NikCharlebois:
@ykuijs All yours to review sir!

Modules/Office365DSC/DSCResources/MSFT_O365AdminAuditLogConfig/MSFT_O365AdminAuditLogConfig.psm1
```diff
@@ -10,8 +10,9 @@ function Get-TargetResource
         $IsSingleInstance,
 
         [Parameter()]
-        [ValidateSet('Present', 'Absent')]
-        [string]$Ensure = 'Present',
+        [ValidateSet('Present')]
+        [string]
```
ykuijs: In the Set method you updated the type to System.String. Lets do it here and in the Test method as well to keep it consistent.

Modules/Office365DSC/DSCResources/MSFT_O365AdminAuditLogConfig/MSFT_O365AdminAuditLogConfig.psm1
```diff
@@ -140,8 +132,9 @@ function Test-TargetResource
...
```
ykuijs: Same as before

NikCharlebois:
@ykuijs Requested Changes Done

PR #155, opened 2019-07-19
BREAKING CHANGES: Removal of CentralAdminUrl in ODSP resources

### LINKED ISSUE #151: Remove CentralAdminUrl as a parameter from all SPOx Resources
We are now determining the CentralAdminUrl for the ReverseDSC path using the GlobalAdminAccount credentials. Since this is already a required parameter for all SPOx resources, we should be able to dynamically determine that URL and therefore no longer require CentralAdminUrl as a parameter.

NikCharlebois:
Removes the mandatory CentralAdminUrl parameter from all ODSP based Resources.

Fixes #151


NikCharlebois:
Hold off before any Reviews, I know the tests will fail. That is to be expected at this stage.

NikCharlebois:
@ykuijs Can I get a code review when you get a chance? Breaking changes which remove the need to pass in CentralAdminUrl in the SPO resources. Thanks mate!

PR #200, opened 2019-09-17
SPOSiteAuditSettings

NikCharlebois:
New resource SPOSiteAuditSettings

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/200)

### DIFF
--- a/Modules/Office365DSC/DSCResources/MSFT_SPOSiteAuditSettings/MSFT_SPOSiteAuditSettings.psm1
+++ b/Modules/Office365DSC/DSCResources/MSFT_SPOSiteAuditSettings/MSFT_SPOSiteAuditSettings.psm1
@@ -1,3 +1,4 @@
+@@ -0,0 +1,163 @@
 function Get-TargetResource
 {
     [CmdletBinding()]
@@ -11,11 +12,6 @@ function Get-TargetResource
         [ValidateSet('All','None')]
         [System.String]
         $AuditFlags,
-
-        [Parameter()]
-        [ValidateSet("Present")]
-        [System.String]
-        $Ensure = "Present",
 
         [Parameter(Mandatory = $true)]
         [System.Management.Automation.PSCredential]
@@ -40,7 +36,6 @@ function Get-TargetResource
             Url                = $Url
             AuditFlags         = $auditSettings.AuditFlags.ToString()
             GlobalAdminAccount = $GlobalAdminAccount
-            Ensure             ="Present"
         }
     }
     catch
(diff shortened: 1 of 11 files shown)


NikCharlebois:
@ykuijs for review.

Modules/Office365DSC/DSCResources/MSFT_SPOSiteAuditSettings/MSFT_SPOSiteAuditSettings.psm1
```diff
@@ -0,0 +1,180 @@
...
+
+        [Parameter(Mandatory = $true)]
+        [ValidateSet('All','None')]
+        [System.String]
+        $AuditFlags,
+
+        [Parameter()]
+        [ValidateSet("Present")]
+        [System.String]
+        $Ensure = "Present",
```
ykuijs: Since Present is the only allowed option, wouldn't it be a better idea to skip this parameter and return Null as AuditFlags? Now Ensure=Absent does not mean the setting is absent, but that something has gone wrong.

Modules/Office365DSC/DSCResources/MSFT_SPOSiteAuditSettings/MSFT_SPOSiteAuditSettings.psm1
```diff
@@ -0,0 +1,180 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $SiteUrl,
```
ykuijs: SPOSite is using the parameter Url. Isn't it a good idea to use a consistent parameter name?

NikCharlebois: Done

Modules/Office365DSC/DSCResources/MSFT_SPOSiteAuditSettings/MSFT_SPOSiteAuditSettings.psm1
NikCharlebois: Good call. Now returning AuditFlags = $null if something goes wrong instead of Absent.

ykuijs: Shouldn't the ensure parameter be removed altogether? Right now it is not being used at all.

NikCharlebois:
Agreed to remove the Ensure parameter from the resource since it isn't used.

PR #221, opened 2019-10-25
Update MSFT_EXOAntiPhishPolicy.psm1

NikCharlebois:
Updating the Get-TargetResource function of several EXO based resources to ensure that they only accept the Key parameters and that they return all associated properties. This is to fix a situation where ReverseDSC only returns a mirror of the PSBoundParameters for these resources.

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/221)


NikCharlebois:
@ykuijs For your review mate

NikCharlebois:
@ykuijs yes they need to unfortunately, since our Integration tenant doesn't currently support ATP.

PR #228, opened 2019-11-03
O365DSCReverse: Add missing EXOAcceptedDomain, EXOAntiPhishPolicy, EXOAntiPhishRule

reijoh:

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/228)

### DIFF
--- a/Modules/Office365DSC/Modules/O365DSCReverse.psm1
+++ b/Modules/Office365DSC/Modules/O365DSCReverse.psm1
@@ -141,14 +141,12 @@ function Start-O365ConfigurationExtract
         $AllComponents -or ($null -ne $Workloads -and $Workloads.Contains("EXO")))
     {
         Write-Information "Extracting EXOAcceptedDomain..."
-        try
-        {
-            $ModulePath = Join-Path -Path $PSScriptRoot `
-                                    -ChildPath "..\DSCResources\MSFT_EXOAcceptedDomain\MSFT_EXOAcceptedDomain.psm1" `
-                                    -Resolve
+        $ModulePath = Join-Path -Path $PSScriptRoot `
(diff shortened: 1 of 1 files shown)


NikCharlebois:
@reijoh thanks for the PR. It appears we may have some overlapping changes between:
https://github.com/microsoft/Office365DSC/pull/221

And the PR you just submitted. What I would like to ask of you is that you give the PR link above a try and see if it actually does the Extract of the AntiPhishPolicy, Rule and ATP as you would expect it to. We are in the process of refactoring the logic within our Export-TargetResource so that all the extraction logic is done in there, and that O365DSC.Reverse becomes just a class to collect results from the various resources' methods. Thanks

reijoh:
> 
> 
> @reijoh thanks for the PR. It appears we may have some overlapping changes between:
> #221
> 
> And the PR you just submitted. What I would like to ask of you is that you give the PR link above a try and see if it actually does the Extract of the AntiPhishPolicy, Rule and ATP as you would expect it to. We are in the process of refactoring the logic within our Export-TargetResource so that all the extraction logic is done in there, and that O365DSC.Reverse becomes just a class to collect results from the various resources' methods. Thanks

I could not find EXOAcceptedDomain in #221. I have removed EXOAntiPhishPolicy and EXOAntiPhishRule and updated EXOAcceptedDomain logic to what you have done for EXOAntiPhish. I will do some testing tomorrow.

reijoh:
I justed tested #221, and can confirm that extraction of chckEXOAntiPhishPolicy works. Also, many components have more parameters now.
The change I did with EXOAcceptedDomain do not work, it asks for Identity and I can see that a change must be made in Export-TargetResource of MSFT_EXOAcceptedDomain.psm1 before it will work. I can do the change if you do not do it yourself, but will wait for #221 to be merged first.

This PR should be rejected.

PR #241, opened 2019-11-18
SCFilePlanPropertyDepartment

NikCharlebois:
Initial Release

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/241)


NikCharlebois:
@ykuijs For review. Fairly simple resource, and almost an exact copy of the other PRs.

NikCharlebois:
Actually, an update regarding the | Out-Null it appears that the Integration pipeline requires it to be present, otherwise there is some verbose being thrown and it somehow interferes with the return value. I will update all 5 PRs.

PR #299, opened 2019-12-09
Fix Issue 298

### LINKED ISSUE #298: SCComplianceTag ReviewerEmail should be String[]
Need to change the parameter type for ReviewerEmail to become an array since it is possible to have multiple reviewers.

NikCharlebois:
Fixes param type for ReviewerEmail in SCComplianceTag. Note that this is not a breaking change since individual string values will continue to work just fine.

 - Fixes #298

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/299)


NikCharlebois:
@ykuijs for your review 5 (fairly small change)

PR #339, opened 2020-01-20
Fix Issue 312

### LINKED ISSUE #312: EXOAcceptedDomain extraction problem with the match operator used
I added some custom domains to my tenant and there seems to be an issue with the extraction process here.
I added a domain and i also added a subdomain. this causes problems for the extraction process since the match operator is used(MSFT_EXOAcceptedDomain.ps1:50)
The bug could also manifest if I had two custom domains ie. mydomain.com and  something-mydomain.com
Get-AcceptedDomain has an Identity paramer, so this could be used directly to single out the accepted domain instead of filtering the retrieved accepted domains list.

```powershell
 EXOAcceptedDomain e5e3b23b-3f5a-4c99-aa1f-68c5c1e27ff0
        {
            DomainType           = @("Authoritative","Authoritative");
            GlobalAdminAccount   = $Credsglobaladmin;
            Identity             = @("test1env.testing.io","testing.io");
            MatchSubDomains      = @(FalseFalse);
            OutboundOnly         = @(FalseFalse);
        }
```

dev

NikCharlebois:
Fixes an issue where the Accepted Domains were not properly extracted if more than one custom domain matched similar patterns.

- Fixes #312

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/339)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+* EXOAcceptedDomain
+  * Fixed an issue where the domains were not properly extracted
+    if multiple domain matches a similar pattern;
 * EXOHostedOutboundSpamFilterPolicy
   * Fixed an error where the resource was not being extracted via
     the Graphical User Interface;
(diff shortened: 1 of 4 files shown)


NikCharlebois:
@desmay can you review when you get a chance? Very simple PR. Thanks

PR #354, opened 2020-02-03
Major Improvements

NikCharlebois:
#### Pull Request (PR) description
- Refactoring of SPOSite;
- New MSCloudLoginAssistant Version;
- Complete Removal of MSOnline;
- Partial removal of SPO Management Shell

#### This Pull Request (PR) fixes the following issues
N/A

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/354)

PR #375, opened 2020-02-18
ODSettings: Removed SPO module

vinkobedek:
I think that OD settings are the last thing stopping the removal of the SharePoint Online PowerShell module, at least when the #354 PR is completed.
This PR rewrites the OD settings with CSOM code since I could not find anything in PnP that would be equivalent to the existing resource. Maybe I'm wrong and did not look hard enough, but this is the solution that I came up with.

I also added a utility function that was found floating around the web, Load-CSOMProperties which lets us select the properties that we need when writing a CSOM query. In the end it was not required since all the required properties on the tenant object are loaded by default but it may come in handy in the future.

Execute-CSOMQueryRetry just calls some PnP extension method to keep the same retry behavior that is within PnP.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/375)

### DIFF
--- a/Modules/Office365DSC/DSCResources/MSFT_ODSettings/MSFT_ODSettings.psm1
+++ b/Modules/Office365DSC/DSCResources/MSFT_ODSettings/MSFT_ODSettings.psm1
@@ -138,7 +138,7 @@ function Get-TargetResource
             $FixedExcludedFileExtensions = @()
         }
 
-        $FixedAllowedDomainList = $tenant.AllowedDomainListForSyncClient
+        $FixedAllowedDomainList = $tenant.AllowedDomainListForSyncClient | ForEach-Object { $_.ToString()}
         if ($FixedAllowedDomainList.Count -eq 0 -or
             ($FixedAllowedDomainList.Count -eq 1 -and $FixedAllowedDomainList[0] -eq ""))
         {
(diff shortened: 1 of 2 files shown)


vinkobedek:
Sorry, got ahead of myself :). we only needed the extraction part so i forgot to implement Test and Set with PnP. Will update sometime this week.

vinkobedek:
@NikCharlebois  I've rewritten the Set-TargetResource method but not sure on how to approach the tests. Previously there was another layer between CSOM and this code which could be stubed out, and now I'm attacking it directly since there is no _tenantsyncclientrestriction_ cmdlets in PnP. Any thoughts are appreciated. Also if you find that there is a better way to  do this for ODSettings I can abandon this PR.
I guess a cleaner (and lengthier) solution might be to create a PR on the PnPPowershell repo that implements the missing cmdlets.

NikCharlebois:
@desmay since you are the resource's owner, can I ask you to have a look at the proposed changes when you get a chance?

desmay:
@NikCharlebois I'll take a look at this tomorrow

desmay:
@NikCharlebois I agree with @vinkobedek I searched and couldn't find a equivalent way in PNP directly. I think we go forward with changes and if logic is added later to PNP we could revise code in project

NikCharlebois:
@vinkobedek @desmay looping in @VesaJuvonen . Vesa and I had a chat a few weeks back. We would need to get this into PnP first, this would be our best option. It shouldn't be much of a challenge to introduce this tenantsyncclientrestriction into PnP.

vinkobedek:
@NikCharlebois, @desmay so just to confirm, I can safely abandon this PR and we wait until it's implemented in PnP?
We will use the current code in this PR in our fork until then.
If it's just a 1:1 mapping between the old SPO cmdlet and the new PnP one, I can create the PnP pull request for this.

PR #416, opened 2020-03-12
ODSettings - Fixing issue where Set-SPOTenant was being called with Ensure

NikCharlebois:
The Set-TargetResource wasn't removing the Ensure parameters from the set before attempting to call into Set-SPOTenant.

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/416)


NikCharlebois:
@desmay Can you review since you are the resource's owner. It should be a VERY quick Code Review (CR)

PR #419, opened 2020-03-18
Support for unattended usage i.e. from within a windows service

vinkobedek:
@NikCharlebois , @brianlala  here's 2/2 PR from the call that we had earlier. As mentioned, this PR is a showcase of how we achieved Office365 extraction from a windows service when the tenant/account has MFA enabled. It is related to the following [MSCloudLogin PR](https://github.com/microsoft/MSCloudLoginAssistant/pull/45).
While not meant to be merged maybe it can give you some ideas.

The main changes that are related to MFA and unattended extraction are:
1. a custom Start-DSCInitializedJob which was inspired by the Invoke-DSCCommand.
   This copies everything required for Test-MSCloudLogin to continue to function inside the new job
2. logic that detects if a specific platform failed and if the next resource uses that platform and if it should be tried

In this PR there are also some random fixes and improvements that we might at some point create separate PRs:
- EXOMalwareFilterPolicy problem with unknown values
- ExoMalwareFilterRule problem with arrays, wrongly typed properties
- O365Group changed the AzureAD command used because it did not return the correct set of groups
- added $RawInputObject parameter in Get-TargetResource for some critically slow resources so that we can propagate already retrieved objects to avoid a large number of extra requests
- ODSettings handled directly with CSOM, this was shown already in another PR
- SPOApp, removed downloading of app packages since we have no need for them inside SysKit Trace, at least currently
- SearchManagedproperty and SPOSearchResultSource ensured that its connecting to the admin site
- removed SPOSiteAuditSettings since its more or less obsolete
- TeamChannel and TeamUser, added TeamMailNickName to ensure uniqueness. Groups/Teams can have duplicate names
- Removed Get-SPOAdministrationUrl and used the implementation from MSCloudLoginAssistant since it is more correct
- added string escaping for Property bag value since the about me field can contain html and quotes

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/office365dsc/419)


vinkobedek:
Closing this one since it was created based on the Dev branch and do not want to push changes to this PR. will recreate.

PR #596, opened 2020-05-14
Fix Issue 595 - Test-Microsoft365DSCParameterState was Comparing same array of object

NikCharlebois:
Test-Microsoft365DSCParameterState was always comparing the same array of objects, always resulting in the method returning true for Object arrays;

- Fixes 595

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/596)


NikCharlebois:
@ykuijs ready for review

PR #724, opened 2020-07-31
Fixes Remove-EmptyValue for edge case when dealing with GenericLists

PrzemyslawKlys:

Fixes an edge case where if entries are generic list with 1 element it would be removed anyways. I reverted it back to verifying for null and empty string rather then using [string]::EmptyOrNull which was giving true in that case.

No reported issues - edge case found in other project

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/724)


PrzemyslawKlys:
Sorry about that. Did it via GitHub interface and copy-paste (I use different formatting in my projects).

PR #800, opened 2020-09-24
MSFT_AADAdministrativeUnit - Created Inital Assets

blevinscm:

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/800)


NikCharlebois (changes requested):
Have a look at the comment WRT the missing Set-TargetResource. Also, we will need to add Pester Unit tests. More than happy to help you with those.

> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/800#-:-MI0lpc0Av-J0E7NwklO:b-okechp)* status: all files reviewed, 3 unresolved discussions (waiting on @blevinscm)

___
*[Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.psm1, line 80 at r2](https://reviewable.io/reviews/microsoft/microsoft365dsc/800#-MI0lMWPAqmqH-08uRAl:-MI0lMWPAqmqH-08uRAm:b-n12wnp) ([raw file](https://github.com/microsoft/microsoft365dsc/blob/ef96501f52649382bbbde50c5e524e6bed1b5221/Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.psm1#L80)):*
> ```PowerShell
>     }
> }
> ```

Somehow you are completely missing the Set-TargetResource method from the file
___
*[Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.psm1, line 123 at r2](https://reviewable.io/reviews/microsoft/microsoft365dsc/800#-MI0lSfx-bD7xQsvd_Pp:-MI0lSfx-bD7xQsvd_Pq:bly0cxf) ([raw file](https://github.com/microsoft/microsoft365dsc/blob/ef96501f52649382bbbde50c5e524e6bed1b5221/Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.psm1#L123)):*
> ```PowerShell
> 
>     $ValuesToCheck = $PSBoundParameters
>     $ValuesToCheck.Remove('GlobalAdminAccount') | Out-Null
> ```

We should also remove the ApplicationId, CertificateThumbprint and TenantID parameters here
___
*[Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.schema.mof, line 4 at r2](https://reviewable.io/reviews/microsoft/microsoft365dsc/800#-MI0ldYn69J5QOmOqAtZ:-MI0ldYn69J5QOmOqAt_:b-ow0mmd) ([raw file](https://github.com/microsoft/microsoft365dsc/blob/ef96501f52649382bbbde50c5e524e6bed1b5221/Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.schema.mof#L4)):*
> ```
> class MSFT_AADAdministrativeUnit : OMI_BaseResource
> {
>     [Key, Description("DisplayName of the app")] string DisplayName;
> ```

The descriptions should be updated and the parameters need to reflect the ones passed into the Get/Set/Test methods

blevinscm:
There is no way to uniquely get and set AAD Administrative Units because the Set-AzureADAdministrativeUnit is invalid.  Do you want the Set-Resource there as a stub?

blevinscm:
Hi,  Still looking for guidance on what to do about the Set-Resource.

andikrueger:
Every resource should have a Get, Set and Test method. The challenge with this resource is related to the fact, that there is a `New-AzureADAdministrativeUnit`cmdlet that would allow duplicate names for new Administrative Units. Every unit has a unique `ObjectID` though (which can't be set during creation). 

Maybe you could overcome this issue by creating the following flow:
1. Only run the Set-Method, if
2. a. there is no Administrative Unit with the same name
2. b. the `ObjectID` Parameter is set and equals an existing Administrative Unit. (Set-AyureADAdministrativeUnit)[https://docs.microsoft.com/en-us/powershell/module/azuread/set-azureadadministrativeunit?view=azureadps-2.0-preview] has a paramter `-ObjectId`.

To do 2. b. you would need to add a required parameter `ObjectID` of type GUID to the resource. For the first run, the GUID could be  "00000000-0000-0000-0000-000000000000" and would need to be changed after the first run by a user. This solution is not very practicable, but the only chance I see at the moment.

PR #1125, opened 2021-03-26
ExoSharedMailbox improvements

### LINKED ISSUE #1079: Exporting EXO Shared Mailbox when > 1000 mailboxes
For every Shared Mailbox when exporting EXO Shared Mailboxes, we see the following.  Either -ResultSize Unlimited should be added, or perhaps -Filter if you're only meaning to get a single Shared Mailbox for each.   In larger tenants -ResultSize Unlimited could take an extraordinary length of time

---

WARNING: By default, only the first 1000 items are returned. Use the ResultSize parameter to specify the number of                                                                                    items returned. To return all items, specify "-ResultSize Unlimited". Be aware that, depending on the actual number of  items, returning all items can take a long time and consume a large amount of memory. Also, we don't recommend storing  the results in a variable. Instead, pipe the results to another task or script to perform batch changes.

NikCharlebois:
Improved how we are retrieving all shared mailboxes in the Export-TargetResource function.

- Fixes #1079

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1125)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,17 @@
 # Change log for Microsoft365DSC
 
-# Unreleased
-* MISC
-  * Fixed issue with ODSettings and ExcludedFileExtensions
+# UNRELEASED
+
+* EXOSharedMailbox
+  * Improved how we are retrieving all shared mailboxes in the
+    Export-TargetResource function.
+
+# 1.21.324.1
+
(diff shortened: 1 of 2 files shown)


NikCharlebois:
@desmay could you do a quick PR review? Thanks

PR #1141, opened 2021-04-08
AADConditionalAccessPolicy: fix for default values and new device state attributes

### LINKED ISSUE #1036: [AADConditionalAccessPolicy] Configuration drifts: false positives when using empty arrays
When using an empty array for settings that take arrays as input, the comparison with the tenant setting will report a configuration drift. This is because the tenant value will be reported as $null whereas the acutal value in the config is an empty array @(). Not quite sure whether this is only true for this resource or a general issue.
```
Get-TargetResource Result: 
... IncludeGroups=$null; ...

Target Values: 
... IncludeGroups=(); ...
Test-TargetResource returned False
Completed processing compare operation. The operation returned 
False.
```

Extend the check for diffs to ensure that $null and @() will be considered equal settings.

```
AADConditionalAccessPolicy 
        {
           IncludeGroups         = @();

```

peterabele:
Update for the AADConditionalAccessPolicy Resource

Fixes the default values if the conditional access policy attributes are undefined, aligned to the returned values by Get-AzureADMSConditionalAccessPolicy:

Explicitly returning 

- empty array for translated Location, Group, User and Role conditions if undefined

- Session control attributes are not returned by Get-TargetResource if not defined

Uses the new device state attributes:

- IncludeDevices and ExcludeDevices
- instead of the previously used IncludeDeviceStates and ExcludeDeviceStates
- Requires AAD Powershell version: 2.0.2.134

- Fixes #1036

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1141)


NikCharlebois (approved):
:lgtm:

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1141#-:-MXlON6CDJNlilOxG4F_:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @peterabele)

PR #1203, opened 2021-05-12
Update MSFT_EXOManagementRole.schema.mof

stvnmbr1:
Edit parameter description so it reflects code execution. When "updating" a management role, the role is first deleted and recreated. At this point deployments break if Parent=Null

Edit parameter description so it reflects code execution. When "updating" a management role, the role is first deleted and recreated. At this point deployments break if Parent=Null

None

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1203)


NikCharlebois (approved):
We should also add a check in the Set-TargetResource function to check if Ensure = Present and Parent is null and print an error

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1203#-:-M_Vs1upB04D5tTJIsr3:b-6be8nq)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stvnmbr1)

NikCharlebois:
@stvnmbr1 can you please review the comment about adding an error handling. Thanks

stvnmbr1:
> We should also add a check in the Set-TargetResource function to check if Ensure = Present and Parent is null and print an error
maybe best to include this check,  parent can't be null for Exchange Online. IMO this should also be set as mandatory in the schema.mof and psm1

NikCharlebois:
Setting it as mandatory in the Schema.mof would unfortunately result in a Breaking change. You are right however in saying that because we are not allowing to UnscopedToTopLevel for this resource. Can you confirm that any configuration which would not have the parent set would fail. If so, then even if this would be considered a breaking change, current configs which would not define it would break anyway, and we could get this in as a normal weekly release.

stvnmbr1:
Just reran tests to confirm

deployment without parent defined:
"Failed to invoke DSC Set method: Cannot process argument transformation on parameter 'Parent'. Cannot convert value \"\" to type \"Microsoft.Exchange.Configuration.Tasks.RoleIdParameter\"

Deployment with parent defined:
No errors

NikCharlebois:
Then Breaking change it is. Can you please update the schema.mof file to make the parameter mandatory? Thanks

Adding /@ykuijs for visibility

stvnmbr1:
added "key" in .mof and added "Mandatory = $true" in .psm1

NikCharlebois (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1203#-:-MbBXcTNDL2ZbZNy4CXY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stvnmbr1)

stvnmbr1:
is missing in Test-TargetResource for Parent

PR #1206, opened 2021-05-12
Update MSFT_EXOActiveSyncDeviceAccessRule.psm1

stvnmbr1:
Get-ActiveSyncDeviceAccessRule "Identity" using $NewActiveSyncDeviceAccessRuleParams

invoke before Set-ActiveSyncDeviceAccessRule @SetActiveSyncDeviceAccessRuleParams as "Identity" is set automatically after creating ActiveSyncDeviceAccessRule
this does not reflect the identity set in $SetActiveSyncDeviceAccessRuleParams and is user given

when setting a ActiveSyncDeviceAccessRule, Identity is mismatched due to user input / auto generated "Name"
getting the Identity from the access rule set by exchange before running the set command

please double check

None

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1206)

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_EXOActiveSyncDeviceAccessRule/MSFT_EXOActiveSyncDeviceAccessRule.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_EXOActiveSyncDeviceAccessRule/MSFT_EXOActiveSyncDeviceAccessRule.psm1
@@ -238,6 +238,7 @@ function Set-TargetResource
     {
         Write-Verbose -Message "Active Sync Device Access Rule '$($Identity)' already exists, but needs updating."
         Write-Verbose -Message "Setting Active Sync Device Access Rule $($Identity) with values: $(Convert-M365DscHashtableToString -Hashtable $SetActiveSyncDeviceAccessRuleParams)"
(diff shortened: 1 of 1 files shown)


NikCharlebois (approved):
:lgtm:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1206#-:-M_WkN-u1TUDKoCLG-83:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stvnmbr1)

stvnmbr1:
This fixed the issue with the Set-ActiveSyncDeviceAccessRule and does not break deployment if a rule was already applied to a policy.
However, if a user does not follow template "queryString (characteristic)" as Identity, Get-ActiveSyncDeviceAccessRule returns the rule is not found and invokes New-ActiveSyncDeviceAccessRule which breaks deployment. Get-ActiveSyncDeviceAccessRule tries to create a new rule with the same template "queryString (characteristic)" as Identity, but breaks as the rule already exists.
Either docs should reflect template "queryString (characteristic)" as Identity or Identity should not be mandatory but derived from set queryString & characteristic, which in turn should be mandatory in this use case.


Any thoughts on this matter?

PR #1215, opened 2021-05-19
EXOMalwareFilterRule: Fix for issue #1213

### LINKED ISSUE #1213: Policy X already has rule Y associated with it
EXOSafeAttachmentPolicy: if a rule is present and $SafeAttachmentPolicy is equal to the set SafeAttachmentPolicy for that rule, deployment breaks. As the Set-SafeAttachmentRule do not allow assignment of the same Safe Attachment Policy on the same rule.

"msg": "Failed to invoke DSC Set method: Policy \"Safe Attachment Policy\" already has rule \"Safe Attachment Rule\" associated with it.",

Add extra check: if $SafeAttachmentPolicy is equal to the set SafeAttachmentPolicy, remove the $SafeAttachmentPolicy variable from  $PSBoundParameters, before invoking  Set-SafeAttachmentRule. This way a different assignment of rule to policy will work and editing parameters of an existing rule will not break deployment 

Dev

stvnmbr1:

EXOMalwareFilterRule: Fix for issue #1213
Policy X already has rule Y associated with it 

Fixes #1213
Policy X already has rule Y associated with it

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1215)


stvnmbr1:
added EXOSafeAttachmentRule: Fix for issue #1213
Policy X already has rule Y associated with it if rule exists already

Can you confirm only one Safe Attachment Rule per Safe Attachment Policy is allowed?

stvnmbr1:
added EXOSafeLinksRule: Fix for issue #1213
Policy X already has rule Y associated with it if rule exists already

Can you confirm also only one Safe Links Rule per Safe Links Policy is allowed?

PR #1303, opened 2021-07-13
EXOOrganizationConfig - need expanded property

stvnmbr1:
need to expand property to get value for next if statement

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/microsoft365dsc/1303)


stvnmbr1:
can be ignored / closed

PR #1621, opened 2021-12-10
Various Fixes

### LINKED ISSUE #1620: Import AntiPhishPolicy Release 1.21.1208.1 not working
Hello,

When i try to use the new module from antiphishpolicy 1.21.1208.1 i get a error that the system already see System.Collections.Hashtable. when i go look in the exchange policy i see that the policy don't get the ocrrect identity and it use System.Collections.Hashtable.

When i import it with module 1.21.1117.2 it work great with the same information.

NikCharlebois:
* AADConditionalAccessPolicy
  * Switched to the beta endpoint to allow the export to capture policies
    with device compliance conditions configured;
* EXOAntiphishPolicy
  * Fixed an issue where trying to create a new policy would result in its
    Identity being set to System.Collections.Hashtable;
    FIXES #1620
* O365User
  * Fixed issue where the extraction wasn't properly formatting the temporary
    password for a user;
* TeamsMeetingPolicy
  * Added support for value 'EveryoneInCompanyExcludingGuests' for the
    AutoAdmittedUsers property;
* DEPENDENCIES
  * Updated MSCloudLoginAssistant to 1.0.82;

- Fixes #1620


Modules/Microsoft365DSC/DSCResources/MSFT_AADGroupsSettings/MSFT_AADGroupsSettings.psm1
```diff
@@ -68,7 +68,6 @@ function Get-TargetResource
         -InboundParameters $PSBoundParameters `
         -ProfileName 'beta'
 
-    $MaximumFunctionCount = 32000
```
ykuijs, line 71: Shouldn't this change be in the changelog as well?

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
...
-#1.21.1208.1
+# 1.21.1215.1
+
+* AADConditionalAccessPolicy
+  * Switched to the beta endpoint to allow the export to capture policies
+    with device compliance conditions configured;
+* EXOAntiphishPolicy
+  * Fixed an issue where trying to create a new policy would result in its
+    Identity being set to System.Collections.Hashtable;
+    FIXES #1620
```
ykuijs, line 11: Why is this in here?

ykuijs (changes requested):
Just a few comments

Modules/Microsoft365DSC/DSCResources/MSFT_AADGroupsSettings/MSFT_AADGroupsSettings.psm1
NikCharlebois: I don't think we need to make a mention of this to be honest. This was fixed by the previous release this week and therefore no longer needed.

CHANGELOG.md
NikCharlebois: Because it fixes that issue. We've always listed changes that are related to a specific issue that way in the past.

ykuijs (approved):

PR #1656, opened 2021-12-30
Fix Issue #1648

### LINKED ISSUE #1648: EXOMalwareFilterPolicy MakeDefault Test problem
When I modify the default EXOMalwareFilterPolicy 'Default' and provide the MakeDefault = True attribute in the DSC configuration and the policy is already the default, the Test method returns false stating that this resource is not in the desired state.

**PowerShell Script**

``` 
PS C:\Windows\system32> get-MalwareFilterPolicy -identity 'Default' | select Action,IsDefault

Action                                 IsDefault
------                                 ---------
DeleteAttachmentAndUseDefaultAlertText      True
```

When I remove the MakeDefault Parameter from the DSC configuration, apply the configuration and test the config again, the resource is not in the desired state.

Hence, it needs to be checked and corrected that the test method returns false when the policy is already the default.
Version of the DSC module that was used ('dev' if using current dev branch)

1.21.1222.1

NikCharlebois:
* Fix for the MakeDefault property where it wasn't properly setting existing policies to default.

 - Fixes #1648

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change log for Microsoft365DSC
 
+# 1.22.105.1
+
+* EXOMalwareFilterPolicy
+  * Fix for the MakeDefault property where it wasn't properly setting existing
+    policies to default.
+    FIXES #1648
+
 # 1.21.1229.1
 
 * DEPENDENCIES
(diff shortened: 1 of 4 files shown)


NikCharlebois:
@ykuijs can you do the code review for this one? Thanks

ykuijs (approved):
LGTM!

PR #1660, opened 2021-12-31
Fixes #1548

### LINKED ISSUE #1548: New-M365DSCDeltaReport issues when using as an application
Hey everyone

I have started moving from using credentials to application auth via certificate, I transcripted an export and comparison where it uses App Auth, the comparison fails and reports on all resources failing.

Exports work fine, it's just the delta comparison fails. Exports are done using the same version of the module as well.

PS>TerminatingError(Compare-Object): "Cannot bind argument to parameter 'ReferenceObject' because it is null."
Error "ResourceName"

This does not happen and works fine with Teams/SecComp since User Credential is required there, but for AAD/EXO/SPO/Intune it fails.

Anyone know what could be wrong? Any assistance would be great.

NikCharlebois:
* Error Handling in Delta Report and removal of Authentication mechanism comparison;

- Fixes #1548


ykuijs (approved):
LGTM

PR #1700, opened 2022-01-21
Fixes #1695

### LINKED ISSUE #1695: [MSFT_TeamsTenantDialPlan] Does not export multiple Normalization rules
I tried exporting a tenant dial plan for Teams which has multiple normalization rules. Only the first rule is being exported, the rest is being ommitted.
I believe the issue is in https://github.com/microsoft/Microsoft365DSC/blob/ccb52e536b12c7a28285bf24bb59504928e0fbc8/Modules/Microsoft365DSC/DSCResources/MSFT_TeamsTenantDialPlan/MSFT_TeamsTenantDialPlan.psm1#L442

I have checked the type of the return value for normalization rules:
```
$p = Get-CsTenantDialPlan
$p[0].NormalizationRules.gettype()


IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     ArrayList                                System.Object
```

The code is checking for a hashtable. Thus, it defaults to the else section where it only picks the first rule. The type needs to be corrected.

```powershell
# insert configuration here
```

NikCharlebois:
*TeamsTenantDialPlan
  * Fixed an issue where the Export only extracted the first
    normalization rule.

N/A

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -77,7 +77,7 @@ function Get-TeamByName
     $loopCounter = 0
     do
     {
-        $team = Get-Team -DisplayName $TeamName
+        $team = Get-Team -DisplayName $TeamName | Where-Object -FilterScript {$_.DisplayName -eq $TeamName}
         if ($null -eq $team)
         {
             Start-Sleep 5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # 1.22.126.1
 
+*TeamsTenantDialPlan
+  * Fixed an issue where the Export only extracted the first
+    normalization rule.
+    FIXES #1695
 * DEPENDENCIES
   * Updated Microsoft.Teams to version 3.1.0;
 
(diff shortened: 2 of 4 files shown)


NikCharlebois:
@ykuijs can you please do the code review? Thanks

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -77,7 +77,7 @@ function Get-TeamByName
     $loopCounter = 0
     do
     {
-        $team = Get-Team -DisplayName $TeamName
+        $team = Get-Team -DisplayName $TeamName | Where-Object -FilterScript {$_.DisplayName -eq $TeamName}
```
ykuijs, line 80: Parameters look the same, both checking DisplayName against $TeamName. Shouldn't it be either the parameter or the Where-Object?

ykuijs (changes requested):
One question

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
NikCharlebois: Get-Team -DisplayName, based on documentation is a wild card search. In our case we want an exact match:
https://docs.microsoft.com/en-us/powershell/module/teams/get-team?view=teams-ps#example-4

ykuijs (approved):
LGTM

PR #1749, opened 2022-02-09
Fixes #1746

### LINKED ISSUE #1746: TeamsChannel: Resource uniqueness for exports - New key is needed
Extraction works on my end with M365DSC V 1.22.119.2. 

image

The extraction looks good but the configuration can not be compiled as we do have duplicate elements within the configuration - The mof file is named accordingly: `localhost.mof.error`

```powershell
       TeamsChannel 7f8d78b7-4d6a-4898-ae3b-13527be9e831
        {
            DisplayName          = "General";
            Ensure               = "Absent";
            TeamName             = "IT";
        }
        .....
        TeamsChannel e7bdd703-d005-4ae2-9dd4-fc644a7e3db7
        {
            DisplayName          = "General";
            Ensure               = "Absent";
            TeamName             = "IT";
        }
``` 

The TeamsTeam resources do have `GroupID` so that we can differentiate the teams. The `TeamsTeam` has handling for that in place: https://github.com/microsoft/Microsoft365DSC/blob/Dev/Modules/Microsoft365DSC/DSCResources/MSFT_TeamsTeam/MSFT_TeamsTeam.psm1#L148-L173

Either we should introduce this for `TeamsChannel` too, or we need some other key. This would not need to be a breaking change, as we would only need to add the `groupId` to this resource.

_Originally posted by @andikrueger in https://github.com/microsoft/Microsoft365DSC/issues/1579#issuecomment-1018789279_

NikCharlebois:
Fixes issue where Teams Channel were not extracted properly when two teams had the same name.

- Fixes #1746

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_TeamsChannel/MSFT_TeamsChannel.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_TeamsChannel/MSFT_TeamsChannel.psm1
@@ -15,6 +15,10 @@ function Get-TargetResource
 
         [Parameter()]
         [System.String]
+        $GroupID,
+
+        [Parameter()]
+        [System.String]
         [ValidateLength(0, 50)]
         $NewDisplayName,
 
@@ -67,7 +71,20 @@ function Get-TargetResource
 
     try
     {
-        $team = Get-TeamByName ([System.Net.WebUtility]::UrlEncode($TeamName))
+        if (-not [System.String]::IsNullOrEmpty($GroupId))
+        {
+            $team = Get-Team -GroupId $GroupId -ErrorAction 'SilentlyContinue'
+        }
+
+        if ($null -eq $team)
+        {
+            $team = Get-TeamByName ([System.Net.WebUtility]::UrlEncode($TeamName))
+        }
+
+        if ($null -eq $team)
+        {
+            return $nullReturn
+        }
 
         Write-Verbose -Message "Retrieve team GroupId: $($team.GroupId)"
 
@@ -93,6 +110,7 @@ function Get-TargetResource
         $results =  @{
             DisplayName           = $channel.DisplayName
             TeamName              = $team.DisplayName
+            GroupId               = $team.GroupId
             Description           = $channel.Description
             Ensure                = "Present"
             ApplicationId         = $ApplicationId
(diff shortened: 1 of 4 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsChannel/MSFT_TeamsChannel.psm1
```diff
@@ -67,7 +71,20 @@ function Get-TargetResource
...
     {
-        $team = Get-TeamByName ([System.Net.WebUtility]::UrlEncode($TeamName))
+        if (-not [System.String]::IsNullOrEmpty($GroupId))
+        {
+            $team = Get-Team -GroupId $GroupId -ErrorAction 'SilentlyContinue'
+        }
+
+        if ($null -eq $team)
+        {
+            $team = Get-TeamByName ([System.Net.WebUtility]::UrlEncode($TeamName))
```
andikrueger, line 81: This could still cause an issue as there might be a team with the same name but not a matching groupID. In this case we would add the channel to another team. This if should have a second condition: `and [System.String]::IsNullOrEmpty($GroupId)`

NikCharlebois: Good catch.

NikCharlebois: However, even if the Group Id is provided and is invalid, this would still result in this being called. If the group ID is valid, then we will never get inside of this condition since it will have been returned by the previous call to Get-Teams -GroupId

andikrueger: Why would we add a channel to a team then, if the groupid is not valid? This would only be the case in tenant to tenant migrations, wouldn't it?

NikCharlebois: Yes, tenant to tenant migration would be a scenario in that bucket here.

ykuijs (approved):

PR #1776, opened 2022-02-22
New Resource: TeamsFederationConfiguration

malauter:
Add new DSC resource TeamsFederationConfiguration


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsFederationConfiguration/MSFT_TeamsFederationConfiguration.psm1
```diff
@@ -0,0 +1,279 @@
...
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftTeams' `
+        -InboundParameters $PSBoundParameters
+
+    $SetParams = $PSBoundParameters
+    $SetParams.Remove("Credential")
+
+    Set-CsTenantFederationConfiguration @SetParams
```
NikCharlebois, line 140: Based on the definition of the Set-CsTenantFederationConfiguration (https://docs.microsoft.com/en-us/powershell/module/skype/set-cstenantfederationconfiguration?view=skype-ps) it doesn't accept parameter Identity. Can you validate if this is a bug or a miss in the documentation?

NikCharlebois (changes requested):
- We will need to add the pester tests for the resource and an example as well.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsFederationConfiguration/MSFT_TeamsFederationConfiguration.psm1
malauter: The cmdlet accepts the Identity parameter: 
image

ykuijs:
There is something in the Unit test that hangs. When reviewing the logs, I see the TeamsEmergencyCallRoutingPolicy unit tests run, but then hangs for six hours:
image

Since the new resource is called TeamsFederationConfiguration, I think there is something wrong with this test.

malauter:
> There is something in the Unit test that hangs. When reviewing the logs, I see the TeamsEmergencyCallRoutingPolicy unit tests run, but then hangs for six hours: image
> 
> Since the new resource is called TeamsFederationConfiguration and the tests are ran alphabetically, I think there is something wrong with this test.

I know, but I can still not find what is wrong with the test. It is a adjusted copy of another working test and I checked it several times. No clue what is wrong here

ykuijs:
When I run the test locally (just run the ps1 file) I am prompted for credentials:
image
That explains why the code waits forever. 

When reviewing your test, the first test is using these parameters:
```PowerShell
$testParams = @{
    AllowFederatedUsers       = $True;
    AllowPublicUsers          = $True;
    AllowTeamsConsumer        = $False;
    AllowTeamsConsumerInbound = $False;
    Identity                  = "Global";
}
```

When I check the resource, I see that the Credential parameter is required:
```PowerShell
[System.Management.Automation.PSCredential]
$Credential
```

So the issue here is that the code requires a Credentials parameter and the test isn't providing that. So then PowerShell prompts for a password, which is never provided since the tests are ran in an unattended process.

If you correct this, the test should run much better.

NikCharlebois (approved):

PR #1793, opened 2022-03-01
AADNamedLocationPolicy: add additional auth Method to Ressource

Timsto:


NikCharlebois:
Hey @Timsto, we are trying to standardize our authentication parameters across the board to either App secret or Thumbprint for certificates. Currently, all AAD resources use Microsoft Graph in the back-end and it currently doesn't support using Path or Password. Therefore simply adding these here won't add support for it. Do you have a use case that justifies using a local certificate path over specifying a thumbprint? Thanks

Timsto:
so PR could be closed

ykuijs:
Closing PR

PR #1849, opened 2022-04-01
New PPTenantIsolationSettings resource, updated settings.json files and other minor updates

### LINKED ISSUE #1844: [PPTenantIsolationSettings] New resource
### Description
In the Power Platform Admin Center, you can limit the people who are able to install Data Gateways:
Power Platform Admin Center >  Policies > Tenant isolation (preview)

### Proposed properties
IsSingleInstance - Boolean
Enabled - Boolean
TenantRules - Array (TenantRule)
TenantRulesToInclude - Array (TenantRule)
TenantRulesToExclude - Array (TenantRule)

**_TenantRule_**
TenantName - String
Direction - String (Inbound/Outbound/Both)

Set-PowerAppTenantIsolationPolicy: https://docs.microsoft.com/en-us/powershell/module/microsoft.powerapps.administration.powershell/set-powerapptenantisolationpolicy
Get-PowerAppTenantIsolationPolicy: https://docs.microsoft.com/en-us/powershell/module/microsoft.powerapps.administration.powershell/get-powerapptenantisolationpolicy

ykuijs:
This PR adds a new resource (PPTenantIsolationSettings) which is able to configure the PowerApps Tenant Isolation settings.

It also updates the settings.json files with updated permissions and fixes a few minor other inconsistencies.

- Fixes #1844

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_PPTenantIsolationSettings/MSFT_PPTenantIsolationSettings.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_PPTenantIsolationSettings/MSFT_PPTenantIsolationSettings.psm1
@@ -1,0 +1,688 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [ValidateSet('Yes')]
+        [System.String]
+        $IsSingleInstance,
+
+        [Parameter()]
+        [System.Boolean]
+        $Enabled = $true,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $Rules,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $RulesToInclude,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $RulesToExclude,
+
+        [Parameter()]
+        [System.Management.Automation.PSCredential]
+        $Credential
+    )
+
+    Write-Verbose -Message 'Getting the Power Platform Tenant Isolation Settings Configuration'
+
+    if ($PSBoundParameters.ContainsKey("Rules") -and `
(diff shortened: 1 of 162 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_PPTenantIsolationSettings/MSFT_PPTenantIsolationSettings.schema.mof
```diff
@@ -0,0 +1,17 @@
+[ClassVersion("1.0.0")]
+class MSFT_TenantRule
```
NikCharlebois, line 2: Should we give the CIM Instance a more specific name (e.g. PPTenantRule) in case there are other workloads somewhere that also need to define tenantrules?

Modules/Microsoft365DSC/DSCResources/MSFT_PPTenantIsolationSettings/MSFT_PPTenantIsolationSettings.psm1
```diff
@@ -0,0 +1,688 @@
...
+                    $direction = 'Inbound'
+                }
+            }
+            elseif ($directions.outbound -eq $true)
+            {
+                $direction = 'Outbound'
+            }
+            else
+            {
+                $direction = 'unknown'
```
NikCharlebois, line 88: Shouldn't we throw an error here instead since unknown is not a valid value?

NikCharlebois (changes requested):
2 Items to review

Modules/Microsoft365DSC/DSCResources/MSFT_PPTenantIsolationSettings/MSFT_PPTenantIsolationSettings.psm1
ykuijs: Throwing an error will cause the Get/Test to fail and the Set method to never be called. By using this option, the Test will fail and the Set method will be called.

NikCharlebois (approved):

NikCharlebois:
:shipit:

PR #1867, opened 2022-04-06
Fix JSON export function in M365DSCReport.psm1

berg95:

The current JSON export function includes a logic error as well as using the unnecessary ConvertTo-json function.
I have implemented a short term fix.


berg95:
Disregard this pull request. I have a much better solution to this problem. I should never has submitted this pull request. I will submit a new one with my updated implementation shortly,

PR #1922, opened 2022-05-04
EXOSharedMailbox: Fix using umlauts in displayname by allowing to set alias (fixes #1921)

### LINKED ISSUE #1921: Creating shared mailbox fails when name contains umlaut
Creating shared mailbox currently fails when name contains an umlaut.

Example DSC:
```
Configuration Example
{
    param(
        [PSCredential]
        $credsGlobalAdmin
    )
    Import-DscResource -ModuleName Microsoft365DSC

    node localhost
    {
        EXOSharedMailbox 'SharedMailbox'
        {
            DisplayName        = "Vertrieb München"
            PrimarySMTPAddress = "vertrieb-muenchen@contoso.com
            Ensure             = "Present"
            Credential         = $credsGlobalAdmin
        }
    }
}
```

The following error occurs:
```
An Azure Active Directory call was made to keep object in sync between Azure Active Directory and Exchange Online. However, it failed. Detailed error message:
        The value specified for property EmailAddresses is incorrect. Reason: InvalidValue DualWrite (Graph) RequestId: 8300a04f-fe5f-4aeb-9d2a-9414fd8bdcd0
The issue may be transient and please retry a couple of minutes later. If issue persists, please see exception members for more information.
At C:\Users\Administrator\AppData\Local\Temp\tmp_gdhhwhyw.3cd\tmp_gdhhwhyw.3cd.psm1:35519 char:9
+         $steppablePipeline.End()
+         ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [New-Mailbox], UnableToWriteToAadException
    + FullyQualifiedErrorId : [Server=FR0P281MB1578,RequestId=a9246f42-c42a-4a7d-9c44-0ea1952fac0e,TimeStamp=04.05.2022 15:27:47] [FailureCategory=Cmdlet-UnableToWriteToAadEx
   ception] C302C903,Microsoft.Exchange.Management.RecipientTasks.NewMailbox
    + PSComputerName        : outlook.office365.com
```                                                                                                                                                                     

While the error message is cryptic, the reason for this is as follows: if no `-Alias` is provided, it is constructed from the `-Name`. In case of `Vertrieb München`, the alias `VertriebMünchen` is constructed. 

This alias is then used to construct the [Microsoft Online Email Routing Address](https://docs.microsoft.com/en-us/troubleshoot/azure/active-directory/objects-dont-sync-ad-sync-tool#update-user-smtp-addresses-by-using-on-premises-active-directory-attributes) (in this case: `VertriebMünchen@contoso.onmicrosoft.com`).  

Saving this email address to azure ad fails because MS does not allow umlauts in the user-part of an emailaddress.

Proposed solution: allow settings `-Alias` via DSC parameter.

Yannik:
Fix using umlauts in displayname by allowing to set alias (fixes #1921)

    - Fixes #1921


NikCharlebois:
Please verify the Unit tests that are failing. Thanks

Yannik:
> Please verify the Unit tests that are failing. Thanks

Hi Nik, I already tried to find out why the unit test fails but didn't manage to isolate the issue. Could you please have a look?

NikCharlebois:
@Yannik I believe this might be due to a previous error in our util module. Can you please update your branch with the latest from dev and commit, this should solve it. Thanks

Yannik:
@NikCharlebois Done.

malauter:
Just had a quick look, but I think the reason for the failing test is in your Set-TargetResource function. You added the parameter $Alias and in line 304 you try to access the value of this parameter. But in line 217 there is a `foreach ($alias in $Aliases)` (Aliases = ProxyAddresses) which may overwrite the value of the parameter $Alias.

Yannik:
@malauter 
The foreach $alias assignment surely only only applies to the foreach-scope?

malauter:
@Yannik As far as I know the assignment is done in the caller´s scope. Therefore the variables persists in the scope after the loop completes. I did a quick manual test on PowerShell, which confirms this assumption.

Yannik:
@malauter You are right. Powershell is weird.

I've renamed the foreach $Alias variable to fix this. 

My suggestion would be to rename `$Aliases` to something different tbh. 

Imo `$Aliases` is pretty confusing at the moment, as there is an official `-Alias`, but `-Aliases` is not using in the official `New-Mailbox` command at all.

`ProxyAddresses` or `EmailAddresses` would be a better choice.

Or, if you want to communicate that the PrimarySmtpAddress is included by default, maybe `$SecondaryAliases` or `AdditionalEmailAddresses` or something similar would be a better choice..

malauter:
@Yannik I think your last fix is incomplete, the variable should be renamed also in lines 217, 277 and 292 to prevent this issue in all cases.

But I fully agree - Aliases is definitely not a good choice. I would prefer AdditionalEmailAddresses. Please feel free to create another PR to improve this. But since existing DSC configurations will not work anymore after the renaming to AdditionalEmailAddresses, this will be a breaking change and must wait until the next breaking change release.

Yannik:
@malauter Thanks for the heads up, I think now I've got them all.

Yannik:
@malauter Fixed

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSharedMailbox/MSFT_EXOSharedMailbox.psm1
```diff
@@ -12,6 +12,10 @@ function Get-TargetResource
         [System.String]
         $PrimarySMTPAddress,
 
+        [Parameter()]
+        [System.String]
+        $Alias,
+
         [Parameter()]
         [System.String[]]
         $Aliases,
```
NikCharlebois, line 21: Since this maps to the EmailAddresses property of the Set-Mailbox cmdlet, we should rename it to match the proper name. This would however be a breaking change, therefore we should probably add a new parameter EmailAddresses and keep the existing $Aliases one there, but mark it as deprecated and print a warning if users are attempting to use it instead of the EmailAddresses one.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSharedMailbox/MSFT_EXOSharedMailbox.psm1
Yannik: I agree that this would be a good change, but it's not within the scope of this PR.

NikCharlebois (approved):

PR #1973, opened 2022-05-29
DRG - Fixes #1963 + multi level EmbeddedInstances

### LINKED ISSUE #1963: DRG - Issues found while testing on MgAgreement
Hey,
I have been testing the DRG on MgAgreement and found the following issues:

New-M365DSCResource -ResourceName AADAgreement -Workload MicrosoftGraph -GraphModuleCmdletNoun 'MgAgreement' -Path $ResourcePath -UnitTestPath $UnitTestPath -ExampleFilePath $ExamplePath -Credential $creds 

Schema:

Complex resources are missing attributes inherited from parent interface
For example: 
the parameter:
[Write, Description(""), EmbeddedInstance("MSFT_MicrosoftGraphagreementfile")] String File;

The embedded class generated is as followed:
[ClassVersion("1.0.0")]
class MSFT_MicrosoftGraphagreementfile
{
    [Write, Description(""), EmbeddedInstance("MSFT_MicrosoftGraphagreementfilelocalization")] String Localizations[];
};

Looking at M365DSCResourceGenerator and Get-M365DSCDRGCimInstances

The code call for 
(Invoke-Expression "[Microsoft.Graph.PowerShell.Models.IMicrosoftGraphAgreementFile]").DeclaredProperties which only returns localizations however this resource implements the interface 
[Microsoft.Graph.PowerShell.Models.IMicrosoftGraphAgreementFile] from where we will be able to return the full list of parameters for MSFT_MicrosoftGraphagreementfile

Because the schema is missing parameters the code and class generated can not be used  to generate new instances 

Resource file:

in Get-TargetResource
The code generated try to retrieve the MgAgreement using a displayname and the additionalproperties.@odata.type as followed:

$getValue = Get-MgAgreement `
            -ErrorAction Stop | Where-Object `
            -FilterScript {
$_.AdditionalProperties.'@odata.type' -eq '#microsoft.graph.agreement' -and  $_.displayName -eq $($DisplayName)
            }
however $_.AdditionalProperties.'@odata.type' -eq '#microsoft.graph.agreement'  is not published, so removing the condition will allow to retrieve the resource by its displayname:
        $getValue = Get-MgAgreement `
            -ErrorAction Stop | Where-Object `
            -FilterScript {
                $_.displayName -eq $($DisplayName)
}

Another issue I am experiencing but this one I am not sure if this is due to my environment or not (even though I have seen some issues reported caused by this #1887) is the following:
        $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftGraph' `
            -InboundParameters $PSBoundParameters -ProfileName 'v1.0'
        Select-MgProfile 'v1.0' -ErrorAction Stop

The command does not connect to Graph all the time if v1.0 is used in New-M365DSCConnection but works if I use beta - It can be intermittent however depending on how many time I run the Export-M365DSCConfiguration: first try only works with beta but all subsequent calls to the function (without reimporting M365DSC module ) will only work with v1.0 - so I have added the following code to ensure successful connection at any times:
 $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftGraph' `
            -InboundParameters $PSBoundParameters -ProfileName 'v1.0'
$context=Get-MgContext
    if($null -eq $context)
    {
        $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftGraph' `
        -InboundParameters $PSBoundParameters -ProfileName 'beta'
    }
    Select-MgProfile 'v1.0' -ErrorAction Stop

I only tested the Export-TargetResource and Get-TargetResource due to the issue with  the missing parameters from complex resources.
The example file was not generated due to the issue with the connection to graph and get-mgagreement in export-targetresource

I hope this helps
Thanks,
Will
```powershell
# insert configuration here
```

William-Francillette:

Hey,

This PR fixes #1963 and other issues when resources includes multi level Embedded Instances

This PR has been tested with the following Resources successfully:
MgAgreement
MgIdentityConditionalAccessPolicy
MgDevice
MgOrganization

I was able to generate the resource and exploit it using 
Export-M365DSCConfiguration
Compiled the resource successfully 
and create/update and delete resource from the generated mof file

MgIdentityConditionalAccessPolicy was mostly used to fix the various issues with recursion

I hope this helps
Thanks,
William


NikCharlebois:
Thanks for the good work @William-Francillette. There seems to be an issue with the logic when encountering parameters of types "". For example, try generating the resource for the AADomain (MgDomain). The **FederationConfiguration** parameter currently gets assigned an invalid type 
```
[Parameter()]
[c(internaldomainfederation)]
$FederationConfiguration,
```

It appears that this is due to the fact that this parameter is not part of the default parameter set. My recommendation here would be to simply ignore a parameter and omit it from the resource if a parameter is not part of the default set.

William-Francillette:
> Thanks for the good work @William-Francillette. There seems to be an issue with the logic when encountering parameters of types "". For example, try generating the resource for the AADomain (MgDomain). The **FederationConfiguration** parameter currently gets assigned an invalid type
> 
> ```
> [Parameter()]
> [c(internaldomainfederation)]
> $FederationConfiguration,
> ```
> 
> It appears that this is due to the fact that this parameter is not part of the default parameter set. My recommendation here would be to simply ignore a parameter and omit it from the resource if a parameter is not part of the default set.

Hey @NikCharlebois , I see the issue: the definition which is returned is outdated thus the attribute is not returned - I have added an additional step find the correct type for those attributes or skip it if not found.
I also found the code not working for CIM instance arrays parameters, so I'm working on that at the moment. As soon as it is sorted, I'll let you know

William-Francillette:
Hi @NikCharlebois,

I fixed the issue you reported with the unknown parameter as suggested by simply ignoring them.

I've also been testing additional resources such as MgSubscribedSku which was a good candidate for CimInstance arrays - it should be all working now.

Next step was to test some other big resources such as MgUser or MgTeams. 
There was an issue with Get-M365DSCDRGCimInstances and the DiscoveredComplexTypes parameters - it was going in an infinite loop so rather passing the attribute in the function definition I used a script variable to fix the issue.

When testing the generated MgUser, I found an issue related to the attribute Drive of type MicrosoftGraphDrive: it contains an attribute called _root_ of type MicrosoftGraphDriveItem. This parameters contains _ListItem_, and _ListItem_ contains _DriveItem_. Those 2 can be queried indefinitely and creates an infinite loop in Test-M365DSCComplexObjectHasValues. If I skip the attribute key called root, it then works properly and was able to export, update and create a user using this generated resource.
I haven't modified module.template because I didn't want to skip _root_ for all resources generated in case of it was needed in other scenarios.

Another one I found was in the generated schema: some classes uses parameters which are keywords in the schema such as DateTime, Method, Scope - Not sure what will be the best path to tackle this conflict? Is there a list of all reserved keywords in the schema we can use as reference? 

Some resources couldn't generate due to a JSON error in Get-DerivedType: the cleanJsonString could not be converted from Json (I didn't looked into that issue)
   ` $rawJson = Invoke-RestMethod -Method Get -Uri "https://metadataexplorerstorage.blob.core.windows.net/$web/$($Version).js#search:$($Entity)"
    # Clean JSON
    $cleanJsonString = $rawJson.TrimStart("const json = ")
    $cleanJsonString = $CleanJsonString -replace ",}", "}"
    $targetIndex = $cleanJsonString.IndexOf(";")
    $cleanJsonString = $cleanJsonString.Remove($targetIndex)
    $entityRelationships = $cleanJsonString | ConvertFrom-Json`

Anyway, that's all my findings for now - I hope this helps - Let me know if there's anything I missed and I'll be happy to try fix it
Thanks,
Will

Issue #1757, opened 2022-02-11
Installing Microsoft365DSC in Scope CurrentUser


andikrueger:
### Description
Currentlz the modules architecture does not allow a setup of the module in the Scope of a user. This is limitation in some cases for extraction as there are user that are not local admins on their environments.

### Proposed properties

Install-Module should be able with -Scope CurrentUser

The crucial part is `Update-M365DSCDependencies` does not have a scope parameter.

PR #2013, opened 2022-06-17
Fixes to installation Scope

NikCharlebois:
Modified the dependency installation functions to for the AllUsers scope.

N/A

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -1051,11 +1051,18 @@ function Export-M365DSCConfiguration
     $outdatedOrMissingAssemblies = Test-M365DSCDependencies
     if ($outdatedOrMissingAssemblies)
     {
-        foreach ($dependency in $outdatedOrMissingAssemblies)
+        if (([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator))
         {
-            Write-Host "Updating dependency {$($dependency.ModuleName)} to version {$($dependency.RequiredVersion)}..." -NoNewline
-            Install-Module $dependency.ModuleName -RequiredVersion $dependency.RequiredVersion -Force -Scope 'AllUsers'  | Out-Null
-            Write-Host $Global:M365DSCEmojiGreenCheckmark
+            foreach ($dependency in $outdatedOrMissingAssemblies)
+            {
+                Write-Host "Updating dependency {$($dependency.ModuleName)} to version {$($dependency.RequiredVersion)}..." -NoNewline
+                Install-Module $dependency.ModuleName -RequiredVersion $dependency.RequiredVersion -Force -Scope 'AllUsers'  | Out-Null
+                Write-Host $Global:M365DSCEmojiGreenCheckmark
+            }
+        }
+        else
+        {
+            Write-Error "Cannot update the dependencies for Microsoft365DSC. You need to run this command as a local administrator."
         }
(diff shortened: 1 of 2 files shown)


andikrueger:
This will resolve #1757 or at least bring a better solution to the handling of scopes.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -1054,7 +1054,7 @@ function Export-M365DSCConfiguration
         foreach ($dependency in $outdatedOrMissingAssemblies)
         {
             Write-Host "Updating dependency {$($dependency.ModuleName)} to version {$($dependency.RequiredVersion)}..." -NoNewline
-            Install-Module $dependency.ModuleName -RequiredVersion $dependency.RequiredVersion -Force | Out-Null
+            Install-Module $dependency.ModuleName -RequiredVersion $dependency.RequiredVersion -Force -Scope 'AllUsers'  | Out-Null
```
ykuijs, line 1057: This will require that the user is running as a local admin, else it will throw an error. Maybe it is a good idea to check if the user has these permissions and if not, show a nice message:

You can use this code to check if the user has admin permissions:
`([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)`

ykuijs (commented):
Just an idea, the rest looks fine

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
NikCharlebois: Great suggestion. Code updated to add this check and throw a meaningful error if the user is not an admin. Thanks

andikrueger: Just a further suggestion from a code refactoring context: Encapsulate this one in a function `Test-AdminAccountUsage` to add additional context information e.g. "You need to run this code as Admin"

```powershell

function Test-AdminAccountUsage{
 $adminrole = ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
 if($adminrole){
  Write-Verbose "This command needs Admin Permissions"
 }
 return $adminrole
}
```

NikCharlebois: While I normally like to wrap as much stuff in functions to promote re-use and reduce duplication of code, this admin check will almost always need to be wrapped inside an if statement itself to prevent further execution of code when found to be false. Because this function would end up being a simply if check itself, it doesn't make sense in my opinion to have if further wrapped in conditional logic.

E.g., 

if (Test-AdminAccountUsage)
{
    #do Work
}
[...]

function Test-AdminAccountUsage{
 $adminrole = ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
 if($adminrole){
  Write-Verbose "This command needs Admin Permissions"
 }
 return $adminrole
}

andikrueger: That is a valid point. Having a second handling within the function is nonsense. Only benefit would be a better readability...

Could we try to self elevate the process to gain a bit more usability?

```
function Test-AdminAccountUsage{
return ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
}
```

PR #2030, opened 2022-06-27
MSFT_ODSettings - DomainGuids fix for #2021

### LINKED ISSUE #2021: MSFT_ODSettings - applying a template as exported with DomainGuids = @() results in an invalid configuration
When extracting a ODsettings configuration if the 'allow syncing only on computers joined to specific domains' value is unchecked the configuration document will return a value:
DomainGuids                               = @();

If this same document is used to re-apply the settings later it enables the 'allow syncing only on computers joined to specific domains' value and the DomainGuids list is left blank.

At the very least it appears to be an invalid configuration - you can't set the values like this via the GUI and if you open the GUI editor after applying the template you'd need to either disable domain blocking or add a GUID to the list to allow you to save the settings.  I would also expect this to prevent desktop clients from syncing because there are no domains on the authorised list.

Omitting the DomainGuids value from the configuration document stops the problem occurring and also fixes the issue if it has already been set like this (so long as another value is changed to ensure the configuration is re-applied)

N/A - settings appear to apply correctly so not really any benefit here

The checkbox is controlled via the TenantRestrictionEnabled -  value isn't available in the configuration at the moment although it can be set via Set-PnPTenantSyncClientRestriction

possible solutions are -
- add in TenantRestrictionEnabled to the configuration

- Use the value of DomainGuids to set both values - with an empty array setting the value of TenantRestrictionEnabled to $false.
 
- amend exported configurations so that the DomainGuids value is omitted from the document if it is empty

I think the second option logic would be required whichever way as there will be configurations out there without the new value which might cause this issue to occur.

```powershell
Configuration ODSettings
{

    Node localhost
    {
        ODSettings 12345678-1234-1234-1234-123456789012
        {
            ApplicationId                             = '12345678-1234-1234-1234-123456789012'
            BlockMacSync                              = $False;
            CertificateThumbprint                     = '12345678-1234-1234-1234-123456789012'
            DisableReportProblemDialog                = $False;
            DomainGuids                               = @(); # This is the value which causes the issue
            Ensure                                    = "Present";
            ExcludedFileExtensions                    = @("pst", "mdb");
            IsSingleInstance                          = "Yes";
            NotificationsInOneDriveForBusinessEnabled = $True;
            NotifyOwnersWhenInvitationsAccepted       = $True;
            ODBAccessRequests                         = "Unspecified";
            ODBMembersCanShare                        = "Unspecified";
            OneDriveForGuestsEnabled                  = $False;
            OneDriveStorageQuota                      = 1048576;
            OrphanedPersonalSitesRetentionPeriod      = 30;
            TenantId                                  = 'sometenant.onmicrosoft.com'
        }
    }
}
```

Windows 10

dev

menswearUK:
Changes to the MSFT_ODSettings configuration to address #2021

I have added in parameter TenantRestrictionEnabled and this setting, if specified, takes precedence over the DomainGuids value

I've tried to get the module to behave the same way as changing the settings in the GUI would.  so if you disable TenantRestrictionEnabled it will also clear the value of DomainGuids.

I have also tried to ensure that existing configuration files won't fail if values are omitted.

If you omit the TenantRestrictionEnabled the module will set the value based on whether a valid DomainGuids entry is present flag up that it has set this based on other values.

If you omit DomainGuids, or if DomainGuids is not a valid value, the module will set TenantRestrictionEnabled to false to avoid invalid configurations being set and flag this up.

Fixes #2021

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_ODSettings/MSFT_ODSettings.schema.mof
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_ODSettings/MSFT_ODSettings.schema.mof
@@ -11,7 +11,7 @@ class MSFT_ODSettings : OMI_BaseResource
     [Write, Description("Lets administrators set policy on access requests and requests to share in OneDrive for Business"),ValueMap{"On","Off","Unspecified"},Values{"On","Off","Unspecified"}] String ODBAccessRequests;
     [Write, Description("Block sync client on Mac")] Boolean BlockMacSync;
     [Write, Description("Disable dialog box")] Boolean DisableReportProblemDialog;
-    [Write, Description("enable/disable Safe domain List - if disabled overrides DomainGuids value")] Boolean TenantRestrictionEnabled;
+    [Write, Description("Enable/disable Safe domain List - if disabled overrides DomainGuids value")] Boolean TenantRestrictionEnabled;
     [Write, Description("Safe domain list")] String DomainGuids[];
     [Write, Description("Exclude files from being synced to OneDrive")] String ExcludedFileExtensions[];
     [Write, Description("Groove block options"),ValueMap{"OptOut","HardOptIn","SoftOptIn"},Values{"OptOut","HardOptIn","SoftOptIn"}] String GrooveBlockOption;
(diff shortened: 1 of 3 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_ODSettings/MSFT_ODSettings.schema.mof
```diff
@@ -11,6 +11,7 @@ class MSFT_ODSettings : OMI_BaseResource
     [Write, Description("Lets administrators set policy on access requests and requests to share in OneDrive for Business"),ValueMap{"On","Off","Unspecified"},Values{"On","Off","Unspecified"}] String ODBAccessRequests;
     [Write, Description("Block sync client on Mac")] Boolean BlockMacSync;
     [Write, Description("Disable dialog box")] Boolean DisableReportProblemDialog;
+    [Write, Description("enable/disable Safe domain List - if disabled overrides DomainGuids value")] Boolean TenantRestrictionEnabled;
```
NikCharlebois, line 14: This drives documentation. Can we put a capital 'E' to Enable.

NikCharlebois (changes requested):
Thanks for submitting this! Just a few minor items to address. Thanks

Modules/Microsoft365DSC/DSCResources/MSFT_ODSettings/MSFT_ODSettings.schema.mof
menswearUK: No worries, I've amended that and also corrected a verbose message I'd altered while I was testing.

NikCharlebois (approved):

PR #2043, opened 2022-06-29
Remove None from SPOSharingSetting DefaultLinkPermissions

### LINKED ISSUE #2034: SPOSharingSettings: DefaultLinkPermission has wrong ValidateSet
Setting the SPOSharingSettings Perameter `DefaultLinkPermission` to `None` will result in an error. Currently, there is only View or Edit supported.

```
Default link permission value must be Edit or View 
+ CategoryInfo : WriteError: (:) [], CimException 
+ FullyQualifiedErrorId : EXCEPTION,PnP.PowerShell.Commands.Admin.SetTenant + PSComputerName : localhost
```

Fix the Validate Set to only allow Edit or View.

desmay:
Fixes #2034

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_SPOSharingSettings/MSFT_SPOSharingSettings.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_SPOSharingSettings/MSFT_SPOSharingSettings.psm1
@@ -93,7 +93,7 @@ function Get-TargetResource
 
         [Parameter()]
         [System.String]
-        [ValidateSet("None", "View", "Edit")]
+        [ValidateSet("View", "Edit")]
         $DefaultLinkPermission,
 
         [Parameter()]
@@ -143,7 +143,7 @@ function Get-TargetResource
 
     #region Telemetry
     $ResourceName = $MyInvocation.MyCommand.ModuleName -replace "MSFT_", ""
-    $CommandName  = $MyInvocation.MyCommand
+    $CommandName = $MyInvocation.MyCommand
     $data = Format-M365DSCTelemetryParameters -ResourceName $ResourceName `
         -CommandName $CommandName `
         -Parameters $PSBoundParameters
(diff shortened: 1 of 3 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_SPOSharingSettings/MSFT_SPOSharingSettings.psm1
```diff
@@ -93,7 +93,7 @@ function Get-TargetResource
 
         [Parameter()]
         [System.String]
-        [ValidateSet("None", "View", "Edit")]
+        [ValidateSet("View", "Edit")]
```
andikrueger, line 96: Wouldn't this be kind of a breaking change, as we remove one possible value (even though it did not work)?

NikCharlebois: Yes, agreed with Andi. Let's park it for the moment.

andikrueger: Actually, I think we should continue with this change, but only internally. We could keep the wrong validateset but still update the inner parts of the resource to handle NONE in the correct way.

desmay: So I'll restore None to validate set. I handle the None on get but will add logic on set to convert None to Edit

NikCharlebois: Can we also print a warning that the value is deprecated if anyone uses the value?

desmay: ok added a warning message, added None back to validateset and handled within app

desmay:
@NikCharlebois This should no longer be a breaking change.

NikCharlebois (approved):
Let's just make sure we remember to remove None from the allowed value for the first release of October (breaking changes). Thanks

PR #2052, opened 2022-06-30
AADSecurityDefaults

NikCharlebois:
Initial Release

N/A

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
@@ -171,12 +171,12 @@ function Set-TargetResource
 
     if ($CurrentValues.DisplayName -ne $DisplayName)
     {
-        Write-Warning -Message "The DisplayName property for the AADSecurityDefaults resource is read-only and cannot be changed."
+        Write-Verbose -Message "The DisplayName property for the AADSecurityDefaults resource is read-only and cannot be changed."
     }
 
     if ($CurrentValues.Description -ne $Description)
     {
-        Write-Warning -Message "The Description property for the AADSecurityDefaults resource is read-only and cannot be changed."
+        Write-Verbose -Message "The Description property for the AADSecurityDefaults resource is read-only and cannot be changed."
     }
 
     $UpdateParameters = @{
@@ -254,9 +254,6 @@ function Test-TargetResource
     Write-Verbose -Message "Target Values: $(Convert-M365DscHashtableToString -Hashtable $PSBoundParameters)"
 
     $ValuesToCheck = $PSBoundParameters
-    $ValuesToCheck.Remove('ApplicationId') | Out-Null
-    $ValuesToCheck.Remove('TenantId') | Out-Null
-    $ValuesToCheck.Remove('ApplicationSecret') | Out-Null
 
     $TestResult = Test-M365DSCParameterState -CurrentValues $CurrentValues `
         -Source $($MyInvocation.MyCommand.Source) `
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
@@ -1,0 +1,362 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        [ValidateSet('Yes')]
+        $IsSingleInstance,
+
+        [Parameter()]
+        [System.String]
+        $DisplayName,
+
+        [Parameter()]
+        [System.String]
+        $Description,
+
+        [Parameter()]
+        [System.Boolean]
+        $IsEnabled,
+
+        [Parameter()]
+        [ValidateSet('Present', 'Absent')]
+        [System.String]
+        $Ensure = 'Present',
+
+        [Parameter()]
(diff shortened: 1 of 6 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
```diff
@@ -0,0 +1,362 @@
...
+    try
+    {
+        $defaults = Get-MgPolicyIdentitySecurityDefaultEnforcementPolicy
+
+        $result = @{
+            IsSingleInstance      = 'Yes'
+            DisplayName           = $defaults.DisplayName
+            Description           = $defaults.Description
+            IsEnabled             = $defaults.IsEnabled
+            ApplicationId         = $ApplicationId
```
ykuijs, line 77: Not really specific to this resource, will for sure also apply to other resources, but should we return the authentication parameters as part of the Get method? I don't think this adds much value.

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
```diff
@@ -0,0 +1,362 @@
...
+    $CommandName  = $MyInvocation.MyCommand
+    $data = Format-M365DSCTelemetryParameters -ResourceName $ResourceName `
+        -CommandName $CommandName `
+        -Parameters $PSBoundParameters
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    if ($CurrentValues.DisplayName -ne $DisplayName)
+    {
+        Write-Warning -Message "The DisplayName property for the AADSecurityDefaults resource is read-only and cannot be changed."
```
ykuijs, line 174: Why add this as a parameter if it is read-only?

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    if ($CurrentValues.Description -ne $Description)
+    {
+        Write-Warning -Message "The Description property for the AADSecurityDefaults resource is read-only and cannot be changed."
```
ykuijs, line 179: Why add this as a parameter if it is read-only?

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
```diff
@@ -0,0 +1,362 @@
...
+    Write-Verbose -Message "Testing configuration of the Azure AD Security Defaults"
+
+    $CurrentValues = Get-TargetResource @PSBoundParameters
+
+    Write-Verbose -Message "Target Values: $(Convert-M365DscHashtableToString -Hashtable $PSBoundParameters)"
+
+    $ValuesToCheck = $PSBoundParameters
+    $ValuesToCheck.Remove('ApplicationId') | Out-Null
+    $ValuesToCheck.Remove('TenantId') | Out-Null
+    $ValuesToCheck.Remove('ApplicationSecret') | Out-Null
```
ykuijs, line 259: What about CertificateThumbprint and Credential?

Or better yet, shouldn't these parameters always be filtered in the Test-M365DSCParameterState function?

ykuijs (changes requested):
Also missing an entry in the changelog

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
NikCharlebois: Looks like they already are ;)

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
NikCharlebois: For reporting purposes, mostly. I know this is currently set as Single Instance, but my guts feeling tells me that there may soon be a way to define more then 1 security defaults. I am willing to change this is you believe it will add confusion to the users. I am not a fan of exposing properties that cannot be changed, but otherwise this resource because a single boolean resource

ykuijs: Then maybe not use Write-Warning, but Write-Verbose instead and add an "Informational Only" text or something.

To me, showing a warning message triggers a "Hey, something is wrong and I need to do something" responds. Where in this case, that is not necessary.

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
ykuijs: True, I noticed that last week as well 😄 In that case, these lines can be removed here. They only add noise.

ykuijs (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_AADSecurityDefaults/MSFT_AADSecurityDefaults.psm1
NikCharlebois: It does for the Export

ykuijs: You are absolutely right! Ignore this comment!

PR #2057, opened 2022-07-02
PR IntuneASRRulesPolicyWindows10ConfigMgr: Initial Release

William-Francillette:
IntuneASRRulesPolicyWindows10ConfigMgr: Initial Release

This is the initial release for this resource: all information are hard coded into the logic.
Further improvements (sent via email) can be implemented to ease setting catalog based policies. I'm happy for this resource to be on standby until feedback received for the proposed improvements

Thanks,
William


William-Francillette:
> I know it will be a mouthful, but can we please spell Manager completely in the resource name. Also, how does that name align with the underlying cmdlets? If we had been generating it with the DRG, what name would that resource have been given?


There is no cmdlet at the moment really - The legacy setting template use deviceManagementIntent but that's getting replaced by policies using the setting catalog; as for the new policies they don't have a cmdlet yet and use the REST query:
`https://graph.microsoft.com/beta/deviceManagement/configurationPolicies`
Those are generic and won't work with the DRG because they are based on templates for each type of policies (bitlocker, Antivirus, Exploit protection, EDR, Firewall and so forth). They can be found in Intune > Endpoint Security. This specific policy is called in Intune: Attack Surface Reduction Rules (ConfigMgr)
image
 
I'm planning to get a resource for all the Endpoint security policies available
image
Let me know, how should I rename this resource and if I should keep Intune name as reference.

Thanks

NikCharlebois:
Thanks for the info. Can we just rename the components to be Intune AttackSurfaceReductionPolicyWindows10 instead?

William-Francillette:
> Thanks for the info. Can we just rename the components to be Intune AttackSurfaceReductionPolicyWindows10 instead?

Hi @Nik,

The ASR rules comes in 3 different flavors of policies in Intune and I have developed a separate resource for each of those:
- IntuneASRRulesPolicyWindows10: legacy policy using deviceManagementIntent cmdlets (new policy can't be created on Intune UI anymore)
- IntuneSettingCatalogASRRulesPolicyWindows10: only for Intune managed devices
- IntuneASRRulesPolicyWindows10configMgr: only for ConfigMgr managed devices

What if I merge those 3 resources into a single IntuneAttackSurfaceReductionRulesPolicyWindows10?
They use different cmdlets and attributes but are used to deploy the same feature so will be probably less confusing and more complete.

If that sounds better then you can cancel this PR and I will be working on IntuneAttackSurfaceReductionRulesPolicyWindows10.

NikCharlebois:
I would rather like to keep them as separate resources. Thanks

William-Francillette:
> 
Hi,

haha I just seen that but completed the merged resource before 😁
If you want to have a look, https://github.com/William-Francillette/Microsoft365DSC-1/tree/IntuneAttackSurfaceReductionRulesPolicyWindows10
It exports the policy independently of the technology used, I simply added the technology as a parameters ("mdm","mdm,microsoftSense","configManager")

In the meantime, I will rename this resource now
Thanks
Will

NikCharlebois:
@William-Francillette what should we do with this PR? Any updates on renaming the files? Thanks

William-Francillette:
> @William-Francillette what should we do with this PR? Any updates on renaming the files? Thanks

Hi @NikCharlebois 

That's ready - The resource was renamed 

Thanks,
Will

PR #2108, opened 2022-07-19
AADAuthorizationPolicy - fixed update-issue raised in #2016

### LINKED ISSUE #2016: AADAuthorizationPolicy: New resource proposal
### Description
As described in https://docs.microsoft.com/en-us/graph/api/resources/authorizationpolicy?view=graph-rest-beta, the Authorization Policy provides a way to configure a number of basic tenant-properties that otherwise can only be managed in the portal which is what I'm trying to avoid (aren't we all...)

### Proposed properties
IsSingleInstance,
DisplayName,
Description,
AllowedToSignUpEmailBasedSubscriptions,
AllowedToUseSSPR,
AllowEmailVerifiedUsersToJoinOrganization,
AllowInvitesFrom,
BlockMsolPowerShell,
DefaultUserRoleAllowedToCreateApps,
DefaultUserRoleAllowedToCreateSecurityGroups,
DefaultUserRoleAllowedToReadOtherUsers,
DefaultUserRolePermissionGrantPoliciesAssigned,
GuestUserRole

(The DefaultUserRole... params are for the hash-table represented by defaultUserRolePermissions)
DisplayName and Description are optional as the policy is single-instance.

At the moment, the Graph SDK doesn't contain any cmdlets for retrieving and/or modifying /policies/authorizationPolicy/authorizationPolicy. Invoke-MgGraphRequest works quite well, though.
I have most of the code ready in a forked repo but am struggling a bit with the unit-test.
I haven't found a clear way to devise tests that include mocks for Invoke-MgGraphRequest given that it is used both to GET the resource and to PATCH it. It may be that my less than intimate relationship with Pester is showing here.
It's possible that Mock can be used with -ParameterFilter to separate the two uses for unit-testing. Any input is welcome.

salbeck-sit:


salbeck-sit:
Hm. This time I explicitly imported dependency-modules prior to running tests and now tests pass when I run Invoke-TestHarness in my environment.

salbeck-sit:
refactored AADAUthenticationPolicy to use Graph-profile v1.0 instead of beta

salbeck-sit:
quirky change caused by Get-MgPolicyAuthorizationPolicy returns DefaultUserRolePermissions with param PermissionGrantPolicyIdsAssigned but Update-MgPolicyAuthorizationPolicy expects DefaultuserRolePermissions to contain param PermissionGrantPoliciesAssigned. I'm starting to regret going back to the v1.0 profile

NikCharlebois (approved):
LGTM

PR #2118, opened 2022-07-20
EXOCASMailboxSettings: Fix issue if there are mailboxes with the same name

### LINKED ISSUE #2117: EXOCASMailboxSettings fails if there are mailboxes with the same name
I have multiple email accounts with the same mailbox "**name**", but in different domains.  The Export-TargetResource function in 'MSFT_EXOCASMailboxSettings.psm1' file fails when it tries to export one of these mailboxes.  The resulting $result (from Get-TargetResource) contains multiple values instead of a single value per parameter.  I suspect this issue could also impact updating or adding new resources with the same name but in different domains but I'm not sure.

Can you please change the value for the $mailboxName variable (under the Export-TargetResource function) from "$mailbox.Name" to  "$mailbox.UserPrincipalName".  The Name value is not a unique identifier for a mailbox in EXO and shouldn't be used.

N/A

N/A
'1.22.713.1'

malauter:
Fixed issue if there are mailboxes with the same name

- Fixes #2117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # UNRELEASED
 
+* EXOCASMailboxSettings
+  * Fixed issue if there are mailboxes with the same name
+  FIXES [#2117](https://github.com/microsoft/Microsoft365DSC/issues/2117)
 * AADGroup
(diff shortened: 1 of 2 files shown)


NikCharlebois (approved):
LGTM

Issue #2074, opened 2022-07-07
Issues due to multiple versions of Microsoft 365 DSC in regards to reporting


andikrueger:
At the moment there are various issues that are caused by the presence of multiple versions of M365 DSC on the local system. These issues can be resolved by uninstalling the versions not needed.

The following command resolves most of the issues:

```powershell
Update-Module Microsoft365DSC
Update-M365DSCDependencies
Uninstall-M365DSCOutdatedDependencies
```

I would suggest to introduce a new method `Test-M365DSCUpdateStatus` to verify if, there are updates available and/or multiple versions present on the local system. This function should be called within the repot and documentation functions and show warnings, if there issues detected.

t0w0ru:
so what you're suggesting is a method that just checks for multiple installations of the m365dsc module, checks if theres multiple versions installed (e.g by running 
```ps
if ($null -ne (Get-Module -Name Microsoft365DSC -List)[1]) { <# uninstall outdated versions by comparing to psgallery version #> }
```
and the integrated method Test-M365DSCDependencies?

andikrueger:
I would go with a function like this (sorry, I’m writing this on my mobile; PseudoCode)

```
$psgalleryversion = Find-module Microsoft365DSC
$localversion = Get-Module Microsoft365DSC

if($psgallery.Version -gt $localversion.Version)
Write-Host “there is a newer version of M365DSC. please run the three commands above”)

```

t0w0ru:
I'll go with your approach, changing my pr in a bit. appreciate the input :)

t0w0ru:
I've just pushed an update to my [PR](https://github.com/microsoft/Microsoft365DSC/pull/2120)
The Uninstall-M365DSCOutdatedDependencies cmdlet still doesn't check for multiple installations of the latest version,
but it does for the Microsoft365DSC module now :)
I hope this is closer to what you imagined the extra cmdlet to do

andikrueger:
Thanks. This is Beth close to what I had in mind. 

It would be great to include this function in  reporting and export cmdLets. This would solve many issues due to outdated modules. 
I would go without prompting the user for input in both scenarios as this might break pipelines or automation.

t0w0ru:
updated again :D

andikrueger:
Reopening this issue, as there is one part missing:

All Reporting functions within M365 should call Test-M365ModuleValidity and display a warning message.

PR #2120, opened 2022-07-20
Add function Test-M365DSCModuleValidity

t0w0ru:
This cmdlet checks whether or not the Microsoft365DSC Module is up2date and suggests the user to update the module and its dependencies if it isnt.

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -2342,7 +2342,6 @@ function Test-M365DSCDependenciesForNewVersions
     $i = 1
     Import-Module PowerShellGet -Force
 
-    $OutdatedModules = [System.Collections.ArrayList]@()
     foreach ($dependency in $dependencies)
     {
         Write-Progress -Activity "Scanning Dependencies" -PercentComplete ($i / $dependencies.Count * 100)
@@ -2358,11 +2357,6 @@ function Test-M365DSCDependenciesForNewVersions
             if (-not $modules -or [Version]($moduleInGallery.Version) -gt [Version]($moduleInstalled[0].Version))
             {
                 Write-Host "New version of {$($dependency.ModuleName)} is available {$($moduleInGallery.Version)}"
-                $DependencyObject = [PSCustomObject]@{
-                    ModuleName    = $($dependency.ModuleName)
-                    ModuleVersion = $($moduleInGallery.Version)
-                }
-                $OutdatedModules.Add($DependencyObject)
             }
         }
         catch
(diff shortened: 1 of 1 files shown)


t0w0ru:
https://github.com/microsoft/Microsoft365DSC/issues/2074

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3482,6 +3490,60 @@ function New-M365DSCMissingResourcesExample
...
+    param(
+        [parameter()]
+        [Switch]
+        $Force
+    )
+    $InformationPreference = 'Continue'
+
+    # validate only one installation of the module is present (and it's the latest version available from the psgallery)
+    $latestVersion = (Find-Module -Name 'Microsoft365DSC').Version
+    Install-Module -Name 'Microsoft365DSC' -Force -AllowClobber -RequiredVersion $latestVersion
```
andikrueger, line 3524: I would not install a new version of M365DSC within a method called Test- … it would be sufficient to check for newer versions of m365dsc within psgallery. All dependency updates do require a new release of m365dsc.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3482,6 +3490,60 @@ function New-M365DSCMissingResourcesExample
...
+
+    [array]$modules = Get-Module -Name 'Microsoft365DSC' -List | Sort-Object Version -Descending
```
andikrueger, line 3526: Uninstall-M365DSCOutdatedDependencies already uninstalls previous versions of M365DSC. There is no need to implement it here.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
t0w0ru: i absolutely agree, just forgot to remove it.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
t0w0ru: you're not wrong, but also not 100% correct.
the cmdlet does uninstall outdated versions of the m365dsc module - as long as theyre in the same psroot path youre executing the cmdlet from. Say you have multiple installations of powershell, e.g 7.1 and 5.2 in different locations, the following is gonna happen: [current module path](https://i.imgur.com/zklmZ0J.png), [All module paths](https://i.imgur.com/tfl5PD0.png)

andikrueger: Fair point on that one. Then we should implement it within the uninstall-… function.

andikrueger:
Please have a look at my comment within #2074

NikCharlebois (approved):
LGTM

PR #2136, opened 2022-07-23
PR: DRG - Added support for EnumType + other fixes

William-Francillette:

This PR adds support to the DRG for EnumType  to automatically generate

- In the module file: ValidateSet()
- In the schema file: Values and ValueMap

Added 2 parameters to New-M365DSCResource

- ParametersToSkip: Parameters to skip during the resource generation - typically used to remove read-only/unwanted parameters 
- AdditionalPropertiesType: Add the additional property type instead of selecting it manually while the resource is generated

Also minor fixes in page layout


NikCharlebois (approved):
LGTM

PR #2145, opened 2022-07-27
DSC Resource property report

malauter:
This function checks if properties of existing resources are up to date.
Creates a report about missing or outdated properties of existing resources and a list of missing resources.

@NikCharlebois as discussed, please do a short code review of this initial draft


NikCharlebois (approved):
LGTM

PR #2161, opened 2022-07-29
Issue 2158

### LINKED ISSUE #2158: Refactoring: reporting functions & Compare-M365DSCConfigurations: should call Test-ModuleValidity and remove of Version should be improved.
Creating reports caused many errors over the past few weeks. 

There were several improvements like #2142 and #1970 
At the moment, each function implements a slightly different interpretation on how to remove  the version from the configuration.
The latest addition of Test-ModuleValidity adds more complexity to maintain the code. 

It would be great to create a function that is used bz all reporting functions. Keeping changes in sync would be more easier.

andikrueger:
Refactoring of the reporting functions

Fixes #2158

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
@@ -1069,7 +1069,7 @@ function Initialize-M365DSCReporting
         Write-Verbose "Error trying to remove Module Version"
     }
 
-    if ($IncludeComment)
+    if ($IncludeComments)
     {
         $parsedContent = ConvertTo-DSCObject -Content $fileContent -IncludeComments:$True
     }
--- a/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
@@ -1061,8 +1061,11 @@ function Initialize-M365DSCReporting
     try
     {
         $startPosition = $fileContent.IndexOf(" -ModuleVersion")
-        $endPosition = $fileContent.IndexOf("`r", $startPosition)
-        $fileContent = $fileContent.Remove($startPosition, $endPosition - $startPosition)
+        if ($startPosition -gt 0)
+        {
+            $endPosition = $fileContent.IndexOf("`r", $startPosition)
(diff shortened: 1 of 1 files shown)


Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
```diff
@@ -22,30 +18,16 @@ function New-M365DSCConfigurationToHTML
...
+        [Parameter()]
+        [System.String]
+        $TemplateName,
+
         [Parameter()]
         [Switch]
         $SortProperties
     )
 
-    if ([System.String]::IsNullOrEmpty($ParsedContent))
```
NikCharlebois, line 30: Why are we trying to remove this? This is required for other scenarios that feed in the content for the report instead of a file. For example, when running from a serverless Azure function

Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
```diff
@@ -10,10 +10,6 @@ function New-M365DSCConfigurationToHTML
     [CmdletBinding()]
     [OutputType([System.String])]
     Param(
-        [Parameter()]
-        [System.String]
-        $ConfigurationPath,
```
NikCharlebois, line 15: Why are we removing the ConfigurationPath here?

NikCharlebois (changes requested):
Why are we changing the way we allow for content to be passed in as either a file or as parsedContent. We need to continue to enable both scenarios.

andikrueger:
Those functions are internal and should not be called directly IMHO. All three functions are called by ‘ New-M365DSCReportFromConfiguration’. This function handles the changes above. Therefore there is no need to expose these parameters at the function level.

This change reduces a lot of duplicated code.

Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
andikrueger: See comment below.

Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
andikrueger: See comment below.

NikCharlebois (approved):

PR #2164, opened 2022-08-01
Unconditionally remove ReplyURLs (and friends) from $currentParameters

serial666:
Whenever an application is both present in the tenant and in the DSC side then "Update-MgApplication @currentParameters" is always called but if $ReplyURLs is empty then it errors out because it will try to use it as parameter for Update-MgApplication which doesn't exist. To fix this just remove $ReplyURLs and friends from $currentParameters. Note that this same problem will happen if app is not present and needs to be recreated with New-MgApplication.

Error below:

VERBOSE: [redacted]:[[AADApplication][redacted]] Updating existing AzureAD Application {[redacted]} with values:

Key            Value
---            -----
DisplayName    [redacted]
IdentifierUris {}
ApplicationId  [redacted]
ReplyURLs      {}
SignInAudience AzureADMyOrg
Verbose        True

A parameter cannot be found that matches parameter name 'ReplyURLs'.
    [+] CategoryInfo          : InvalidArgument: (:) [], CimException
    [+] FullyQualifiedErrorId : NamedParameterNotFound,Update-MgApplication
    [+] PSComputerName        : localhost

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+* AADApplication
+  * Fixed issue where Update-MgApplication could be called with parameter $ReplyURLs which is invalid.
 * AADConditionalAccessPolicy
   * DEPRECATED then IncludeDevices and ExcludeDevices parameters.
   * Fixed issue extracting a policy that had invalid users or groups (deleted from AAD).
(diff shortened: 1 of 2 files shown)


ykuijs (changes requested):
Can you please add an entry with your changes to the Changelog, under the Unreleased section?

NikCharlebois (approved):

NikCharlebois (approved):

NikCharlebois (approved):

PR #2165, opened 2022-08-01
Enable export/import of application owners

serial666:
This enables exporting and importing app owners, filtering out those which have their accounts deleted. Copied almost verbatim from MSFT_AADGroup.{psm1,schema.mof}.

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
@@ -53,6 +53,10 @@ function Get-TargetResource
         $ReplyURLs,
 
         [Parameter()]
+        [System.String[]]
+        $Owners,
+
+        [Parameter()]
         [Microsoft.Management.Infrastructure.CimInstance[]]
         $Permissions,
 
@@ -148,6 +152,17 @@ function Get-TargetResource
                 $currentOauth2RequirePostResponseValue = $false
             }
 
+            [Array]$Owners = Get-MgApplicationOwner -ApplicationId $AADApp.Id -All:$true | `
+                ?{ !$_.DeletedDateTime }
+            $OwnersValues = @()
+            foreach ($Owner in $Owners)
+            {
+                if ($Owner.AdditionalProperties.userPrincipalName)
+                {
+                    $OwnersValues += $Owner.AdditionalProperties.userPrincipalName
+                }
+            }
+
             $result = @{
                 DisplayName                = $AADApp.DisplayName
                 AvailableToOtherTenants    = $AvailableToOtherTenantsValue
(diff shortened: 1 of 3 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
```diff
@@ -386,6 +409,44 @@ function Set-TargetResource
...
+        }
+        if (!$backCurrentOwners)
+        {
+            $backCurrentOwners = @()
+        }
+        $ownersDiff = Compare-Object -ReferenceObject $backCurrentOwners -DifferenceObject $desiredOwnersValue
+        foreach ($diff in $ownersDiff)
+        {
+            if ($diff.InputObject.Contains('@')) {
+                $OwnerId = $(Get-MgUser -UserId $diff.InputObject).Id
```
andikrueger, line 426: Please have a look at this recent change in another resource:  
https://github.com/microso95ft/Microsoft365DSC/blob/17984cf89f4c13c591e1b935564e060f2be81273/Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1#L256

The error action was introduced due to some issues with non-existing users.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.schema.mof
```diff
@@ -19,9 +19,10 @@ class MSFT_AADApplication : OMI_BaseResource
     [Write, Description("User-defined URI(s) that uniquely identify a Web application within its Azure AD tenant, or within a verified custom domain.")] string IdentifierUris[];
     [Write, Description("Client applications that are tied to this resource application.")] string KnownClientApplications[];
     [Write, Description("The logout url for this application.")] string LogoutURL;
-    [Write, Description("Set this to true if an Oauth2 psot response is required.")] Boolean Oauth2RequirePostResponse;
+    [Write, Description("Set this to true if an Oauth2 postt response is required.")] Boolean Oauth2RequirePostResponse;
```
andikrueger, line 22: typo post not postt

serial666: fixed 👍

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
serial666: I always use exceptions whenever possible, but just started looking at this project, for which I'll need in my $DAYJOB, and didn't want to drift too much of the code already in place so I basically started with a copy from the AADGroup resource. That being said in the new commit it now checks if objects exist enclosed in exceptions before proceeding, but that being said it shows that AADGroup also has the same problem. If this one now is OK I can apply the same fix there.

andikrueger: Good catch. Handling errors would be the preferred solution :-)

serial666: I was on the fence on that one, but I agree it should stop to give the admin the chance to inspect what went wrong in the first place, I've updated it.

andikrueger:
Could you update the change log with the changes you made?

NikCharlebois (approved):

PR #2167, opened 2022-08-01
Don't call Get-MgSubscribedSku on 2 loops

serial666:
Optimize 2 loops by not having to call Get-MgSubscribedSku on every single license assignment user has to be added and/or removed, just call it once out of those loops.


andikrueger:
Could you update the change log with the changes you made?

NikCharlebois (approved):

PR #2169, opened 2022-08-01
Dedup some code and fix typo in variable on O365SiteGroup

serial666:
Avoid redefining $SiteGroupSettings always to the same value, just define it once, and call it as is on Set-PnPGroup, except if it's to be used on remove the SiteGroup, in this case just remove the Owner key/value. Additionally to keep the order of setting the group owner first, and then the group permissions add a new object ($GroupPermissionsParameters) to keep track of which parameters are required and then finally call Set-PnPGroup/Set-PnPGroupPermissions in the right order. Note that if the order of calling both cmdlets is not important then of course the diff would be even smaller since the calls to Set-PnPGroupPermissions could be kept as is.

While here fix typo in variable, currently this is only a nuisance since even if Identity and Owner of SiteGroup are the same as $currentValues it tries to update the group anyway to those same values, but if not noticed and code inside condition is altered then it could cause problems, so make it clear that if they are the same then calling Set-PnPGroup is not required and it can just be skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   * DEPRECATED then IncludeDevices and ExcludeDevices parameters.
   * Fixed issue extracting a policy that had invalid users or groups (deleted from AAD).
   FIXES [#2151](https://github.com/microsoft/Microsoft365DSC/issues/2151)
+* SPOSiteGroup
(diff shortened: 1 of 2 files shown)


ykuijs (changes requested):
Can you please also add a description for this change to the Changelog, under the Unreleased section?

NikCharlebois (approved):

PR #2207, opened 2022-08-23
Remove support for importing/exporting app owners for now since it's not working

serial666:
Remove support for importing/exporting application owners on AADApplication resource, at least for now, since currently the backend is failing with "The request is currently not supported" on both adding and removing owners.

Not only the cmdlet New-MgApplicationOwnerByRef lost the param BodyParameter, the whole backend is responding "The request is currently not supported" with anything app owner related even manually sending the requests towards Graph.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 * AADApplication
   * Fixed issue where Update-MgApplication could be called with parameter ReplyURLs which is invalid.
-  * Added support to export/import app owners.
 * EXOTransportRule
   * Fix issue setting IncidentReportContent
   FIXES [#2196](https://github.com/microsoft/Microsoft365DSC/issues/2196)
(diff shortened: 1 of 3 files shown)


NikCharlebois:
@peombwa @maisarissi another related issue with 1.11.0 it seems

NikCharlebois:
Let's not implement regression hacks. If the problem is with v1.11.0, then we simply shouldn't upgrade to that version. We need to conduct additional investigation internally first.

serial666:
Right! FYI just confirmed that downgrading to 1.10.0 I'm able to manage app owners through {New,Remove}-MgApplicationOwnerByRef and Invoke-GraphRequest without issues whereas with 1.11.0 they always fail with "The request is currently not supported" so it's definitely a regression on this new version.

peombwa:
@NikCharlebois, the issue has been fixed in `1.11.1`. Please upgrade to the latest version of the module. Related to https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1129.

serial666:
@NikCharlebois Tested it right now and as @peombwa mentioned this issue is fixed with 1.11.1, BodyParameter is back and the cmdlets work as expected, as soon as the dependencies on DSC are updated this PR can be closed. Thank you for the quick fix and sorry to jump the gun so quickly.

andikrueger:
The dependencies are going to be update with this PR #2211. This will push the MS.Graph.* Modules to V 1.11.1

PR #2213, opened 2022-08-24
Avoid using Get-MgDirectoryObject cmdlet

serial666:
Do not use Get-MgDirectoryObject cmdlet since it's not used anywhere else in M365DSC and it would need to bring the dependency on Microsoft.Graph.DirectoryObjects just to call it once here.


serial666:
This one slipped under the rug undetected since I had all Microsoft.Graph.* modules installed and not just M365DSC's dependencies where I'd be able to see the issue.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
```diff
@@ -471,7 +471,7 @@ function Set-TargetResource
                 } else {
                     $OwnerId = $diff.InputObject
                     $DirectoryObjectUri += "{0}" -f $OwnerId
-                    Get-MgDirectoryObject -DirectoryObjectId $DirectoryObjectUri -ErrorAction Stop
+                    Invoke-MgGraphRequest -Method GET -Uri $DirectoryObjectUri -ErrorAction Stop
```
NikCharlebois, line 474: This begs the question: Dom we even need this call in here? We are not even assigning it to any variables.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
serial666: Yes we need, it's behind a try/catch block, if the object doesn't exist then it should stop, the return value isn't needed.

serial666: Not only that, we actually should do the same around a lot of cmdlets that could fail on other resources. Ie've seen alot of places without exceptions around.

serial666: To give a little bit more of context and clarify this, since we need to get the ID of the user with Get-MgUser if it doesn't exist then we must catch the error upfront and bail out, for a service principal on the other hand strictly speaking this call to Invoke-MgGraphRequest wouldn't be needed since we already have its ID, and there's also a try/catch further down the code when adding or removing the owner, but it would look weird to not catch the error upfront as well in the same place as for users.

serial666: After reading what I just wrote above I realized we can make the code much simpler, when adding users as owners we don't need their ID and only the UPN is required as long as we differentiate the corresponding Uri from when adding service principals. Then saw that I also missed a call to Remove-MgDirectoryObject that shouldn't be there and replaced it with Remove-MgApplicationOwnerByRef, for this specific case if it's a user we need to remove as an owner then we still need to call Get-MgUser to get its ID since unfortunately the previous mentioned cmdlet doesn't have an Uri as parameter and instead uses DirectoryObjectId, but at least there's no need for a separate try/catch block as in my initial diff.

NikCharlebois (approved):

PR #2234, opened 2022-08-26
fixes #2233

### LINKED ISSUE #2233: Export-M365DSCConfiguration within Azure Run Book fails due to hardcoded value of $AzureAutomation
When running `Export-M365DSCConfiguration` within Azure Run Books, there is an error about the call to the local configuration manager. This should not happen, as there is a already code to prevent this call. Anyhow, the variable used, to differentiate Azure Automation form other scenarios is hardcoded and set to $false

https://github.com/microsoft/Microsoft365DSC/blob/4dcf59d529e5e576fd136754d7693b12151b8145/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1#L277

This variable gets no other assignment afterwards. 

This assignment should be change to something like this:

```powershell
$AzureAutomation = $false 
if ("AzureAutomation/" -eq $env:AZUREPS_HOST_ENVIRONMENT) {
   $AzureAutomation = $true
}
```
see: https://stackoverflow.com/questions/37224874/what-is-the-correct-way-to-detect-whether-a-script-is-running-in-azure-automatio

andikrueger:
This PR will add a check if the code is run within Azure Automation or locally
- Fixes #2233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 * IntuneDeviceConfigurationPolicyAndroidOpenSourceProject
   * Don't export all policies if none match the type
   FIXES [#2228](https://github.com/microsoft/Microsoft365DSC/issues/2228)
+* MISC
+  * Fix issue of running Export-M365DSCConfiguration within Azure Run Book. FIXES [#2233](https://github.com/microsoft/Microsoft365DSC/issues/2233)
 
 # 1.22.824.1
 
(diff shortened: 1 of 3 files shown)


ykuijs:
@andikrueger Can you also update the changelog?

andikrueger:
Totally forgot about the change log. Added the change right now.

NikCharlebois (approved):

PR #2284, opened 2022-09-14
Fixes #2271

### LINKED ISSUE #2271: TeamsCallingPolicy: Export causes error with validateset for BusyOnBusyEnabledType Param
Running an export of TeamsCallingPolicies causes an issue with the configuration. The BusyOnBusyEnabledType param gets a value that is not part of the params validateset.

image

Fix the validateset to support this addtional value or add a output transformation to include proper values for this resource and the powershell cmdlets used.

This is part of the docs for the [`Set-CSTeamsCallingPolicy`](https://docs.microsoft.com/en-us/powershell/module/skype/New-CsTeamsCallingPolicy?view=skype-ps):

-BusyOnBusyEnabledType
Setting this parameter lets you configure how incoming calls are handled when a user is already in a call or conference or has a call placed on hold. Valid options are: Enabled, Unanswered, and Disabled. When set to Enabled, new or incoming calls will be rejected with a busy signal. When set to Unanswered, the user's unanswered settings will take effect, such as routing to voicemail or forwarding to another user. **Note: UserOverride option value is not available for use currently, if set it will be read as setting the value to Disabled**

1.22.831.1

NikCharlebois:
* TeamsCallingPolicy
  * Added UserOverride as an accepted value for the BusyOnBusyEnabledType parameter.

* Fixes #2271


andikrueger:
LGTM

PR #2295, opened 2022-09-16
Aad conditional access policy custom controls - Fix #2292

### LINKED ISSUE #2292: AADConditionalAccessPolicy: Ehancement request - Add support for customAuthenticationFactors
We use a 3rd party MFA solution (symantec VIP) configured as a [custom control](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/controls).  When using the module to export my config i noticed that my polices are being exported with no Grant controls, as it doesn't appear to support custom controls.

While technically custom controls ARE still a preview feature, the config for them is included in both the beta and v1.0 graph api for conditional access policies `https://graph.microsoft.com/v1.0/identity/conditionalAccess/policies` and can be set/retrieved with the string collection attribute `customAuthenticationFactors` as shown in the [conditionalAccessGrantControls resource type](https://docs.microsoft.com/en-us/graph/api/resources/conditionalaccessgrantcontrols?view=graph-rest-1.0) properties.  

As this feature has been in preview for several years now, and [microsoft's hinted-at replacement](https://techcommunity.microsoft.com/t5/microsoft-entra-azure-ad-blog/upcoming-changes-to-custom-controls/ba-p/1144696) has yet to materialize after over 2 years, it would greatly enhance the usability of the module if we could set/retrieve this attribute as part of the AADConditionalAccessPolicy resource, even if the module doesn't support creation of the control resource itself since that is still preview.

A CustomAuthenticationFactors or CustomControls attribute could be added to the AADConditionalAccessPolicy resource  that accepts a string array as input, which maps to the array of IDs returned by the customAuthenticationFactors attribute in the conditionalAccess/policies API.

Export-M365DSCConfiguration -Components @( "AADConditionalAccessPolicy") -Credential $Credential

n/a

1.22.914.1

garretth9:
Add ability to Get/Set custom controls for Conditional Access Policies.

Fixes #2292

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1
@@ -688,7 +688,7 @@ function Get-TargetResource
             #no translation or conversion needed
             BuiltInControls                          = [System.String[]](@() + $Policy.GrantControls.BuiltInControls)
             #no translation or conversion needed
-            CustomControls                          = [System.String[]](@() + $Policy.GrantControls.CustomControls)
+            CustomControls                          = [System.String[]](@() + $Policy.GrantControls.CustomAuthenticationFactors)
             #no translation needed, return empty string array if undefined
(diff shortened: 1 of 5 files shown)


NikCharlebois:
@garretth9 looks like we may have a duplicate. I believe this has already been addressed with https://github.com/microsoft/Microsoft365DSC/pull/2297

garretth9:
> 

Yep, looks like that resolves the issue, I'll close this out.  Thanks for taking care of it!

PR #2280, opened 2022-09-13
IntuneDeviceConfigurationAndroidDeviceOwner: Fixes #2272

William-Francillette:
#### This Pull Request (PR) fixes the following issues

Resource rebuilt using the DRG + added descriptions
Removed applicability rules as not implemented in the other resources: Fix #2272

Issue #2272, opened 2022-09-12
IntuneDeviceConfigurationPolicyAndroidDeviceOwner - Export results in wrong configuration.


andikrueger:
#### Details of the scenario you tried and the problem that is occurring
The export of MSFT_IntuneDeviceConfigurationPolicyAndroidDeviceOwner results in a configuration that is not valid. The params:

- DeviceManagementApplicabilityRuleDeviceMode
- DeviceManagementApplicabilityRuleOsEdition
- DeviceManagementApplicabilityRuleOsVersion

Are not written properly. 

![image](https://user-images.githubusercontent.com/15608729/189647464-a64cde7c-ba5b-4929-87ff-f44d065f2638.png)

#### Suggested solution to the issue
Create proper instances of the CIM objects within the configuration.

#### Version of the DSC module that was used ('dev' if using current dev branch)
1.22.831.1

William-Francillette:
This resource was probably generated a while ago with one of the first version of the DRG as it is very different from the other one.
The applicability rules haven't been deployed to any resources yet.
I would suggest to generate that resource again and add support for assignments and descriptions.
Not sure about the credentials issue as I use certificate authentication
I'll be happy to work on that next

andikrueger:
Yesterday I was working on the Credential issue and I was not able to find the root cause for this one. The handling of credential is done by one method and works for all other resources.

I think, we can skip this part for the moment and should focus on fixing the other params first.

Thanks for picking this up!

PR #2306, opened 2022-09-20
Add ManagedIdentity support to resource generator

mlhickey:
Add ManagedIdentity switch support to resource generator.

None


andikrueger:
@mlhickey Just out of curiosity: what kind of code editor or code formatting rules are you using? The formatting looks good. On the other hand, it changes the quoting styles from double quotes to single quotes.

@NikCharlebois Do we have a preferred style for quotes?

mlhickey:
Everything is done in VS Code with PS extension - Allman for style,  standard is to use single quotes for non-evaluated strings.  I’ve noticed it seems to vary across resources.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

andikrueger:
vscode should respect the settings file within the .vscode folder… it’s interesting, as I did not see a changes of quote styles on my end in the past.

mlhickey:
I agree, purpose behind that folder but...

andikrueger:
Good point. Looks like we need to implement this change within the .vscode settings. This was introduced with V 1.19.1 and targets the PSScriptAnalyzer Rule: https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoidusingdoublequotesforconstantstring?view=ps-modules

Prior to implementing this change, we need to consider the impact.

mlhickey:
Q - should we consider offering use of Update- when New- isn't available/relevant for a given noun, e.g. AuthorizationPolicy?

andikrueger:
Authorization Policy is a singleton within the tenant. Therefore there is no New function available. (see: https://learn.microsoft.com/en-us/graph/api/resources/authorizationpolicy?view=graph-rest-1.0)

Updating policies would be valid case for M365DSC.  Is it worth the efforts to modify DRG for this case? Are there more objects within graph without a new method?

mlhickey:
Yup, just an example.  The effort is minimal, but if there are only a few unique then maybe it doesn't matter - if it isn't broken then ...

andikrueger:
DRG has a great benefit over all other approaches. It gives us the opportunity to build new and even rebuild existing resources on the fly. As soon as there are changes to the APIS, we can react in a fast way (e.g. like in #2272 / #2280)

I just went through the DRG code. The change would not be that big, as there is already code to change the default noun for the parameters:

https://github.com/microsoft/Microsoft365DSC/blob/b8624c55452713cf6a487513144186cfb9689770/ResourceGenerator/M365DSCResourceGenerator.psm1#L948-L952

and if I see it correctly, this line needs some updating too:

https://github.com/microsoft/Microsoft365DSC/blob/b8624c55452713cf6a487513144186cfb9689770/ResourceGenerator/M365DSCResourceGenerator.psm1#L1223

Which would result in a new parameter :D so that the change can be made.

Would be worth a try.

mlhickey:
Yup.  In my test I set a flag early around 1011:
`$hasNew = (-not [System.String]::IsNullOrEmpty($cmdletCommandDetails))`
then do conditional based on that around 1250 and 2095 (relative based on code addition)

NikCharlebois (approved):

Issue #2301, opened 2022-09-20
AADGroup: Add properties MemberOf and AssignedToRole


salbeck-sit:
### Description
For a group that is created with IsAssignableToRole, it would be helpful to also be able to assign it to one or more AAD roles and subsequently to maintain such role-assignments. Therefore, an additional property AssignedToRole is proposed

Also, it would be nice to be able to build and maintain security-groups that are intended as role-groups, ie members of a role-group are recursively members of one or more security-groups. Therefore, an additional property MemberOf is proposed

### Proposed properties
MemberOf
AssignedToRole

### Special considerations or limitations
I have already written the necessary addition to the DSC-resource (based off of an earlier version and retro-fitted to the current dev-version of the DSC-resource) and tested it against a test-tenant.

However, since I'm not really at home with Pester, I'm contemplating how unit-tests could - and/or should - be written to verify that the new functionality works as intended. It gets complicated quickly as there's a need for mocking additional groups and one or more directory roles, so it tends to become more similar to a kind of integration-test. Please advise..

salbeck-sit:
I think I have the unit-tests sorted by using -ParameterFilter in the Get-MgGroup mocks. Tests include whether a group is a member of another group and whether a group is assigned to the correct AAD-role. I'll run the test-harness and report back

salbeck-sit:
Unit-testing is now under wraps. It isn't pretty but the power of Pester really shows.

salbeck-sit:
PR #2316 created

salbeck-sit:
Closed

PR #2316, opened 2022-09-22
AADGroup: Implements #2301

salbeck-sit:
AADGroup: Add properties MemberOf and AssignedToRole to resource

Implements #2301


NikCharlebois:
@salbeck-sit can you please take a look at the failing Unit Tests? Thanks

NikCharlebois (changes requested):
Small typo. Can you also add an entry in the ChangeLog? Thanks

Modules/Microsoft365DSC/DSCResources/MSFT_AADGroup/MSFT_AADGroup.psm1
```diff
@@ -164,7 +172,7 @@ function Get-TargetResource
             if ($Group.MembershipRuleProcessingState -ne 'On')
             {
                 # Members
-                [Array]$members = Get-MgGroupMember -GroupId $Group.Id -All:$true
+                [Array]$members = Get-MgGroupMember -GroupId $Group.Id
```
NikCharlebois, line 175: Why are we removing the -All here? It should be included to make sure we catch all members

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_AADGroup/MSFT_AADGroup.psm1
salbeck-sit: -ALL is apparently no longer supported by Get-MgGroup* cmdlets in v1.12.0 of the Microsoft.Graph.Groups module. 
Running the unit-tests with -All parameters resulted in a number of 'caught throws' where the text in the corresponding M365DSC eventlog stated that "A parameter cannot be found that matches parameter name 'All'". The only cmdlets in the Get-TargetResource function that used -All were the Get-MgGroup* cmdlets. Removing -All everywhere resulted in a passed unit-test.

The question of whether this should have been brought to the attention of the folks responsible for Graph-modules is another matter. I guess you could have gotten them to update the relevant module in a day or two as something similar has happened before, but I really just wanted to get the change under wraps.

andikrueger: There was an intermediate release. Graph modules at up to 1.12.1 now and do allow the -all parameter

andikrueger: See https://github.com/microsoft/Microsoft365DSC/commit/e01d15b4d208a1fa666afd01f25c423323b87816

NikCharlebois (approved):

PR #2317, opened 2022-09-22
EXOMailContact - New Resource

### LINKED ISSUE #2279: Creating EXO Contacts
Is it possible to create EXO Contacts within M365DSC? I was not able to find any of it... Help very appreciated

NikCharlebois:
* EXOMailContact
  * Initial Release 

* Fixes #2279


malauter:
@NikCharlebois I think we should exclude the OrganizationalUnit parameter from testing in the Test-TargetResource function, if OrganizationalUnit parameter is empty, like we do e.g. in distribution group resource.
OrganizationalUnit is rarely set by an user, but it is never empty. Including this parameter in the test can result in false-positive drift alerts.

andikrueger:
- Fixes #2322

PR #2353, opened 2022-09-29
AADTenantDetails: Fixed issue with applicationsecret parameter being passed to Update-MGOrganization

### LINKED ISSUE #2340: AADTenantDetails: Can not set tenant details
I tried to update the "technical contact" in "AADTenantDetails". So I exported the resource "AADTenantDetails" with application secret, which should work for AzureAD settings. The export works fine and I could also generate the MOF-Files with different ConfigurationData, so I could test the import on two different tenants. On both tenants we have the app registration with application secret with the following permissions:
image

So if the tenant is in desired starte, the configuration runs without an error. If it is not in desired state and should set the "TechnicalNotificationMails" I get the information "Cannot Set AzureAD Tenant Details":
image

Do you know why I get this error? I checked with "Get-M365DSCCompiledPermissionList" to get the permissions which I set on the app registration.

Thank you for your help.

andikrueger:
Fixes #2340

AADTenantDetails: Fixed issue with applicationsecret parameter being passed on to Update-MGOrganization

- Fixes #2340

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * AADGroup
   * Added properties MemberOf and AssignedToRole
   Implements [#2301](https://github.com/microsoft/Microsoft365DSC/issues/2301)
+* AADTenantDetails
+  * Fixed an issue where ApplicationSecret was send to Update-MgOrganization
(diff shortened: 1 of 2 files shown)


NikCharlebois (approved):

mario-beeler:
Thank you very much, works now.

PR #2381, opened 2022-10-07
Fixes #2270

### LINKED ISSUE #2270: IntuneDeviceConfigurationPolicyAndroidDeviceOwner: Credential Parameter as String in Export
1. Export settings
The export of the resource does not generate a well formatted configuration. This configuration cannot be applied to a tenant. The parameter Credential is of type string rather than PSCredential. This breaks the configuration.

image

Change Credential Param in Exported configuration object.

1.22.831.1

William-Francillette:

Fixes #2270
Added  a space as described in the issue


NikCharlebois:
I believe we will be better served by doing this at the ReverseDSC level since this could happen with other resources in the future as well where a parameter would have 'Credential' as part of its name. Let me submit a PR and update ReverseDSC instead. Thanks for taking this on.

Issue #2204, opened 2022-08-23
[EXOAtpPolicyForO365] Using BlockUrls is giving a deprecated error


ykuijs:
#### Details of the scenario you tried and the problem that is occurring
When using the BlockUrls parameter of the EXOAtpPolicyForO365 resource, I am getting the below error. These blocked urls are now being managed via the Tenant Allow/Block List in [Microsoft Defender)[https://docs.microsoft.com/en-us/microsoft-365/security/office-365-security/manage-tenant-allow-block-list?view=o365-worldwide].

This line is giving the error:
https://github.com/microsoft/Microsoft365DSC/blob/d3bbe62f2f94212facbe4c0e84526b31075e40ad/Modules/Microsoft365DSC/DSCResources/MSFT_EXOAtpPolicyForO365/MSFT_EXOAtpPolicyForO365.psm1#L289

#### Verbose logs showing the problem
```PowerShell
2022-08-23T08:22:52.8920638Z ##[error]SafeLinks BlockURLs is being replaced by the Tenant Allow/Block List. Going forward all tenant blocks must be managed 
there. Deletes to BlockURLs are supported.
    + CategoryInfo          : InvalidArgument: (:) [], CimException
    + FullyQualifiedErrorId : [Server=VI1PR07MB6366,RequestId=bccaa3c2-1818-4665-9f6c-6ba25b09ffad,TimeStamp=8/23/2022 
    8:22:52 AM] [FailureCategory=Cmdlet-ValidationException] BE991DE7,Microsoft.Exchange.Management.SystemConfigurati  
  onTasks.SetAtpPolicyForO365
    + PSComputerName        : localhost
```

#### Suggested solution to the issue
Use the Microsoft Defender cmdlets to create these Blocked URL lists:
https://docs.microsoft.com/en-us/microsoft-365/security/office-365-security/set-up-safe-links-policies?view=o365-worldwide#use-powershell-to-create-safe-links-policies

#### The DSC configuration that is used to reproduce the issue (as detailed as possible)
```powershell
# insert configuration here
```

#### Version of the DSC module that was used ('dev' if using current dev branch)
dev

NikCharlebois:
Since we already have an EXOSafeLinksPolicy resource that uses this cmdlet, would the appropriate fix not be to remove these parameters from the EXOAtpPolicyForO365 resource altogether instead?

ykuijs:
[...]

PR #2393, opened 2022-10-12
EXOAtpPolicyForO365: Add BlockUrls parameter

malauter:
Add BlockUrls parameter

None


malauter:
> One of the breaking changes from last week's release was to remove blockurls from that resource. I believe we should close this PR and not add this parameter back. #2204

You are right. The documentation of the cmdlet is not up-to-date. I will update the cmdlet documentation...

PR #2438, opened 2022-10-24
PR: AADEntitlementManagementSetting - Initial release

William-Francillette:

Initial release

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADEntitlementManagementSetting/MSFT_AADEntitlementManagementSetting.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADEntitlementManagementSetting/MSFT_AADEntitlementManagementSetting.psm1
@@ -1,0 +1,494 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter()]
+        [System.String]
+        $Id,
+
+        [Parameter()]
+        [System.String]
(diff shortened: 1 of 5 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADEntitlementManagementSetting/MSFT_AADEntitlementManagementSetting.psm1
```diff
@@ -0,0 +1,494 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter()]
+        [System.String]
+        $Id,
```
NikCharlebois, line 9: I believe Id has to be 'singleton', therefore I would recommend not including it as part of the resource, but instead add the IsSingleInstance property with default value set to Yes. Take a look at the O365OrgCustomizationSetting resource for an example.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_AADEntitlementManagementSetting/MSFT_AADEntitlementManagementSetting.psm1
William-Francillette: All done, I also switched to the beta API as the v1.0 was very buggy and getting inconsistent results
I will stick to the beta for all the EntitlementManagement resources

NikCharlebois (approved):

PR #2442, opened 2022-10-25
PR: AADEntitlementManagementAccessPackageCatalog - Initial release

William-Francillette:


andikrueger:
Could you add an entry in the changelog? It looks like other resources are causing the unit test issues. Could you check if this change has any impact on this?

William-Francillette:
> Could you add an entry in the changelog? It looks like other resources are causing the unit test issues. Could you check if this change has any impact on this?

Hi Andi,
It's failing because I need to add the stub for the command - doing it now
I'll modify the changelog as well

Thanks

PR #2446, opened 2022-10-26
fixes for issue #2423

### LINKED ISSUE #2423: Error while running export using application id and secret
ISSUE TITLE:Error while running export using application id and secret
    Please prefix the issue title with the resource name, e.g.
    'ResourceName: Exporting AAD Group

    ISSUE DESCRIPTION (this template):

While running an export for any component- I get this error

Method invocation failed because [System.String] does not contain a method named 'GetNetworkCredential'.

menswearUK:
changed ApplicationSecret variable from string to PSCredential in functions  Start-M365DSCConfigurationExtract,  Export-M365DSCConfiguration, and Get-M365DSCTenantDomain. Removed line converting the string to  a pscredential as it is no longer required, added a line to convert back to string for final output

Since The updates to require ApplicationSecret to be in the form of a PSCredential I have had errors when exporting previous configurations.
I think the cause is a few functions which are still expecting ApplicationSecret to be passed as a string.  I've amended these functions to use the PSCredential.  There are also a couple of changes where the new type needs to be accounted for. I've tested it exporting the Intune workload and the IntuneAppProtectionPolicyAndroid component with no errors,
I believe All the configurations were updated at the same time to require the ApplicationSecret to be a PSCredential but I haven't tested any other configurations

Fixes #2423

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
@@ -62,7 +62,7 @@ function Start-M365DSCConfigurationExtract
         $ApplicationId,
 
         [Parameter()]
-        [System.String]
+        [System.Management.Automation.PSCredential]
         $ApplicationSecret,
 
         [Parameter()]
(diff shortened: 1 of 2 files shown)


NikCharlebois:
Changing this will have a major breaking impact on customers. We had taken the decision to minimize the impact of the ApplicationSecret DSC param to a minimum by keeping the export process as string. I am reviewing issue #2447 and will have to work on a fix that doesn't involve changing the Export-M365DSCConfiguration cmdlet to take AppSecret as PSCredential. Will update this PR once the new one is available.

PR #2449, opened 2022-10-26
Fixes #2447

### LINKED ISSUE #2447: Export does not work with ApplicationSecret
With the switch of the ApplicationSecret from String to PSCredential, the code in the Export function has not been updated. It still accepts the ApplicationSecret as a string and does not convert it later on. But then it calls New-M365DSCTenantDomain here:
https://github.com/microsoft/Microsoft365DSC/blob/7d3f6413f7b764b4b4bd6061a722dc5f33cd411c/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1#L237-L241

That in turn is calling the New-M365DSCConnection function and passing in all parameters, including the ApplicationSecret as string:
https://github.com/microsoft/Microsoft365DSC/blob/7d3f6413f7b764b4b4bd6061a722dc5f33cd411c/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1#L1398-L1399

The New-M365DSCConnection function is then trying to get the secrets from the PSCredential object (which it is not):
https://github.com/microsoft/Microsoft365DSC/blob/7d3f6413f7b764b4b4bd6061a722dc5f33cd411c/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1#L1522

This results in a "Method invocation failed because [System.String] does not contain a method named 'GetNetworkCredential'." error

Update the export functions to also require the ApplicationSecret as a PSCredential

```powershell
# insert configuration here
```

dev

NikCharlebois:
* Fixed and issue with Export and ApplicationSecret where it wasn't properly converting to a PSCredential.

* Fixes #2447


manualbashing:
@NikCharlebois This fix unfortunately breaks the certificate-based export.

You cannot expect `$ApplicationSecret` to be set, if `$CertificateThumbprint` or `$CertificatePath` is used. `ConvertTo-SecureString` will throw an exception in this case.

```powershell
# Get Tenant Info
        $organization = ''
        if ($AuthMethods -contains 'CertificateThumbprint' -or `
                $AuthMethods -contains 'CertificatePath' -or `
                $AuthMethods -contains 'ApplicationWithSecret')
        {
            [SecureString]$secStringPassword = ConvertTo-SecureString $ApplicationSecret -AsPlainText -Force
            [PSCredential]$AppSecretAsPSCredential = New-Object System.Management.Automation.PSCredential ('ApplicationSecret', $secStringPassword)
```
It would probably be useful to use a separate if-condition to deal with the `$ApplicationSecret `.

mlhickey:
Accounted for this in PR #2460

Issue #2453, opened 2022-10-27
IntuneAppProtectionPolicyAndroid/iOS - Errors when pushing test policies


MichealKennard:
#### Details of the scenario you tried and the problem that is occurring
Performing testing of Microsoft365DSC, able to export components referenced in subject, compile and create .mof file without much issue. However, when pushing the configuration most of the push seems to run fine until the end which generates several errors. The end result is neither policy is being configured. Given my relatively new experience level, may be simple issue but wanted to post for support.

#### Verbose logs showing the problem
![image](https://user-images.githubusercontent.com/116776712/198177377-0ca76d24-ee05-44fa-b717-985ba301bd7e.png)
VERBOSE: [...-6143KL3]:                            [[IntuneAppProtectionPolicyiOS]da4bdc60-9c8a-4e86-8a35-fd48eaa9fb6d] POST https://graph.microsoft.com/beta/deviceAppManagement/managedAppPolicies
HTTP/1.1 400 Bad Request
Transfer-Encoding: chunked
Vary: Accept-Encoding
Strict-Transport-Security: max-age=31536000
request-id: d6af9341-8c33-4c8e-b37b-0bc45766feee
client-request-id: d6af9341-8c33-4c8e-b37b-0bc45766feee
x-ms-ags-diagnostic: {"ServerInfo":{"DataCenter":"South Central US","Slice":"E","Ring":"5","ScaleUnit":"000","RoleInstance":"SN1PEPF000051D0"}}
Date: Thu, 27 Oct 2022 02:15:38 GMT
Content-Encoding: gzip
Content-Type: application/json

{"error":{"code":"BadRequest","message":"Unable to read JSON request payload. Please ensure Content-Type header is set and payload is of valid JSON format.","innerError":{"date":"2022-10-27T02:15:38","request-id":"d6af9341-8c33-4c8e-b37b-0bc45766feee","clien
t-request-id":"d6af9341-8c33-4c8e-b37b-0bc45766feee"}}}

VERBOSE: [...-6143KL3]:                            [[IntuneAppProtectionPolicyiOS]da4bdc60-9c8a-4e86-8a35-fd48eaa9fb6d] Found Policy with Id {}
Cannot bind argument to parameter 'PolicyId' because it is an empty string.
    + CategoryInfo          : InvalidData: (:) [], CimException
[...]

PR #2474, opened 2022-10-31
IntuneAppProtectionPolicyAndroid - Add New Parameters

### LINKED ISSUE #2467: IntuneAppProtectionPolicyAndroid - AppGroupType & Managedidentity not pulling correctly
Performed export of all Intune components against target tenant. Have performed several and each time the fields referenced have to be modified manually before compiling as it doesnt pull correctly.

No logs, just screen snip showing issue
image

Not sure, but potentially adjust logic so it correctly adds "" to variable input or $

Windows 10 via Powershell ISE

# insert configuration here
```

<!--
    Please provide as much as possible about the target node, for example
    edition, version, build and language.
    On OS with WMF 5.1 the following command can help get this information.

    Get-ComputerInfo -Property @(
        'OsName',
        'OsOperatingSystemSKU',
        'OSArchitecture',
        'WindowsVersion',
        'WindowsBuildLabEx',
        'OsLanguage',
        'OsMuiLanguages')
-->

Current 1.22.1026.2 but have seen this on previous versions as well testing last couple days.

### PR DESCRIPTION
**Pull request by menswearUK**
#### Pull R#### Pull Request (PR) description
This pull requests adds functionality to the IntuneAppProtectionPolicyAndroid configuration.  This should now support all the values which can currently be set by the graph cmdlet.

The below parameters have been added:
    ScreenCaptureBlocked
    EncryptAppData
    DisableAppEncryptionIfDeviceEncryptionIsEnabled
    
The following parameters are also added and work in conjunction with the managed browser value.
    ManageBrowserToOpenLinksRequired
    CustomBrowserDisplayName
    CustomBrowserPackageID
I had to add some analysis to prevent then setting invalid configurations - I've set it to only allow configurations which can be set via the web portal and to interpret and alter the values entered as the web portal would do

I've also added the ability to use the configuration ID.  If the specified ID is not found it will default to searching for the DisplayName value rather than creating new, potentially duplicated, profiles

I also added a couple of amendments to the get and test functions to address a problem where the AppGroupType value 
    wasn't being reported correctly and so the test was reporting drift

fixes #2467

### CODE REVIEW

**Comment by menswearUK** [comment]
I need to address issue #2467 so I'll get that added in today on this branch

File: Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppProtectionPolicyAndroid/MSFT_IntuneAppProtectionPolicyAndroid.psm1 [hunk 1]
```diff
@@ -786,11 +887,18 @@ function Test-TargetResource
 
         [Parameter()]
         [Switch]
-        $ManagedIdentity
+        $ManagedIdentity,
+
+        [Parameter()]
+        [System.String]
+        $Id
```
NikCharlebois, line 894: We need to make sure we ignore Id in the Test-TargetResource. We shouldn't compare Ids

NikCharlebois (changes requested):

menswearUK:
@NikCharlebois I've removed the ID comparison from the test-TargetResource function.  I also skipped the comparison when no existing policy is returned as this had been causing some errors on the checks and may be the reason the android policy was not created in #2453.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppProtectionPolicyAndroid/MSFT_IntuneAppProtectionPolicyAndroid.psm1
```diff
@@ -884,6 +998,9 @@ function Test-TargetResource
     # wipe out the current apps value if AppGroupType is anything but selectedpublicapps to match the appshash values
     if ($CurrentValues.AppGroupType -ne 'selectedPublicApps') { $CurrentValues.Apps = @() }
 
+    # remove thre ID from the values to check as it may not match
+    $targetvalues.remove('ID')
```
NikCharlebois, line 1002: Please pipe with | Out-Null when removing values. Otherwise these will attempt to throw a verbose message and sometimes cause conflict with the export feature.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppProtectionPolicyAndroid/MSFT_IntuneAppProtectionPolicyAndroid.psm1
```diff
@@ -1125,52 +1241,120 @@ function set-AppsHash
...
+
+    $ManagedBrowserHash = @{
+        'ManagedBrowser' = $ManagedBrowser;
+        'ManagedBrowserToOpenLinksRequired' = $ManagedBrowserToOpenLinksRequired;
+        'CustomBrowserDisplayName' = $CustomBrowserDisplayName;
+        'CustomBrowserPackageId' = $CustomBrowserPackageId
+    }
+
+    return $ManagedBrowserHash
+
```
NikCharlebois, line 1299: Remove a few empty lines to keep it clean. Picky, I know ;)

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppProtectionPolicyAndroid/MSFT_IntuneAppProtectionPolicyAndroid.psm1
menswearUK: @NikCharlebois No worries, I've made those amendments now - let me know if there's anything else needed.  thanks

NikCharlebois (approved):

Issue #2469, opened 2022-10-31
New DSC-resource: AADAdministrativeUnit


salbeck-sit:
### Description
New DSC resource for managing AAD Administrative Units

### Proposed properties
Id
Description
DisplayName
Visibility
MembershipType
MembershipRuleProcessingState
Members
ScopedRoleMembers
(+ standard properties)

### Special considerations or limitations
I've decided to *omit* support for Extensions - at least for now. One thing is that the format is tricky (arbitrary property-names with arbitrary values - perhaps feasible using MSFT_KeyValuePair?), another is that required permissions would include users, groups and devices since an update to Extensions of an AU would require properties to propagate to every user, group and/or device in scope for that AU.
Then there is MembershipType and MembershipRuleProcessingState. These are features of a public preview but I've chosen to include them as we use that preview in several places and it's been in preview for so long that I believe it's a matter of weeks before it's GA.

salbeck-sit:
I have built the resource and am in the process of ensuring that unit-tests are comprehensive enough to validate functionality.

NikCharlebois:
@salbeck-sit I actually had cycles yesterday morning and did a first stab at it. It currently only supports configuringdynamic memberships, but reports on assigned permissions. This resource had proven to be more complex than anticipated to build. Let's compare and merge in whatever additional properties you are handling with yours.

salbeck-sit:
@NikCharlebois I also included code for ScopedRoleMember and some preliminary thoughts on Extensions that led me to exclude the attribute. I can't do a PR on top of your PR but have a look at https://github.com/salbeck-sit/Microsoft365DSC/tree/AADAdministrativeUNit/Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit
PS When I built it, I think I found a flaw in DRG when it builds code for Test-TargetResource for handling comparisons of attributes where one is empty. I'll see if I can find it again.

salbeck-sit:
PR #2483 submitted. See what you think

PR #2483, opened 2022-11-03
Alternative implementation of AADAdministrativeUnit

salbeck-sit:
Alternative implementation of #2469

Implementation of AADAdministrativeUnit including support for Member and ScopedRoleMember


salbeck-sit:
hmm. Graph-modules were updated. I'll do another test-run locally and see exactly what fails

salbeck-sit:
It seems unit-testing has an issue with cmdlets that only exist in Graph's beta profile - such as New-MgAdministrativeUnitMemberByRef. If I do 'Select-MgProfile beta' before calling Invoke-TestHarness then tests pass. Which is odd, because I also do that explicitly in the Pester-script. But it seems to me that perhaps because Pester invokes the script and not Powershell, some commands are skipped

salbeck-sit:
When I run unit-tests locally, everything passes. The tests that are run by the PR Action complains that microsoft.graph.* modules v1.15.0 are in use ? Something in the environment is obviously different. Could it be that the workflow-file specifies older Graph-versions ?

salbeck-sit:
Nope. The workflow doesn't use module-versions but simply installs latest versions. Which may or may not be the same as versions in the Dependencies manifest. Another difference is that the workflow installs Pester v5.0.4 whereas I use v5.3.1. That shouldn't be an issue.

salbeck-sit:
It is remotely possible to replace all calls to the beta-cmdlets with Invoke-MgGraphRequest and the corresponding Uris and a bunch of BodyParameter's, but I'd rather be able to create DSC-resources that utilize the beta-profile.
In any case, depending on whether and/or how you'll prefer to move forward, it would be nice to see a line or two about the issue in the M365DSC blog

salbeck-sit:
I believe this is a Pester issue. See https://github.com/pester/Pester/issues/2256

salbeck-sit:
Now using v1.0 cmdlet-names for everything but the AU itself (ie members and scopedrolemembers). Local test passes

NikCharlebois (approved):

PR #2566, opened 2022-11-20
Update-M365DSCAzureAdApplication with AppOwnerTenantID Filter

### LINKED ISSUE #2565: Update-M365DSCAzureAdApplication could grant permissions to non-MS service application
Update-M365DSCAzureAdApplication is used to update a dedicated application with service endpoint permissions - Microsoft Graph, Office 365 SharePoint Online and/or Office 365 Exchange Online - retrieved via Get-AzADServicePrincipal, filtering the result set based on displayName.  Since displayName is a non-unique value in Azure, this opens the possibility of retrieving multiple entries for a single name and does not guarantee the servicePrincipal is the actual Microsoft service,e.g.
```
$gsp = Get-AzADServicePrincipal -Filter "displayName eq 'Microsoft Graph'"
$gsp.Count
3
$gsp

DisplayName     Id                                   AppId
-----------     --                                   -----
Microsoft Graph b19d498e-6687-4156-869a-2e8a95a9d659 00000003-0000-0000-c000-000000000000
Microsoft Graph ad92c665-9f52-4566-9b96-470bbcd40392 bdfa89eb-5d89-45f9-9a46-fcb772f808b9
Microsoft Graph 5bbe29f8-4ed6-48b9-a300-babff8fc21b6 fc381d29-52ab-4e40-a70d-aa2c356927a7
```
1. Filter the query rather than the result set - an unbound query could take significant time in a large organization
2. Validate the returned value based on AppOwnerTenantId 1st party tenant
e.g.
 ```$gsp = Get-AzADServicePrincipal -Filter "displayName eq 'Microsoft Graph'" | Where-Object -FilterScript {$_.AppOwnerTenantId-eq 'f8cdef31-a31e-4b4a-93e4-5f571e91255a'}'"```
```powershell
# insert configuration here
```

andikrueger:
Added addtional filter for AppOwernTenantId in Update-M365DSCAzureAdApplication

- Fixes #2565

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
@@ -60,7 +60,7 @@ function Get-M365DSCCompiledPermissionList
         }
         'Exchange'
         {
-            Write-Verbose -Message "Retrieving Exchange Permissions"
+            Write-Verbose -Message 'Retrieving Exchange Permissions'
             $results = @{
                 RequiredRoles      = @()
                 RequiredRoleGroups = @()
@@ -169,7 +169,7 @@ function Get-M365DSCCompiledPermissionList
                     }
                     else
                     {
-                        Write-Warning "Error in reading Exchange permissions. Missing exchange node in settings.json."
+                        Write-Warning 'Error in reading Exchange permissions. Missing exchange node in settings.json.'
                         continue
                     }
                 }
@@ -269,24 +269,24 @@ function Update-M365DSCAllowedGraphScopes
     {
         Write-Verbose -Message 'All parameter specified'
         $dscResourcesRoot = Join-Path -Path $PSScriptRoot -ChildPath '..\DSCResources'
-        $resourceNames = (Get-ChildItem -Path $dscResourcesRoot -Directory).Name -replace "MSFT_", ''
+        $resourceNames = (Get-ChildItem -Path $dscResourcesRoot -Directory).Name -replace 'MSFT_', ''
         $permissions = Get-M365DSCCompiledPermissionList -ResourceNameList $resourceNames -Source 'Graph' -PermissionsType 'Delegated'
     }
     else
     {
         if ($PSBoundParameters.ContainsKey('ResourceNameList') -eq $false)
         {
-            throw "You have to specify either the All or ResourceNameList parameter!"
+            throw 'You have to specify either the All or ResourceNameList parameter!'
         }
 
-        Write-Verbose -Message "Specified resources: $($ResourceNameList -join ", ")"
+        Write-Verbose -Message "Specified resources: $($ResourceNameList -join ', ')"
         $permissions = Get-M365DSCCompiledPermissionList -ResourceNameList $ResourceNameList -Source 'Graph' -PermissionsType 'Delegated'
     }
 
     if ($Type -eq 'Read')
     {
         Write-Verbose -Message 'Specified type: Read'
-        Write-Verbose -Message "Found permissions: $($permissions.ReadPermissions -join ", ")"
+        Write-Verbose -Message "Found permissions: $($permissions.ReadPermissions -join ', ')"
         $params = @{
             Scopes = $permissions.ReadPermissions
         }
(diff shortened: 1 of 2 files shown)


Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
```diff
@@ -1273,12 +1273,12 @@ function Update-M365DSCAzureAdApplication
         $context = Get-AzContext
     }
 
-    $graphSvcprincipal = Get-AzADServicePrincipal | Where-Object -FilterScript { $_.DisplayName -eq "Microsoft Graph" }
-    $spSvcprincipal = Get-AzADServicePrincipal | Where-Object -FilterScript { $_.DisplayName -eq "Office 365 SharePoint Online" }
-    $exSvcprincipal = Get-AzADServicePrincipal | Where-Object -FilterScript { $_.DisplayName -eq "Office 365 Exchange Online" }
+    $graphSvcprincipal = Get-AzADServicePrincipal | Where-Object -FilterScript { $_.DisplayName -eq 'Microsoft Graph' -and $_.AppOwnerTenantId -eq 'f8cdef31-a31e-4b4a-93e4-5f571e91255a' }
```
NikCharlebois, line 1276: Why are we doing this, and also, it looks like the property is appOwnerOrganizationId and not appOwnerTenantId
image
Will this owner id be the same in foreign clouds?

andikrueger: Good point. 

The display name is not unique across all tenants. At the moment, any other application with the same name could be the first and we would try to add permissions to this app. 

Do you have any insight on other clouds, if the IDs are the same?

NikCharlebois: Not at the moment. I'll need to ask around and investigate

mlhickey: Get-AzureADServicePrincipal -> appOwnerTenantId
Get-AzADServicePrincipal/Get-MgServicePrincipal -> AppOwnerOrganizationId

f8cdef31-a31e-4b4a-93e4-5f571e91255a is the 1st party tenant for commercial and usgov, can't speak to other dedicated clouds.

andikrueger: …what a mess. I’ll look into moving to MG methods tomorrow. Maybe there is another criteria available to make sure the apps are published by Microsoft.

mlhickey: Agreed.  For me it's always been primary of tenantId, secondary of 
```
publisherName eq 'Microsoft Services'
```

andikrueger: Publisher Name sound promising. Maybe we could verify it in two steps. Plain get with the id and id there are more than one, we would start to filter and add some verbose/warning messages.

mlhickey: Sounds good.  For the sake of efficiency, I'd provide the displayname as part of the query rather than filter on the complete results set - the difference in query time can be significant when there are a large number of SPs in the tenant.

andikrueger: Just had another thought: How does PnP PowerShell do it?

This is the answer: https://github.com/pnp/powershell/blob/ed883559841bb89e8bb08515d0fdfed4f69faf17/src/Commands/Model/PermissionScopes.cs#L11-L13

```c#
        public static string ResourceAppId_Graph = "00000003-0000-0000-c000-000000000000";
        public static string ResourceAppId_SPO = "00000003-0000-0ff1-ce00-000000000000";
        public static string ResourceAppID_O365Management = "c5393580-f805-4401-95e8-94b7a6ef2fc2";
```

Maybe we can go with fixed IDs within our code as well and just warn, if we do not find a thing. As soon as we do not find any app, we could switch back to the old "query".

mlhickey: I was thinking about that as well, create a table of well-known GUIDs and call it a day.

NikCharlebois: That works

andikrueger: Will update accordingly.

andikrueger:
Will create a new PR to add the points that were discussed above.

PR #2568, opened 2022-11-21
Fix for issues with Update-M365DSCAzureAdApplication

### LINKED ISSUE #2567: Missing dependency Az.Resources for Update-M365DSCAzureAdApplication
Followed the user guide to install Microsoft365DSC by running:

```
Install-Module Microsoft365DSC -Force
Update-M365DSCDependencies
```

Ran the following with the intention of getting all update permissions and then creating an Azure AD App Registration for certificate authentication using Update-M365DSCAzureAdApplication:

```
$AllPermissions = Get-M365DSCCompiledPermissionList -ResourceNameList (Get-M365DSCAllResources)
$UpdatePermissions = $AllPermissions.UpdatePermissions

$PermList = @()
foreach ($UpdatePermission in $UpdatePermissions) {
    $PermObject = @{Api="Graph";PermissionName="$UpdatePermission"}
    $PermList += $PermObject
}

Connect-MgGraph -Scopes $UpdatePermissions

Update-M365DSCAzureAdApplication -ApplicationName 'Microsoft365DSC' -Permissions $PermList -AdminConsent -Type Certificate -CreateSelfSignedCertificate -CertificatePath c:\temp\M365DSC.cer
```
Then I am met with:

image

N/A
N/A
N/A

OsName               : Microsoft Windows 10 Enterprise
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsBuildLabEx    : 19041.1.amd64fre.vb_release.191206-1406
OsLanguage           : en-US
OsMuiLanguages       : {en-US}

Release 1.22.1109.1

andikrueger:
  * Fixed an issue with missing dependency on Az.Resources and an issue with certificate creation.
    FIXED [2567](https://github.com/microsoft/Microsoft365DSC/issues/2567)

- Fixes #2567

### DIFF
--- a/Modules/Microsoft365DSC/Dependencies/Manifest.psd1
+++ b/Modules/Microsoft365DSC/Dependencies/Manifest.psd1
@@ -73,8 +73,8 @@
             RequiredVersion = '4.9.1'
         },
         @{
-            ModuleName      = "MSCloudLoginAssistant"
-            RequiredVersion = "1.0.98"
+            ModuleName      = 'MSCloudLoginAssistant'
+            RequiredVersion = '1.0.98'
         },
         @{
             ModuleName      = 'PnP.PowerShell'
@@ -83,6 +83,14 @@
         @{
             ModuleName      = 'ReverseDSC'
             RequiredVersion = '2.0.0.13'
+        },
+        @{
+            ModuleName      = 'Az.Resources'
+            RequiredVersion = '6.4.1'
+        },
+        @{
+            ModuleName      = 'Az.Resources'
+            RequiredVersion = '6.4.1'
         }
     )
 }
--- a/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
@@ -60,7 +60,7 @@ function Get-M365DSCCompiledPermissionList
         }
         'Exchange'
         {
-            Write-Verbose -Message "Retrieving Exchange Permissions"
+            Write-Verbose -Message 'Retrieving Exchange Permissions'
             $results = @{
                 RequiredRoles      = @()
                 RequiredRoleGroups = @()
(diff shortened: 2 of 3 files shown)


Modules/Microsoft365DSC/Dependencies/Manifest.psd1
```diff
@@ -83,6 +83,14 @@
         @{
             ModuleName      = 'ReverseDSC'
             RequiredVersion = '2.0.0.13'
+        },
+        @{
```
NikCharlebois, line 87: Why do we have it listed twice, and can we please add it to the top of the dependency manifest? Also, I would prefer we don't ever call in the Az module. Everything we need should be available as part of the Microsoft.Graph cmdlets. E.g. Get-MgServicePrincipal instead of Get-AzADServicePrincipal.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
```diff
@@ -1201,51 +1201,52 @@ function Update-M365DSCAzureAdApplication
...
-                if ($PSBoundParameters.ContainsKey("CreateSelfSignedCertificate"))
+                if ($PSBoundParameters.ContainsKey('CreateSelfSignedCertificate'))
                 {
                     # CreateSelfSignedCertificate is specified and path specified in CertificatePath already exists.
                     if ((Test-Path -Path $CertificatePath) -eq $true)
                     {
                         Write-LogEntry -Message "Specified CertificatePath '$CertificatePath', where the Self Signed Certificate should be exported, already exists." -Type Error
                         return
                     }
+                    NEw-M365DSC
```
ykuijs, line 1249: What is this for? Looks like an incomplete edit.

Modules/Microsoft365DSC/Dependencies/Manifest.psd1
ykuijs: As Nik also mentions, this PR has the Az.Resources module mentioned twice. One can be removed.

I agree with Nik that it would be great to update to the Graph module, but that will require some more testing. Probably better to implement in a separate change.

ykuijs (changes requested):

NikCharlebois:
@andikrueger what should we do with this PR?

andikrueger:
I need to start over with this one. The rebase is a mess. I Would go for a simple fix first with fixed IDs and we can plan to move to a graph based variant afterwards.

andikrueger:
This issue got resolved by PR #2844

PR #2637, opened 2022-12-08
Fix #2599

### LINKED ISSUE #2599: MSFT_AADUser: hard coded password for new user
The password for new users is hard coded in line 495. The parameter "password" you can pass is ignored and never used. This is critical, because every user created with M365DSC will have the same public available password.

no error/no logs

Make sure, that the password provided as a parameter is used in the resource.

Password             = New-Object System.Management.Automation.PSCredential('Password', (ConvertTo-SecureString 'your password' -AsPlainText -Force));;

N/A
1.22.1109.1 (Resource code did not change in the newer releases)

NikCharlebois:
* AADUser
  * Fixes an issue where provided password wa never honored.

* Fixes #2599

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 * AADServicePrincipal
   * Fixes an issue where the service principals weren't created or updated when using ApplicationSecret to authenticate.
     FIXES [#2615](https://github.com/microsoft/Microsoft365DSC/issues/2615)
+* AADUser
+  * Fixes an issue where provided password wa never honored.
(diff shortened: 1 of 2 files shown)


andikrueger:
Please have a look at my latest comment within #2599

andikrueger:
LGTM 👍

andikrueger:
This is a good remark on this resource and I think there is just some adjustment to do. 

for Update Scenarios:
We can update the password only in cases that a value for password is passed and set within the configuration. 

for create scenarios:
for newly created users we would go with the current change.

PR #2641, opened 2022-12-08
Fixes #2598

### LINKED ISSUE #2598: MSFT_AADUser: Bug getting passwordpolicies
Passwordpolicy not retrieved because property is missing in $propertiesToRetrieve.

No error/no logs.

old code (line 183):

$propertiesToRetrieve = @('Id', 'UserPrincipalName', 'DisplayName', 'GivenName', 'Surname', 'UsageLocation', 'City', 'Country', 'Department', 'FacsimileTelephoneNumber', 'Mobile', 'OfficeLocation', 'TelephoneNumber', 'PostalCode', 'PreferredLanguage', 'State', 'StreetAddress', 'JobTitle', 'UserType')

new code:

$propertiesToRetrieve = @('Id', 'UserPrincipalName', 'DisplayName', 'GivenName', 'Surname', 'UsageLocation', 'City', 'Country', 'Department', 'FacsimileTelephoneNumber', 'Mobile', 'OfficeLocation', 'TelephoneNumber', 'PostalCode', 'PreferredLanguage', 'State', 'StreetAddress', 'JobTitle', 'UserType', 'PasswordPolicies')

N/A

N/A

1.22.1109.1 (Resource code did not change in the newer releases)

NikCharlebois:
  * Added support for the PasswordPolicies property.

* FIXES #2598

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@
 * AADUser
   * Fixes an issue where provided password wa never honored.
     FIXES [#2599](https://github.com/microsoft/Microsoft365DSC/issues/2599)
+  * Added support for the PasswordPolicies property.
+    FIXES [#2598](https://github.com/microsoft/Microsoft365DSC/issues/2598)
 * PlannerBucket & PlannerPlan
-  * Changed invalid permissions in the setting.json files
+  * Changed invalid permissions in the setting.json files.
     FIXES [#2629](https://github.com/microsoft/Microsoft365DSC/issues/2629)
 * TeamsChannel
   * Fixes an issue where channels weren't created if a non-existing GroupId was specified.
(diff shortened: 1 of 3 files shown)


andreasmarx:
@NikCharlebois The property 'PasswordPolicies' does not need to be exposed as a parameter, because we already have the parameter 'PasswordNeverExpires' which actually sets the property 'PasswordPolicies'. Or we remove the parameter 'PasswordNeverExpires' and expose 'PasswordPolicies'. Exposing both makes no sense from my point of view. With the current code change we would have the new parameter which is never honored. 'Set-TargetResource' would have to be modified for that.

'PasswordPolicies' would accept 'DisablePasswordExpiration', 'DisableStrongPassword' or both 'DisablePasswordExpiration, DisableStrongPassword' and should return NONE or NULL when not defined. (would have to be tested ...)
https://learn.microsoft.com/en-us/powershell/module/microsoft.graph.users/new-mguser?view=graph-powershell-1.0

For just fixing the original issue I would recommend to remove the new parameter.

PR #2652, opened 2022-12-09
Fixes #2444

### LINKED ISSUE #2444: Ignore certain properties in Assert-M365DSCBlueprint
We're using Assert-M365DSCBlueprint to assert and monitor configuration drift for a number of tenants. Therefore we created a blueprint which is not specific for one tenant, but contains settings which we would assert on many tenants. For example, we assert the following named location must be present:

```JSON
AADNamedLocationPolicy 'Trusted Countries for Administrative Access' {
    CountriesAndRegions               = @("CH");
    Credential                        = $Credscredential;
    DisplayName                       = "Trusted Countries for Administrative Access";
    Ensure                            = "Present";
    IncludeUnknownCountriesAndRegions = $False;
    OdataType                         = "#microsoft.graph.countryNamedLocation";
    Managedidentity                   = $False;
    Id                                = "";
}
```

The problem is the property 'Id' will differ in every tenant. I've tried '*' and '%' as wildcard to accept any value, and even exclude this property, but still every time this is reported as configuration drift. 

I've noted [here](https://github.com/microsoft/Microsoft365DSC/blob/Dev/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1#L519) that some properties are excluded from reporting, is there a way to exclude others as well (apart from a fork and adjust this line)?

This happens with other resources as well, it is not specific to AADNamedLocationPolicy. I know one obvious solution would be having a blueprint for each tenant with correct Id, but with many tenants this is a cumbersome task.

Thanks in advance
Christian

NikCharlebois:
  * Added support for the ExcludedProperties parameter in the New-M365DSCDeltaReport function.

* FIXES #2444

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
@@ -427,7 +427,10 @@ Local path of the destination configuraton.
 Array that contains the list of configuration components for the source.
 
 .Parameter DestinationObject
-Array that contains the list of configuration components for the destination. |
+Array that contains the list of configuration components for the destination.
+
+.Parameter ExcludedProperties
+Array that contains the list of configuration components for the destination.
 
 .Example
 Compare-M365DSCConfigurations -Source 'C:\DSC\source.ps1' -Destination 'C:\DSC\destination.ps1'
@@ -459,7 +462,11 @@ function Compare-M365DSCConfigurations
 
         [Parameter()]
         [Array]
-        $DestinationObject
+        $DestinationObject,
+
+        [Parameter()]
+        [Array]
+        $ExcludedProperties
     )
 
     if ($CaptureTelemetry)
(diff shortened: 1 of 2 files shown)


Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
```diff
@@ -427,7 +427,10 @@ Local path of the destination configuraton.
 Array that contains the list of configuration components for the source.
 
 .Parameter DestinationObject
-Array that contains the list of configuration components for the destination. |
+Array that contains the list of configuration components for the destination.
+
+.Parameter ExcludedProperties
+Array that contains the list of configuration components for the destination.
```
andikrueger, line 433: Please check the parameter’s description. Looks like it’s the one of destinationobject.

Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
```diff
@@ -806,6 +815,9 @@ Specifies that file that contains a custom header for the report.
 .Parameter Delta
 An array with difference, already compiled from another source.
 
+.Parameter ExcludedProperties
```
andikrueger, line 818: Same as above

PR #2666, opened 2022-12-14
Update Node.js to version 16

bladeoner:
This PR updates Node.js from version 12/14 to version 16

Under the Workflows you can see that Node.js 12 is deprecated and that you should update to version 16, this PR does that:
image


bladeoner:

PR #2695, opened 2022-12-20
TeamsShiftsPolicy: Initial Release

NikCharlebois:
* New TeamsShiftsPolicy Resource

* N/A


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsShiftsPolicy/MSFT_TeamsShiftsPolicy.psm1
```diff
@@ -0,0 +1,476 @@
...
+    )
+
+    try
+    {
+        New-M365DSCConnection -Workload 'MicrosoftTeams' `
+            -InboundParameters $PSBoundParameters | Out-Null
+    }
+    catch
+    {
+        Write-Verbose -Message $_
```
ykuijs, line 68: Why do you only write a verbose here? If the catch is triggered, doesn't that mean creating the connection failed? And in that case, shouldn't the rest of the code be aborted?

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsShiftsPolicy/MSFT_TeamsShiftsPolicy.psm1
```diff
@@ -0,0 +1,476 @@
...
```
ykuijs, line 189: Same comment as before

ykuijs (changes requested):
Question needs clarification

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsShiftsPolicy/MSFT_TeamsShiftsPolicy.psm1
NikCharlebois: Yes, it should. This Teams resource was entirely generated via the Dynamic Resource Generator. I will regenerate with your comments included.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsShiftsPolicy/MSFT_TeamsShiftsPolicy.psm1
NikCharlebois: To be aligned with the other resources, we will remove the try/catch. The error will be trapped inside the New-M365DSCCOnnection function instead

ykuijs (approved):
LGTM

PR #2697, opened 2022-12-21
Major Changes to the DRG to Handle Non-Graph based Workloads

NikCharlebois:
* Major Changes to the DRG to Handle Non-Graph based Workloads

* N/A


NikCharlebois:
@William-Francillette FYI, I had to split the logic between Graph based and non-graph based resources. I've also introduced a Workload manifest file so that we can automatically regenerate all resources from a list. Please consider these changes when working on Intune resources. The code isn't optimally refactored, but the goal is for it to eventually disappear once all workloads have their Microsoft Graph footprint in place.

PR #2733, opened 2023-01-03
Teams group policy assignment

sandrola:
Initial Release

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADAdministrativeUnit/MSFT_AADAdministrativeUnit.psm1
@@ -92,10 +92,8 @@ function Get-TargetResource
     }
     catch
     {
-        Write-Verbose -Message 'Reloading1'
+        Write-Verbose -Message ($_)
     }
-
-
 
     #Ensure the proper dependencies are installed in the current environment.
     Confirm-M365DSCDependencies
@@ -1003,8 +1001,8 @@ function Test-TargetResource
     # Removing the visibility parameter from the check since this is always being returned as null currently by the Microsoft Graph.
     $ValuesToCheck.Remove('Visibility') | Out-Null
 
-    #Write-Verbose -Message "Current Values: $(Convert-M365DscHashtableToString -Hashtable $CurrentValues)"
-    #Write-Verbose -Message "Target Values: $(Convert-M365DscHashtableToString -Hashtable $ValuesToCheck)"
+    Write-Verbose -Message "Current Values: $(Convert-M365DscHashtableToString -Hashtable $CurrentValues)"
+    Write-Verbose -Message "Target Values: $(Convert-M365DscHashtableToString -Hashtable $ValuesToCheck)"
 
     #Convert any DateTime to String
     foreach ($key in $ValuesToCheck.Keys)
--- a/Tests/Unit/Microsoft365DSC/Microsoft365DSC.AADAdministrativeUnit.Tests.ps1
+++ b/Tests/Unit/Microsoft365DSC/Microsoft365DSC.AADAdministrativeUnit.Tests.ps1
@@ -81,10 +81,19 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
                     Description = 'FakeStringValue1'
                     DisplayName = 'FakeStringValue1'
                     Id          = 'FakeStringValue1'
+                    Members     = (New-CimInstance -ClassName MSFT_MicrosoftGraphIdentity -Property @{
+                        Type = 'User'
+                        Identity = 'john.smith@contoso.com'
+                    } -ClientOnly)
                     Visibility  = 'Public'
-
                     Ensure      = 'Present'
                     Credential  = $Credential
+                }
+
+                Mock -CommandName Get-MgUser -MockWith {
+                    return @{
+                        Id = "123456"
+                    }
                 }
 
                 Mock -CommandName Get-MgAdministrativeUnit -MockWith {
@@ -109,7 +118,10 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
                     Description = 'FakeStringValue2'
                     DisplayName = 'FakeStringValue2'
                     Id          = 'FakeStringValue2'
-
+                    Members     = (New-CimInstance -ClassName MSFT_MicrosoftGraphIdentity -Property @{
+                        Type = 'User'
+                        Identity = 'john.smith@contoso.com'
+                    } -ClientOnly)
                     Ensure      = 'Absent'
                     Credential  = $Credential
                 }
@@ -174,7 +186,9 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
                             )
                         #>
                     Visibility        = 'Public'
-
+                    MembershipType    = 'Assigned'
+                    MembershipRule = 'Canada'
+                    MembershipRuleProcessingState = 'On'
                     Ensure            = 'Present'
                     Credential        = $Credential
                 }
@@ -207,6 +221,11 @@ Describe -Name $Global:DscHelper.DescribeHeader -Fixture {
                             }
                         )
                         Visibility        = 'Public'
+                        AdditionalProperties = @{
+                            membershipType = 'Assigned'
+                            membershipRule = 'Canada'
+                            membershipRuleProcessingState = 'On'
+                        }
                     }
                 }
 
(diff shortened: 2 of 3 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+        -CommandName $CommandName `
+        -Parameters $PSBoundParameters
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftTeams' -InboundParameters $PSBoundParameters
+    $CurrentValues = Get-TargetResource @PSBoundParameters
+
+    #check policyname
+    $command = "get-cs" + $PolicyType
```
ykuijs, line 182: What are you trying to do here? According to the documentation, the PolicyType parameter can be any of these values:

- CallingLineIdentity
- TeamsAppSetupPolicy
- TeamsAudioConferencingPolicy
- TeamsCallingPolicy
- TeamsCallParkPolicy
- TeamsChannelsPolicy
- TeamsComplianceRecordingPolicy
- TenantDialPlan
- TeamsMeetingBroadcastPolicy
- TeamsMeetingPolicy
- TeamsMessagingPolicy
- TeamsShiftsPolicy
- TeamsUpdateManagementPolicy
- TeamsVerticalPackagePolicy
Source: https://learn.microsoft.com/en-us/powershell/module/teams/new-csgrouppolicyassignment?view=teams-ps#-policytype

For several of these values, there is no Get-Cs..... cmdlet. Like the first two: CallingLineIdentity and TeamsAppSetupPolicy. I think this can result in errors, especially since you are not doing any validation and directly inserting the command into Invoke-Expression. This can lead to security issues, where someone specifies "whatever -ErrorAction SilentlyContinue;Remove-Item -Path C:\ -Recurse" into the PolicyType parameter.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    #get groupid
+    if($GroupId.Length -eq 0)
+    {
+        $Group = Find-CsGroup -SearchQuery $GroupDisplayname
+        if($group.Length -gt 1)
+        {
+            Write-Verbose -Message "Found $($group.Length) groups with the name $GroupDisplayname"
+            $Group = $Group | Where-Object { $_.DisplayName -eq $GroupDisplayname }
```
ykuijs, line 205: Just wondering: On line 201 you search on the GroupDisplayName. If that results in two groups with that specific display name, when you filter again for that same GroupDisplayName on this line: Doesn't this result in the same set of groups?

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+            new-csGroupPolicyAssignment -GroupId $GroupId -PolicyType $PolicyType -PolicyName $PolicyName -Rank $Priority
+        }
+        elseif($Ensure -eq 'Absent' -and $CurrentValues.Ensure -eq 'Present')
+        {
+            Write-Verbose -Message "Remove GroupPolicyAssignment for $GroupDisplayname"
+            Remove-CsGroupPolicyAssignment -GroupId $CurrentValues.GroupId -PolicyType $CurrentValues.PolicyType
+        }
+    }
+    catch{
+        Write-Verbose -Message "Error while setting GroupPolicyAssignment for $GroupDisplayname"
```
ykuijs, line 236: Please use the same logging logic as used on line 106 to 110. That will also output to the event log and other logs.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+            Write-Host "`r`n" -NoNewline
+        }
+        $dscContent = [System.Text.StringBuilder]::new()
+        $j = 1
+        foreach ($item in $instances)
+        {
+            try
+            {
+                $Group = Find-CsGroup -SearchQuery $item.GroupId
+                $totalCount = $instances.Length
```
ykuijs, line 378: This line should be placed before the foreach statement. That will make sure it is only updated once, instead of with each iteration of the foreach loop.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+                        -Results $Results `
+                        -Credential $Credential
+                    $dscContent.Append($currentDSCBlock) | Out-Null
+                    Save-M365DSCPartialExport -Content $currentDSCBlock `
+                        -FileName $Global:PartialExportFileName
+                    Write-Host $Global:M365DSCEmojiGreenCheckMark
+            }
+            catch
+            {
+                Write-Verbose -Message $_
```
ykuijs, line 412: Please use the same logging logic as used on line 421 to 427. That will also output to the event log and other logs.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+        [System.String]
+        $GroupDisplayname,
+
+        [Parameter()]
+        [System.String]
+        $GroupId,
+
+        [Parameter()]
+        [System.String]
+        $PolicyType,
```
ykuijs, line 21: The PolicyType parameter has a specific set of possible values, as you can see here: https://learn.microsoft.com/en-us/powershell/module/teams/new-csgrouppolicyassignment?view=teams-ps#-policytype

Please make sure you use a ValidateSet, like is also done for the Ensure parameter, to make sure just this list of values can be provided.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+    #endregion
+
+    $nullReturn = $PSBoundParameters
+    $nullReturn.Ensure = 'Absent'
+
+    try
+    {
+        Write-Verbose -Message "get GroupPolicyAssignment for $GroupDisplayname"
+        $Group = Find-CsGroup -SearchQuery $GroupDisplayname
+        if($group.Length -gt 1)
```
ykuijs, line 74: A small code styling comment:
Please add a space between the keyword and the open parenthesis, so "if (....)".

This applies to multiple places in the code, so please check the entire file for this issue.

Modules/Microsoft365DSC/Examples/Resources/TeamsGroupPolicyAssignment/1-TeamsgrouppolicyAssignment.ps1
```diff
@@ -0,0 +1,28 @@
...
+        $credsGlobalAdmin
+    )
+    Import-DscResource -ModuleName Microsoft365DSC
+
+    node localhost
+    {
+        TeamsGroupPolicyAssignment "TeamsGroupPolicyAssignment"
+        {
+            Id					  = "1"
+			Ensure                = "Present";
```
ykuijs, line 19: Please make sure all parameters are aligned, have values that make sense and are valid. Examples are used in generating the documentation on the microsoft365dsc.com website, so must be as accurate and descriptive as possible.

ykuijs (changes requested):
Great work! I do have a few comments to look at.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
sandrola: Yes, you are right, in this case is no possible to find the right group. The second Filter is useful in case there are groups named "AllUser", "AllUserUS", "AllUserDE" and so on, with a Filter "AllUser" i will get it all, with the second filter -eq i will get only "AllUser". The -SearchQuery is like a -match.

PR #2753, opened 2023-01-06
Teams group policy assignment

sandrola:
Initiel release


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,439 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $Id,
```
NikCharlebois, line 9: ID shouldn't be mandatory since it is not something returned by the Get method. Instead, we should look at making the GroupDisplayName/PolicyName pair mandatory to uniquely identify an instance.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,439 @@
...
+
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $GroupDisplayname,
+
+        [Parameter()]
+        [System.String]
+        $GroupId,
```
NikCharlebois, line 17: We should modify the Get-TargetResource function to first try and retrieve the group instance by ID if it is provided. If the group can't be retrieved by ID, then we fallback to getting it by name.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,439 @@
...
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $Id,
+
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $GroupDisplayname,
```
NikCharlebois, line 13: To be consistent, please ensure title caps 'GroupDisplayName' with capital 'N' across the resource and its schema.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,439 @@
...
+
+    $CurrentValues = Get-TargetResource @PSBoundParameters
+
+    Write-Verbose -Message "Current Values: $(Convert-M365DscHashtableToString -Hashtable $CurrentValues)"
+    Write-Verbose -Message "Target Values: $(Convert-M365DscHashtableToString -Hashtable $PSBoundParameters)"
+
+    $TestResult = Test-M365DSCParameterState -CurrentValues $CurrentValues `
+        -Source $($MyInvocation.MyCommand.Source) `
+        -DesiredValues $PSBoundParameters `
+        -ValuesToCheck @('Ensure', `
```
NikCharlebois, line 325: Where possible, we should try to pass in $PSBoundParameters.Keys for the values to check. This ensures we don't forget to check any additional property we add to the resource in the future.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,439 @@
...
+        $j = 1
+        $totalCount = $instances.Length
+        foreach ($item in $instances)
+        {
+            $Group = Find-CsGroup -SearchQuery $item.GroupId
+            if ($null -eq $totalCount)
+            {
+                $totalCount = 1
+            }
+            Write-Host "    > [$j/$totalCount] GroupPolicyAssignment {$($Group.DisplayName)}" -NoNewline
```
NikCharlebois, line 394: All resources should use:

    |---

Instead of

     > 

When displaying the current instance being exported.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,439 @@
...
+            $results = @{
+                GroupDisplayname      = $Group.DisplayName
+                GroupId               = $Group.Id
+                PolicyType            = $item.PolicyType
+                PolicyName            = $item.PolicyName
+                Priority              = $item.Priority
+                Ensure                = 'Present'
+                Id                    = $j
```
NikCharlebois, line 402: Refer to comment above. Id shouldn't be a property of this resource if it's not something that exists on the M365 server side.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.schema.mof
```diff
@@ -0,0 +1,16 @@
+[ClassVersion("1.0.0.0"), FriendlyName("TeamsGroupPolicyAssignment")]
+class MSFT_TeamsGroupPolicyAssignment : OMI_BaseResource
+{
+    [Key, Description("Unique Id, used to identify the policy (DSC specific)")] string Id;
+    [Write, Description("Group Displayname of the group the policys are assigned to")] string GroupDisplayname;
```
NikCharlebois, line 5: Rename to GroupDisplayName with capital N

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.schema.mof
```diff
@@ -0,0 +1,16 @@
...
+    [Write, Description("Teams PolicyName. The name of the policy to be assigned.")] string PolicyName;
+    [Write, Description("Teams Priority. The rank of the policy assignment, relative to other group policy assignments for the same policy type")] string Priority;
+    [Write, Description("Present ensures the policy exists, absent ensures it is removed."), ValueMap{"Present","Absent"}, Values{"Present","Absent"}] String Ensure;
```
NikCharlebois, line 10: Description should mention "group policy assignment" instead of just "policy"

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/readme.md
```diff
@@ -0,0 +1,5 @@
+# TeamsGroupPolicyAssignment
+
+## Description
+
+This resource is used to add Teams policy to a group
```
NikCharlebois, line 5: Description should be updated to reflect group policy assignment

Modules/Microsoft365DSC/Examples/Resources/TeamsGroupPolicyAssignment/1-TeamsgrouppolicyAssignment.ps1
```diff
@@ -0,0 +1,28 @@
...
+        $credsGlobalAdmin
+    )
+    Import-DscResource -ModuleName Microsoft365DSC
+
+    node localhost
+    {
+        TeamsGroupPolicyAssignment "TeamsGroupPolicyAssignment"
+        {
+            Id					  = "1"
+			Ensure                = "Present";
```
NikCharlebois, line 19: Please align parameters

NikCharlebois (changes requested):
Few items to address. Thanks for your contributions!

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
NikCharlebois: Should we even keep ID as a parameter here since it doesn't exist on the "server"? I would recommend simply dropping it at this stage.

NikCharlebois: Maybe a better Key is Priority since it is unique?

NikCharlebois (changes requested):

sandrola:
I thin kDSC require a Key parameter that must be unique if multiple config are present i the same file. Id is only there for this reason.

NikCharlebois:
Would using the Priority parameter as the key make sense?

sandrola:
Priority can be the same with another PolicyType or PolicyName.

NikCharlebois:
Can it really though? I am trying to set two assignments to the same Rank/Priority and it automatically increases the number. The GUI gives you the impression that you can set it to be the same but doesn't enforces it it seems.

**Creating the policy and manually setting a priority that already exists for another policy/group**
image

**M365 automatically increased the value of the other ones**
image

sandrola:
If you have multiple assignemnt in the same config file and you have multiple priority "1"  and the Key is the priority, it cannot be compiled.

NikCharlebois:
Agreed, however based on the behavior of the M365 admin center, having multiple items with priority 1 is not possible, therefore we would expect the compilation to fail

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -0,0 +1,433 @@
...
+        [Parameter()]
+        [ValidateSet('CallingLineIdentity', 'TeamsAppSetupPolicy', 'TeamsAudioConferencingPolicy', 'TeamsCallingPolicy', 'TeamsCallParkPolicy', 'TeamsChannelsPolicy', 'TeamsComplianceRecordingPolicy', 'TenantDialPlan', 'TeamsMeetingBroadcastPolicy', 'TeamsMeetingPolicy', 'TeamsMessagingPolicy', 'TeamsShiftsPolicy', 'TeamsUpdateManagementPolicy', 'TeamsVerticalPackagePolicy')]
+        [System.String]
+        $PolicyType,
+
+        [Parameter()]
+        [System.String]
+        $PolicyName,
+
+        [Parameter()]
```
NikCharlebois, line 277: Now that Priority is a Key, we need to set it to Mandatory otherwise compilation will fail

NikCharlebois (changes requested):

NikCharlebois (approved):

PR #2823, opened 2023-01-20
DRG: CIM Instance improvement

### LINKED ISSUE #2819: DRG - generates schema where parameter Description-content can contain quotes
When I build a resource using DRG and the description for a parameter contains quotes, these are not replaced or edited meaning that the resulting mof-file is unreadable to DSC.

Location in the DRG-module where comments are added to the mof-definition: Lines 721 and 726, $mofSchemaContent.Appendline(... $property.Description)....). Code should be added to ensure the description is edited so quotes are replaced by apostrophes or something similar.
I don't have available time to do a PR myself but thought I'd bring it up.

```powershell
# insert configuration here
New-M365DSCResource....

dev

### PR DESCRIPTION
**Pull request by William-Francillette**

This PR improves CIM instance management:
- CIM constructor in Get-TargetResource
- Improved management of AdditionalProperties
- Improved DateTime, DateTimeOffset, Timespan management
- Fixed UnitTest
- Generate Stubs if required 
- Fixes #2819 

Tested with a fairly complicated resource including nested CIMs + dates + enums: MgEntitlementManagementAccessPackageAssignmentPolicy

### CODE REVIEW

File: Tests/Unit/Stubs/Microsoft365.psm1 [hunk 1]
```diff
@@ -17179,7 +17179,8 @@ function Update-MgDeviceManagementIntent
 function Update-MgDeviceManagementIntentSetting
```
NikCharlebois, line 17179: Just as an FYI, this file is automatically generated on a regular basis by the New-M365DSCStub cmdlet. There is a chance that any manual change get overwritten.

NikCharlebois (approved):

PR #2838, opened 2023-01-23
PR: IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled - Initial release

### LINKED ISSUE #2604: [Feature] Intune - App protection policies - Windows 10 WIP policy
### Description
In Intune/Endpoint there the option to create windows 10 app protection policies, but with M365DSC there isn't an option to export and import those
### Proposed properties

- Endpoint ManagerApps - App Protection Policies - Create -Windows Information Protection

William-Francillette:

Fixes #2604 
Initial release fully generated via DRG - only modified test file due to nested array within a Cim instance issue with Pester

@NikCharlebois @ykuijs @andikrueger, could check the code thoroughly and let me know if I need to change anything in the DRG please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@
   * Fixed UnitTest
   * Generate Stubs if required
   * Fixes #2819
+* IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled
+  * Initial Release
+  * Fixes #2604
 * DEPENDENCIES
   * Updated Microsoft.PowerApps.Administration.PowerShell to version 2.0.155.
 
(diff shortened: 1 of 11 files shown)


NikCharlebois:
Looks good. As always, the only thing that worries me is the generic nature of the CIMInstances. We should try to name them with the actual resource's name as a prefix ven if that makes them very long. This is to prevent having two resources in the future implementing the same CIM Instance class name which would cause issues. For example, I could see two resources implementing a class named:
MicrosoftGraphIpRange

William-Francillette:
I see so rather to add the workload as a prefix we could use the resource name 
The name could become long name though and may also impact the export file for nested CimInstances

Is there a way to list existing CimInstances? That way we could ensure that they are unique by adding a counter/digit at the end and keeping the name as short as possible.

I also encountered issue when a parameter is using a keyword used by the DSC engine - the error generated is not obvious at all if I reckon properly - The other day I was working on a resource and couldn't build the mof file because I had a parameter called Definition - The error message was unspecified cim module 
I should also add a logic to rename such parameter - Is there a list of keyword?

William-Francillette:
Couldn't find a way to get all cim classes from the DSC Engine but can retrieve that by extracting the text from the mof files - I will implement the logic tonight after work and add a counter at the end of the cimInstance to prevent any duplicate - I'll update this PR please don't merge it yet

NikCharlebois:
Thanks!

William-Francillette:
@NikCharlebois it's good to go  - if the cim class is already present in any mof files in the module, it will add a counter

NikCharlebois (approved):

PR #2920, opened 2023-02-17
Fix for #2918 (and probably #2198)

### LINKED ISSUE #2198: Exporting Configuration Profiles Intune Workload
Hi Team,
Just wondering if configuration profiles should be exported as part of the INTUNE workload?

I have managed to get everything else exported / imported except for those.

Just wondering if it is a limitation of the tool or not.

### LINKED ISSUE #2918: IntuneDeviceConfigurationPolicyiOS and IntuneAppProtectionPolicyiOS: Missing mandatory parameter causes `Get-M365DSCResourceKey` to fail
When exporting *IntuneDeviceConfigurationPolicyiOS* and *IntuneAppProtectionPolicyiOS*, `Get-M365DSCResourceKey` fails because no parameter is mandatory. 

This causes: 
```
PSMessageDetails      :
Exception             : System.Management.Automation.RuntimeException: You cannot call a method on a null-valued expression.
                           at ....
ScriptStackTrace      : at Get-M365DSCResourceKey, C:\Users\User\GITs\Microsoft365Dsc\modules\M365DSCReport.psm1: line 724
                        at Compare-M365DSCConfigurations, C:\Users\User\GITs\Microsoft365Dsc\modules\M365DSCReport.psm1: line 657
                       .....
PipelineIterationInfo : {}
```

The problem is may come at any line where the first argument is true. If it tries  `$mandatoryParameters.Name.Contains('Something'))` and that is `$true` while `$mandatoryParameters` is $null, the line fails with the above exception:  
```powershell
if ($Resource.Contains('IsSingleInstance') -and $mandatoryParameters.Name.Contains('IsSingleInstance')) # This may fail
    {
        return @('IsSingleInstance')
    }
    elseif ($Resource.Contains('DisplayName') -and $mandatoryParameters.Name.Contains('DisplayName')) # This may fail
    ....
```

Note that this may be somewhat related to #2198 as well.

n/a

Use same mandatory params as W10

Just Exporting

n/a
Dev

bjoernf73:
Since exports uses the `Get-M365DSCResourceKey`, each DSC resource must have a mandatory key (or property). No mandatory key was defined for the IntuneDeviceConfigurationPolicyiOS and IntuneAppProtectionPolicyiOS resources, causing  `Get-M365DSCResourceKey` to fail. 

It seems there is a convention not to throw errors, but rather write errors *somewhat quietly* to the error-stream, so this may go *under the radar* if you run in a pipeline, for instance. Custom scripts that calls functions from Microsoft365DSC may use `-ErrorVariable MyErrorVariable` - all errors will be visable there.

I believe this PR will resolve #2198 as well, in addition to my own issue #2918. 

- Fixes #2918
- Fixes #2198


ykuijs:
PR #2968 replaces this PR since this PR only fixes one resource and the other PR fixes all resources that have missing Key parameters. It also adds a QA check to validate that all resources have Key parameters. Closing this PR now.

PR #2968, opened 2023-03-06
[BREAKING CHANGE] Added QA check for presence of Key parameter

### LINKED ISSUE #2925: Multiple resources missing mandatory parameters causing `Get-M365DSCResourceKey` to fail
When creating DeltaReports, `Get-M365DSCResourceKey` is used, but fails on resources that are not defined with mandatory params.
```powershell
ipmo Microsoft365DSC
$all = Get-DscResource -Module Microsoft365DSC
(0..$($all.count-1)) |  
foreach { 
    $rname = $all[$_].Name;
    if (($all[$_].Properties | where { 
        $_.IsMandatory -eq $true 
    }).count -eq 0) {
        $rname 
    }
}
```
outputs: 
```
EXOIRMConfiguration
EXOPerimeterConfiguration
EXOResourceConfiguration
IntuneAppProtectionPolicyiOS
IntuneDeviceConfigurationPolicyAndroidDeviceAdministrator
IntuneDeviceConfigurationPolicyAndroidDeviceOwner
IntuneDeviceConfigurationPolicyAndroidOpenSourceProject
IntuneDeviceConfigurationPolicyIOS
IntuneDeviceConfigurationPolicyMacOS
IntuneWifiConfigurationPolicyAndroidDeviceAdministrator
IntuneWifiConfigurationPolicyAndroidEntrepriseDeviceOwner
IntuneWifiConfigurationPolicyAndroidEntrepriseWorkProfile
IntuneWifiConfigurationPolicyAndroidForWork
IntuneWifiConfigurationPolicyAndroidOpenSourceProject
IntuneWifiConfigurationPolicyIOS
IntuneWifiConfigurationPolicyMacOS
IntuneWifiConfigurationPolicyWindows10
IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled
AADEntitlementManagementAccessPackage
AADEntitlementManagementAccessPackageCatalog
AADEntitlementManagementAccessPackageCatalogResource
```

n/a

Each resource should define at least one mandatory parameter, but as a quick fix, perhaps:
```powershell
$resourceInfo = Get-DscResource ("MSFT_$($Resource.ResourceName)") -Module 'Microsoft365DSC'
[Array]$mandatoryParameters = $resourceInfo.Properties | Where-Object -FilterScript { $_.IsMandatory }
# for resources where no mandatory params defined: 
if ($null -eq $mandatoryParameters)
{
    Write-Warning "Resource $($Resource.ResourceName) does not have any mandatory parameters, `
    meaning there is still work to be done with this resource. This function may return a plausible identifying property, `
    if such a parameter can be found"
    
    # is it a single instance resource, at least return that
    if ($Resource.Contains('IsSingleInstance')) {
        return @('IsSingleInstance') 
    }

    # return any plausible identifying param
    if ($Resource.Contains('DisplayName'))
    {
        return @('DisplayName')
    }
    elseif ($Resource.Contains('Name'))
    {
        return @('Name')
    }
    elseif ($Resource.Contains('Identity'))
    {
        return @('Identity')
    }
    elseif ($Resource.Contains('Id'))
    {
        return @('Id')
    }
    else
    {
        throw "Resource $($Resource.ResourceName) does not have any mandatory parameters, and no other unique identifier could be guessed"
    }
  }
  elseif ($Resource.Contains('IsSingleInstance') -and $mandatoryParameters.Name.Contains('IsSingleInstance'))
  .....
```

(Any call to New-M365DSCDeltaReport that includes one or more of the resources listed above)

n/a

Dev

ykuijs:
This PR adds a QA check to test for the presence of a Key parameter. Since this can break existing configurations, we have to merge this for the next Breaking Changes release of April 5th.

- Fixes #2925


ykuijs:
This PR replaces PR #2920. Where that PR fixes one single resource is this PR fixing all resources that have missing Key parameters.

PR #2973, opened 2023-03-07
AADAdministrativeUnit : Updated schema wrt Identity and ScopedRoleMembership

salbeck-sit:
Updated schema-property Type for CIMInstance MSFT_MicrosoftGraphIdentity used in AADAdministrativeUnit to enable validation of content while editing a configuration. Also, for clarity, added link to show available roles in description of property RoleName for CIMInstance MSFT_MicrosoftGraphScopedRoleMembership

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change log for Microsoft365DSC
 
 # UNRELEASED
+* AADAdministrativeUnit
+  * Updated validation of properties in schema to assist usage
 
 * AADAdministrativeUnit
   * Fixed general issues caused by improper handling of nested CIMInstances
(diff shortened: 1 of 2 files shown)


ykuijs:
Hi @salbeck-sit, can you please also add an entry to the changelog?

And isn't this a breaking change since it is limiting the possible inputs? Existing configurations that use this resource might break because of this.

salbeck-sit:
It's not a breaking change per se in that the valuemap contains the values that Graph will accept. The change is simply meant to assist usage given that the resource-page on the M365 website could arguably be more helpful. I'll update the changelog

ykuijs (approved):

PR #3057, opened 2023-03-22
TeamsGroupPolicyAssignment Exported configuration cannot be used beca…

andikrueger:
…use of duplicated key value

- Fixes #3054

PR #3060, opened 2023-03-22
Add support for multiple Rank / group in the same configuration.

sandrola:
Fix Issue with duplicated Key value in exported configuration
Fixes #3057


ykuijs:
@sandrola, this PR looks like a duplicate of #3057. Is that true?

sandrola:
Ok, I have see it now, but in #3057 the Groupid is now a Key, this eliminate the possibility to use the same configuration for multiple Tenant using the Group Displayname.

andikrueger:
Thank you for this PR and the second look at this issue. I would take the discussion back to the initial issue #3057 and hold both PRs until we do have a common understanding.

NikCharlebois:
We need to make sure that users writing the configuration manually aren't forced providing an ID, which doesn't exist just for configurations to compile. My recommendation is to put #3057 on hold and consider merging this PR here instead.

ykuijs:
#3057 was merged. Closing this PR

PR #3105, opened 2023-03-31
Fix issue if External Guest not set cause empty Odata type

sandrola:
Fix issue with Empty ExternalGuest causes an empty OdataType


Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1
```diff
@@ -1125,55 +1125,49 @@ function Set-TargetResource
...
-        }
-        else
-        {
-            $externalTenants.Add('@odata.type', '')
-        }
-        $externalTenants.Add('membershipKind', $IncludeExternalTenantsMembershipKind)
-        if ($IncludeExternalTenantsMembers)
-        {
-            $externalTenants.Add('members', $IncludeExternalTenantsMembers)
+        if($IncludeGuestOrExternalUserTypes){
```
ykuijs, line 1128: Please follow the code styline guidelines for this project by placing a space behind the if-statement and the open brace on the next line. Also, wouldn't it be better to check if the count is not equal to zero:

```PowerShell
if ($IncludeGuestOrExternalUserTypes.Count -ne 0)
{
    # The code here
}
```

ykuijs (changes requested):
Please also add an entry in the changelog.

And could you also create an issue in the issuelist for the problem you are fixing here, including the details and potentially errors you are getting. That way we can track the issues that are being fixed.

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Change log for Microsoft365DSC
 
 # UNRELEASED
-
+* AADConditionalAccessPolicy
+  * Add condition for empty External Guest/User include/exclude
```
ykuijs, line 5: There is already an entry for the AADConditionalAccessPolicy header on line 25. Please add this item to that header.

ykuijs (changes requested):

ykuijs (approved):
LGTM!

PR #3123, opened 2023-04-03
Fixes extraction of the ScopedRoleMembers property

ricmestre:

Commit f60f2b9434db70f86374f9f2b22c612dc96549d8 solved the extraction for the Members property, but ScopedRoleMembers was not changed, this fixes it by using the correct CIM instance name.


ykuijs (approved):
LGTM

PR #3155, opened 2023-04-11
Remove ValidateScript from param Credential in Export-M365DSCConfiguration

### LINKED ISSUE #3153: Assert-M365DSCBlueprint error with ver 1.23.405.1 using cert auth
Script [script.txt](https://github.com/microsoft/Microsoft365DSC/files/11199792/script.txt) using Assert-M365DSCBlueprint and certificate thumbprint auth works with 1.23.322.1. Fails with 1.23.405.1

Exported ODsettings are very basic but all exports fail with the same errors 
[od.ps1.txt](https://github.com/microsoft/Microsoft365DSC/files/11199790/od.ps1.txt)

```
PS C:\github\M365DSC\Assess> .\CompareTargetWithBlueprint.ps1
Generate blueprint assessment for OD.ps1
VERBOSE: Dependencies were already successfully validated.
VERBOSE: GET file:///C:/Source/OD.m365 with 0-byte payload
VERBOSE: received 1704-byte response of content type application/octet-stream
Selected BluePrint contains (1) components to assess.
Initiating the Export of those (1) components from the tenant...
Export-M365DSCConfiguration : Cannot validate argument on parameter 'Credential'. The argument is null, empty, or an element of the argument collection contains a null
value. Supply a collection that does not contain any null values and then try the command again.
At C:\Program Files\WindowsPowerShell\Modules\Microsoft365DSC\1.23.405.1\modules\M365DSCUtil.psm1:2663 char:25
+             -Credential $Credentials `
+                         ~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Export-M365DSCConfiguration], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Export-M365DSCConfiguration

New-M365DSCDeltaReport : Cannot find file specified in parameter Source: C:\Users\LOCALA~1\AppData\Local\Temp\c54d7565-72e6-46c0-bf33-ecf7a9306a74.ps1. Please make sure
the file exists!
At C:\Program Files\WindowsPowerShell\Modules\Microsoft365DSC\1.23.405.1\modules\M365DSCUtil.psm1:2674 char:9
+         New-M365DSCDeltaReport -Source $ExportPath `
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,New-M365DSCDeltaReport

Processing time for OD.ps1 is 00:00:21.5608381
Processing time was 00:00:21.5608381
PS C:\github\M365DSC\Assess>
```

ricmestre:
Remove ValidateScript from param Credential, in function Export-M365DSCConfiguration, since it may be null when using App creds, and validation will throw an error, instead do the validation afterwards in the body of the function itself.

- Fixes #3153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
       FIXES [#3068](https://github.com/microsoft/Microsoft365DSC/issues/3068)
 * TeamsUpdateManagementPolicy
   * Added support for the Forced value for the AllowPublicPreview property.
+* MODULES
+  * M365DSCUtil: Fixed an issue when calling Assert-M365DSCBlueprint with App credentials
+    FIXES [#3153](https://github.com/microsoft/Microsoft365DSC/issues/3153)
 
 # 1.23.405.1
 
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -1095,7 +1095,7 @@ function Export-M365DSCConfiguration
                 }
                 else
                 {
-                    Write-Warning -Message "We recommend providing the TenantId property in the format of <tenant>.onmicrosoft.*"
+                    Write-Host -Object "[WARNING] We recommend providing the TenantId property in the format of <tenant>.onmicrosoft.*" -ForegroundColor Yellow
                 }
             }
             return $true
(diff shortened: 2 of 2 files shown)


ykuijs (changes requested):
Could you please add an entry to the changelog as well?

ricmestre:
Changelog is updated now.

andikrueger:
Just out of curiosity: Isn't this a design flaw on the Assert-Function? Should we not better check there, if the credential object is null and in this case remove the parameter from the call to the Export function?

ykuijs:
@andikrueger I understand what you mean here. In the Assert function all parameters are passed into the Export function regardless of if they are supplied to the Assert function. That way the Export function always received the parameters that are passed into the Assert function. It requires more code in the Assert function if we want to make it dynamic, which the Export function already handles properly.

Since this PR also changes some Write-Warning into Write-Host, so they are displayed to the users. I am going to merge this PR.

ykuijs (approved):
LGTM

ricmestre:
@andikrueger Hi, yes that would also work for the Assert function, the problem is that a user could possibly still manually invoke Export with "... -Credential $null ..." and trip the error. If that's acceptable I can cook up a diff with your suggestion and putting back ValidateScript.

PR #3164, opened 2023-04-13
SCLabelPolicy: If label policy is set to None don't get its label display name since it's not required

### LINKED ISSUE #3104: DSC SCLabelPolicy export error
```powershell
# At the end in pictures


```
Newest version from 03.28.23

#### Pictures
ErrorInPS
ErrorlogOutput

ricmestre:
SCLabelPolicy: If label policy is set to None don't get its label display name since it's not required. This solves an issue when for instance a label policy has outlookdefaultlabel set to None, there is no label with such name and therefore Get-Label fails.

- Fixes #3104


ricmestre:
With latest commit I'm able to export outlookdefaultlabel set to "None", changing it to "All employees (unrestricted)" and then back to "None" without errors.

PR #3166, opened 2023-04-13
Exclude Auth params from Reports

NikCharlebois:
* MISC
  * Reports will now exclude the authentication parameters (e.g., CertificateThumbprint, Credential, etc.).

* N/A


ricmestre:
@NikCharlebois Hi Nik, shouldn't ApplicationSecret also be excluded here?

NikCharlebois:
Absolutely, that was an oversight. Thanks

PR #3218, opened 2023-04-25
[SCLabelPolicy] Fixing issue with the Labels parameter being ignored for existing policies

### LINKED ISSUE #3216: [SCLabelPolicy] Labels aren't updated on existing policies when using Labels property

When you want to update an existing label policy, the resource doesn't do anything when using the Labels property. This is because the Labels parameter is removed in the code, since the Set-LabelPolicy cmdlet doesn't support the Labels parameter. 

N/A

We need to change the code to populate the AddLabels/RemoveLabels, based on the labels specified in Labels and the actually applied labels

```powershell
# insert configuration here
```

Dev

ykuijs:
This PR fixes an issue in SCLabelPolicy where the Labels parameter is ignored for existing policies

- Fixes #3216


andikrueger (approved):
LGTM

Issue #3259, opened 2023-05-02
IntuneDeviceConfigurationAdministrativeTemplatePolicyWindows10: Issue with ConvertTo-DSCObject


ricmestre:
#### Details of the scenario you tried and the problem that is occurring

Given the blueprint c:\intune.txt, taken from #3214 and provided here at the bottom, then invoking ConvertTo-DSCObject on it will choke with error message 'Exception calling "Add" with "2" argument(s): "Item has already been added...'. Is this an issue that can be solved in the Resource itself or in DSCParser module?

#### Verbose logs showing the problem

```powershell
PS C:\> ConvertTo-DSCObject -Path C:\intune.txt

Exception calling "Add" with "2" argument(s): "Item has already been added. Key in dictionary: 'Id'  Key being added: 'Id'"
At C:\Program Files\WindowsPowerShell\Modules\DSCParser\1.3.0.7\modules\DSCParser.psm1:383 char:17
+                 $result.Add($token.Content, "")
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : ArgumentException

Exception calling "Add" with "2" argument(s): "Item has already been added. Key in dictionary: 'Id'  Key being added: 'Id'"
At C:\Program Files\WindowsPowerShell\Modules\DSCParser\1.3.0.7\modules\DSCParser.psm1:383 char:17
+                 $result.Add($token.Content, "")
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : ArgumentException

Exception calling "Add" with "2" argument(s): "Item has already been added. Key in dictionary: 'Id'  Key being added: 'Id'"
At C:\Program Files\WindowsPowerShell\Modules\DSCParser\1.3.0.7\modules\DSCParser.psm1:383 char:17
+                 $result.Add($token.Content, "")
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : ArgumentException

Exception calling "Add" with "2" argument(s): "Item has already been added. Key in dictionary: 'Id'  Key being added: 'Id'"
[...]

PR #3262, opened 2023-05-02
Assert-M365DSCBlueprint: Add support to exclude resources from being compared

ricmestre:

When running Assert-M365DSCBlueprint one might want to just compare policies and not content resources such as AADUser, AADGroup, SPOSite, etc. This PR adds a new parameter to the function which allows to exclude certain types of resources from the report.

As an example these resources could be excluded from being compared:

```powershell
$ExcludedResources = @($Global:DefaultComponents, $Global:FullComponents)
```

Additionally the function now throws an exception if ConvertTo-DSCObject fails, this is based on the issue #3259 I raised a couple of hours ago.


NikCharlebois:
Would it also make sense to add support for it in the New-M365DSCDeltaReport while we're at it?

ricmestre:
Sure, just added support for it on both New-M365DSCDeltaReport and Compare-M365DSCConfigurations.

NikCharlebois (approved):

PR #3286, opened 2023-05-10
Fix issues with Ensure property on several resources

ricmestre:

Currently several resources have $Ensure set to $true whereas it should be a string set to 'Present' as ValidateSet already expects, additionally it shouldn't be mandatory since it always has 'Present' as its default value.

This PR fixes this by ensuring that $Ensure is always set to 'Present' by default and it's not mandatory.

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_EXOJournalRule/MSFT_EXOJournalRule.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_EXOJournalRule/MSFT_EXOJournalRule.psm1
@@ -158,7 +158,7 @@ function Set-TargetResource
         [Parameter()]
         [ValidateSet('Present', 'Absent')]
         [System.String]
-        $Ensure = $true,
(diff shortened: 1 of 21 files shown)


andikrueger:
Could you do a quick check if this is related to the DRG Template?

ricmestre:
I checked DRG prior to making the PR, and no it's not because of that, this seems to be resources that were created manually.

ricmestre:
Actually it could have been, I went far back in history and the DRG had both problems but now they are not present so it's fixed there already.

ricmestre:
@andikrueger The unit testing failed but it seems to be some internal issue not related to my changes.

andikrueger:
image
 Indeed.

ricmestre:
It's most likely related to https://www.githubstatus.com/incidents/pr3498h3qkfy

andikrueger:
Yes. Thanks for checking. This is the root cause right now. We will have to wait for this github issue to be fixed.

PR #3338, opened 2023-05-24
Release 1.23.524.1

NikCharlebois:


salbeck-sit:
It seems to me that the module-version wasn't incremented?

PR #3356, opened 2023-06-02
Fixing error Policy  already has rule associated

max-freshmethod:

Fixing error "Policy already has a rule associated with it".

If we leave HostedContentFilterPolicy when we update the existing HostedContentFilterRule, the next error will appear, so we need to remove HostedContentFilterPolicy from the parameters that we pass to the Set-HostedContentFilterPolicy 

"Write-ErrorMessage : |Microsoft.Exchange.Management.Tasks.ValidationException|Policy "policy name" already has rule "Rule name" associated with it."

None

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_EXOHostedContentFilterRule/MSFT_EXOHostedContentFilterRule.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_EXOHostedContentFilterRule/MSFT_EXOHostedContentFilterRule.psm1
@@ -326,6 +326,7 @@ function Set-TargetResource
         $UpdateParams.Remove('CertificatePassword') | Out-Null
         $UpdateParams.Remove('ManagedIdentity') | Out-Null
         $UpdateParams.Remove('Enabled') | Out-Null
+        $UpdateParams.Remove('HostedContentFilterPolicy') | Out-Null
         Write-Verbose -Message "Updating HostedContentFilterRule {$Identity}"
         Set-HostedContentFilterRule @UpdateParams
     }


Modules/Microsoft365DSC/DSCResources/MSFT_EXOHostedContentFilterRule/MSFT_EXOHostedContentFilterRule.psm1
```diff
@@ -326,6 +326,7 @@ function Set-TargetResource
         $UpdateParams.Remove('CertificatePassword') | Out-Null
         $UpdateParams.Remove('ManagedIdentity') | Out-Null
         $UpdateParams.Remove('Enabled') | Out-Null
+        $UpdateParams.Remove('HostedContentFilterPolicy') | Out-Null
```
NikCharlebois, line 329: Set-HostedContentFilterRule can still be called with -HostContentFilterPolicy passed in if it doesn't already have an associated rule. We should check to see if the policy has an associated rule and ensure it's not the default policy before deciding whether or not to remove the parameter.

NikCharlebois (changes requested):
Please review comment. Thanks for your help

Modules/Microsoft365DSC/DSCResources/MSFT_EXOHostedContentFilterRule/MSFT_EXOHostedContentFilterRule.psm1
max-freshmethod: OK, I got your point, let me do a few tests

max-freshmethod:
@NikCharlebois It should be better now

NikCharlebois (approved):

PR #3357, opened 2023-06-02
Handling GlobalQuarantinePolicy update

max-freshmethod:
We need to handle the situation when we update GlobalQuarantinePolicy because we can't use Set-QuarantinePolicy directly to update the policy, so we need to call Get-QuarantinePolicy and pipe it to Set-QuarantinePolicy.

We also need to remove the Identity parameter, otherwise, we will get an error.
https://learn.microsoft.com/en-us/powershell/module/exchange/set-quarantinepolicy?view=exchange-ps#-esncustomsubject

None


NikCharlebois (changes requested):
Thanks for the contribution, can I please ask you to add a new entry in the changelog to describe your change and then I'll merge. Thanks

max-freshmethod:
@NikCharlebois Just added it. Thanks.

NikCharlebois (approved):

andikrueger:
@max-freshmethod Thanks for your contribution! Could you please resolve the changelog conflict to get it ready to be merged.

Issue #3335, opened 2023-05-24
Issues exporting with several resources in Intune that have embedded instances


ricmestre:
#### Details of the scenario you tried and the problem that is occurring

Exporting a blueprint from my production tenant with latest version of M365Dsc results in several resources having a problem with embedded instances and if the array has more than one object where it adds a comma in-between them whereas with earlier versions this problem was not present (can't recall which version I tested that worked with it), the problem is that this is not happening with all of them so there's no pattern here, but it could also be happening with other resources that I'm currently not using and I don't mention below.

From the resources I have and that I could test these are currently the ones that I've seen affected/not affected:

Affected (with MSFT_DeviceManagementConfigurationPolicyAssignments):

IntuneApplicationControlPolicyWindows10
IntuneDeviceConfigurationDeliveryOptimizationPolicyWindows10
IntuneDeviceConfigurationEndpointProtectionPolicyWindows10 
IntuneDeviceConfigurationHealthMonitoringConfigurationPolicyWindows10 
IntuneDeviceConfigurationIdentityProtectionPolicyWindows10 
IntuneWindowsUpdateForBusinessFeatureUpdateProfileWindows10 
IntuneWindowsUpdateForBusinessRingUpdateProfileWindows10 

Affected (with MSFT_MicrosoftGraphwindowsInformationProtectionApp):
IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled 

Not affected (with MSFT_DeviceManagementConfigurationPolicyAssignments):

IntuneASRRulesPolicyWindows10
IntuneDeviceCompliancePolicyAndroidDeviceOwner
IntuneDeviceCompliancePolicyiOs
IntuneDeviceCompliancePolicyWindows10
 IntuneDeviceConfigurationAdministrativeTemplatePolicyWindows10 
IntuneDeviceConfigurationPolicyiOS 
IntuneDeviceConfigurationPolicyWindows10 

#### Verbose logs showing the problem

```powershell
        IntuneApplicationControlPolicyWindows10 "IntuneApplicationControlPolicyWindows10-REDACTED"
        {
            ApplicationId                    = $ConfigurationData.NonNodeData.ApplicationId;
            AppLockerApplicationControl      = "auditComponentsStoreAppsAndSmartlocker";
            Assignments                      = @(
                MSFT_DeviceManagementConfigurationPolicyAssignments{
                    deviceAndAppManagementAssignmentFilterType = 'none'
[...]

PR #3362, opened 2023-06-05
PR: Fixes 3335

William-Francillette:

* Intune Resources
  * Fixes an issue with trailing charaters in the export job
    FIXES [#3335](https://github.com/microsoft/Microsoft365DSC/issues/3335)
- IntuneApplicationControlPolicyWindows10
- IntuneDeviceConfigurationDeliveryOptimizationPolicyWindows10
- IntuneDeviceConfigurationEndpointProtectionPolicyWindows10
- IntuneDeviceConfigurationHealthMonitoringConfigurationPolicyWindows10
- IntuneDeviceConfigurationIdentityProtectionPolicyWindows10
- IntuneWindowsUpdateForBusinessFeatureUpdateProfileWindows10
- IntuneWindowsUpdateForBusinessRingUpdateProfileWindows10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 * IntuneAntivirusPolicyWindows10SettingCatalog
   * Fixes an issue for policies with template endpointSecurityAntivirus that had a templateId not expected by the code
     FIXES [#3360](https://github.com/microsoft/Microsoft365DSC/issues/3360)
+* Intune Resources
+  * Fixes an issue with trailing charaters in the export job
+    FIXES [#3335](https://github.com/microsoft/Microsoft365DSC/issues/3335)
 
 # 1.23.524.1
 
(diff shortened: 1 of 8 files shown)


ricmestre:
@William-Francillette LGTM but it still misses the same chunk of code, inserted below, in one of the affected resources I reported (IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled). As I initially reported there might be other resources affected but these were the ones I was able to figure out for now, if I find others I'll let you know.

https://github.com/microsoft/Microsoft365DSC/blob/0866c08d62737cf0cd535e754c9a0ea2f31d35f5/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled/MSFT_IntuneWindowsInformationProtectionPolicyWindows10MdmEnrolled.psm1#L1161

ricmestre:
A quick grep shows me that both IntuneWindowsAutopilotDeploymentProfileAzureADHybridJoined and IntuneWindowsAutopilotDeploymentProfileAzureADJoined are also affected.

ricmestre:
@William-Francillette Hi, please have a look at my PR #3364 , as mentioned there "Chunk of code added to all resources that call Convert-DSCStringParamToVariable and that at least one of those calls has an object with an array of CIM instances (or may have depending on the exported resource)."

ricmestre:
Yes I saw your discussion with Andi about this on the other thread, but at least for the currently available resources I need this resolved since I'm soon going to production and this is one of the major issues I've still have opened so I'd like to get this in ASAP.

NikCharlebois:
Would #3364 also address this. I want to make sure we only fix this in a central location and to me, fixing it in ReverseDSC is the logical thing to do.

ricmestre:
@NikCharlebois I have a PR opened in ReverseDSC to address this issue https://github.com/microsoft/ReverseDSC/pull/33, as mentioned there #3362 and #3364 can be cancelled once that fix is in.

William-Francillette:
Makes more sense to fix ReverseDSC imo - this can be cancelled if that's the case

PR #3375, opened 2023-06-09
Remove Az PowerShell dependencies

NikCharlebois:
* Removing all Az.Accounts dependancies in the module

* N/A

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
@@ -1354,27 +1354,19 @@ function Update-M365DSCAzureAdApplication
         return
     }
 
-    $context = Get-AzContext
-    if ($null -eq $context)
-    {
-        Connect-AzAccount
-        $context = Get-AzContext
-    }
-
     $resourceAppIdMsGraph = '00000003-0000-0000-c000-000000000000'
     $resourceAppIdSharePoint = '00000003-0000-0ff1-ce00-000000000000'
     $resourceAppIdExchange = '00000002-0000-0ff1-ce00-000000000000'
 
-    $allPrincipals = Get-AzADServicePrincipal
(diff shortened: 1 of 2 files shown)


NikCharlebois:
@andikrueger & @ykuijs would appreciate your review on this one. Removing dependency on Az.Accounts in favor of Microsoft.Graph.Application

Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
```diff
@@ -1354,28 +1358,21 @@ function Update-M365DSCAzureAdApplication
         return
```
ykuijs, line 1358: This section can be removed since we are no longer using Az.Resources

Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
```diff
@@ -1354,28 +1358,21 @@ function Update-M365DSCAzureAdApplication
...
     }
 
-    $context = Get-AzContext
-    if ($null -eq $context)
-    {
-        Connect-AzAccount
-        $context = Get-AzContext
-    }
-
```
ykuijs, line 1363: Shouldn't we include some authentication to the Graph befor calling the Get-MgServicePrincipal cmdlet?

ykuijs (changes requested):
Two small comments

William-Francillette:
Should the Az.Accounts not be part of MSCloudLoginAssistant dependencies as the Azure workload for example uses Connect-AzAccount?
image

andikrueger:
Non of the supported workloads is part of the manifest. This is due to the critical load times of all the modules...

William-Francillette:
Oh I see, thanks for the clarification @andikrueger

NikCharlebois:
The Azure workload is never used as part of M365DSC so we should be good on that front.

NikCharlebois:
@ykuijs are we good to merge in tomorrow's release? Thanks

ykuijs (approved):
lgtm

PR #3461, opened 2023-07-11
Update Manifest.psd1

NikCharlebois:
* Removed dependency for Microsoft.Graph.Beta.Identity.DirectoryManagement since we already require the beta module.

* N/A

### DIFF
--- a/Modules/Microsoft365DSC/Dependencies/Manifest.psd1
+++ b/Modules/Microsoft365DSC/Dependencies/Manifest.psd1
@@ -53,10 +53,6 @@
             RequiredVersion = '2.0.0'
         },
         @{
-            ModuleName      = 'Microsoft.Graph.Identity.DirectoryManagement'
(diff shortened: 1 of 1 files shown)


NikCharlebois:
@William-Francillette somehow I missed this in your PR. Since we are already requiring the Microsoft.Graph.Beta.Identity.DirectoryManagement module, we shouldn't be also requiring the v1.0 version. Can you please confirm that this won't have any impact on your stuff? Thanks

William-Francillette:
@NikCharlebois, it will break AADAdministrativeUnit as developed using v1.0
I can migrate teh feature and test it quickly - it shouldn't take too long will raise a new PR

NikCharlebois:
My understanding is that AADAdministrativeUnit is already updated to run on Beta.

William-Francillette:
> My understanding is that AADAdministrativeUnit is already updated to run on Beta.

My bad I missed that so we should be good to go

PR #3474, opened 2023-07-14
Fixes for issues in Pipelines with Graph PS 2.0.0

### LINKED ISSUE #3454: Azure Pipelines on DevOps with certificate-based authentication and Graph v2.0 not working

Note: I know this is not a problem with M365Dsc itself, but that's what I'm using in the pipelines and I don't know where else to report the problem.

Currently Windows containers running on Azure Pipelines on DevOps have Graph v1.28 pre-installed, when installing the latest M365Dsc version v1.23.705.1 they get updated to v2.0 (Update-M365DSCDependencies) and whatever is outdated is uninstalled (Uninstall-M365DSCOutdatedDependencies).

Nevertheless calling simple cmdlets such as Get-MgUser or Get-MgGroup ends up in error, see below, I cannot test every single resource to check which ones are affected by this but my solution relies on calling Get-MgGroup to translate GroupIds to DisplayNames and vice-versa due to the way Assignments work on Intune resources (only by GroupId) which makes upgrading not an option, which at the same time makes me also not able to test the other issues I reported with the latest version.

I already tested this unsuccessfully:
	- Removing all Microsoft.Graph.* v1.28 from the session
	- Removing all their associated folders from "C:\Program Files\WindowsPowerShell\Modules"
	- Install Microsoft.Graph.Authentication v2.0.0
	- Install Microsoft.Graph.Groups v2.0.0
	- Install MSCloudLoginAssistant v1.0.116
	- Connect to Graph using Connect-M365Tenant with certificate
	- Calling Get-MgGroup, error happens here

```powershell
##[error]Method not found: 'Void Microsoft.Graph.Authentication.AzureIdentityAccessTokenProvider..ctor(Azure.Core.TokenCredential, System.String[], Microsoft.Kiota.Authentication.Azure.ObservabilityOptions, System.String[])'
```

Update Windows container image to have Graph v2.0 by default instead of v1.28.0.

N/A

OsName               : Microsoft Windows Server 2022 Datacenter
OsOperatingSystemSKU : DatacenterServerEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsBuildLabEx    : 20348.1.amd64fre.fe_release.210507-1500
OsLanguage           : en-US
OsMuiLanguages       : {en-US}

1.23.705.1

andikrueger:

This PR updates Graph PS to V 2.1.0

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -2761,6 +2761,12 @@ function Update-M365DSCDependencies
         $ValidateOnly
     )
 
+
+    $Global:MaximumFunctionCount = 16000
+
+    Write-Verbose  "Current value for Maxmimum Function Count: §($global:MaximumFunctionCount)"
+
+
     $InformationPreference = 'Continue'
     $currentPath = Join-Path -Path $PSScriptRoot -ChildPath '..\' -Resolve
     $manifest = Import-PowerShellDataFile "$currentPath/Dependencies/Manifest.psd1"
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -2795,7 +2795,8 @@ function Update-M365DSCDependencies
         }
         catch
         {
-            Write-Host "Could not update {$($dependency.ModuleName)}"
+            Write-Host "Could not update or import {$($dependency.ModuleName)}"
+            Write-Host "Error-Mesage: $($_.Exception.Message)"
         }
         $i++
     }
--- a/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
@@ -1157,6 +1157,8 @@ function Export-M365DSCConfiguration
 
     $Global:MaximumFunctionCount = 16000
 
+    Write-Verbose  "Current value for Maxmimum Function Count: §($global:MaximumFunctionCount)"
+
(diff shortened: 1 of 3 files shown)


Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -1660,7 +1662,11 @@ function New-M365DSCConnection
         [System.Boolean]
         $SkipModuleReload = $false
     )
+
     $Global:MaximumFunctionCount = 16000
+
+    Write-Verbose  "Current value for Maxmimum Function Count: §($global:MaximumFunctionCount)"
```
NikCharlebois, line 1668: Strange character. Can we please remove.  §

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -2755,6 +2761,12 @@ function Update-M365DSCDependencies
         $ValidateOnly
     )
 
+
+    $Global:MaximumFunctionCount = 16000
```
NikCharlebois, line 2765: Why are we doing this twice?

NikCharlebois (changes requested):

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
ricmestre: @NikCharlebois Hi, this doesn't actually solve the issue please read the whole thread #3454 on this topic.

andikrueger: as @ricmestre points out - it does not solve the issue. this was for testing purposes.

NikCharlebois (approved):

PR #3521, opened 2023-07-27
Avoid Unnecessary Instantiation of ApplicationInsights

Alik-Agarwala:
This pull request introduces an optimization to avoid unnecessary instantiation of the `ApplicationInsights` class in the telemetry service. In the original code, the `appInsights` variable was initialized with an empty configuration object, even though the actual initialization with the proper configuration occurred later inside the `initialize` function. This change ensures that the `appInsights` variable is only instantiated when the `initialize` function is called, thereby reducing unnecessary resource consumption.

This improvement does not introduce any functional changes, but it helps improve code organization and resource optimization. It avoids the creation of an empty `ApplicationInsights` instance when the telemetry service is first imported, and initialization now takes place only when necessary. The overall performance impact may be minimal, but it aligns with best practices for resource management.

### DIFF
--- a/generator/src/services/TelemetryService.ts
+++ b/generator/src/services/TelemetryService.ts
@@ -2,9 +2,7 @@ import { ApplicationInsights } from '@microsoft/applicationinsights-web';
 import { ReactPlugin } from '@microsoft/applicationinsights-react-js';
 
 let reactPlugin = new ReactPlugin();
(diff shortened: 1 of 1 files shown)


sebastienlevert (approved):
LGTM. Interesting find a I agree, minor but somewhat "useful" change here. Thanks!

Issue #3579, opened 2023-08-16
MSFT_EXOSafeAttachmentPolicy: ActionOnError Deprecated


coolriku:
### Description of the issue

When you try to apply a SafeAttachmentPolicy with Redirect set to true it fails because the "ActionOnError" parameter is [Deprecated](https://learn.microsoft.com/en-us/powershell/module/exchange/new-safeattachmentpolicy?view=exchange-ps#-actiononerror).

If you forcefully add this parameter to the *.mof file then you get the following error:
`{ A parameter cannot be found that matches parameter name 'ActionOnError'. } \ at Set-TargetResource, C:\Program Files\WindowsPowerShell\Modules\Microsoft365DSC\1.23.809.1\DscResources\MSFT_EXOSafeAttachmentPolicy\MSFT_EXOSafeAttachmentPolicy.psm1: line 305`
[See Reference:](https://github.com/microsoft/Microsoft365DSC/blob/eed60020e389bfa6c1e53b3a981eef7ce1fcc788/Modules/Microsoft365DSC/DSCResources/MSFT_EXOSafeAttachmentPolicy/MSFT_EXOSafeAttachmentPolicy.psm1#L305)

Solution:
Remove ActionOnError logic from module

### Microsoft 365 DSC Version

1.23.809.1

### Which workloads are affected

Exchange Online

PR #3580, opened 2023-08-16
MSFT_EXOSafeAttachmentPolicy: Remove ActionOnError Parameter

coolriku:
Fixes: #3579


coolriku:

andikrueger:
This would be a breaking change since we remove a parameter. Any chance to handle this in a non breaking change way by just deprecating the parameter and adding some verbose messages about the parameter no longer being supported?

coolriku:
I made a mess with the commits, I will make a new PR

PR #3584, opened 2023-08-17
MSFT_EXOSafeAttachmentPolicy: Deprecated ActionOnError Parameter

coolriku:
Fixes: #3579

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_EXOSafeAttachmentPolicy/MSFT_EXOSafeAttachmentPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_EXOSafeAttachmentPolicy/MSFT_EXOSafeAttachmentPolicy.psm1
@@ -254,27 +254,23 @@ function Set-TargetResource
         $StopProcessingPolicy = $false
         if ($Redirect -eq $true)
         {
-            $Message = 'Cannot proceed with processing of SafeAttachmentPolicy because Redirect is set to true '
-            if ($ActionOnError -eq $false)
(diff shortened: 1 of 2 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_EXOSafeAttachmentPolicy/MSFT_EXOSafeAttachmentPolicy.psm1
```diff
@@ -254,27 +254,23 @@ function Set-TargetResource
         $StopProcessingPolicy = $false
         if ($Redirect -eq $true)
         {
+            if ($ActionOnError -eq $true){
```
NikCharlebois, line 257: If the parameter is deprecated shouldn't we ignore it from all method and print a warning to the users letting them know that it is deprecated?

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSafeAttachmentPolicy/MSFT_EXOSafeAttachmentPolicy.psm1
coolriku: Yes, but if this parameter is set in older configuration than you get an error. That's why I remove this parameter from the payload.

NikCharlebois:
Perfect, can you please add an entry to the ChangeLog.md file describing your change and then we will merge. Thanks

PR #3726, opened 2023-09-28
Policy sets configuration

### LINKED ISSUE #3341: NewResourceName: MSFT_IntunePolicySets
### Description
I'm proposing a new resource to manage policy sets within intune
### Proposed properties
DisplayName
Description
Assignments
GuidedDeploymentTags
Items (could be separated into different policy types)

I had a go at creating the resource using the DRG for the cmdletNoun MgDeviceAppMgtPolicySet - it did create a configuration but it isn't functioning correctly.  I've not had much luck using the graph cmdlets and I think it'll need graph calls to configure all the values correctly.
I've made some small progress on amending the DRG created files but thought it made sense to put in the proposal and see if it is acceptable before doing too much more.

menswearUK:

A new configuration to support Intune Policy Sets.
I created this initially using the DRG but some of the functionality did't work, it appears a few of the graph cmdlets aren't behaving as expected.  I've amended it from the base DRG code include the Items in the policy set.
Where possible I've used the graph cmdlets and splatted the parameters but I've had to use Invoke-MgGraphRequest to set Items and Assignments where the cmdlets were not functioning.
I've tested within my test tenant and create/amend/delete functionality is working, as is the test-targetresource to detect settings drift.

Fixes #3341


NikCharlebois (changes requested):
Thanks for the contribution. Can you please also add an example in the Example folder for the new resource. This automatically drives the documentation.

NikCharlebois:
:shipit: LGTM

PR #3728, opened 2023-09-28
#3727 - TestHarness.psm1 -DscTestsPath  parameter

### LINKED ISSUE #3727: Suggestion to use unused parameter in TestHarness.psm1 to target tests to  a specific configuration
### Description of the issue

When working on updates we are generally looking at a single configuration but running the test harness by itself will run tests on all the configurations in the module.  This can take some time to complete

There is an existing parameter on the Invoke-TestHarness command called DscTestsPath which appears to be unused.
I'm not sure if this was originally intended for this purpose but it could be used to target a specific set of tests.

I propose an amendment to the function so, when this parameter is specified, it will only run the set of tests requested.  usage would be as below:
Invoke-TestHarness -DscTestsPath .\Tests\Unit\Microsoft365DSC\Microsoft365DSC.SomeConfiguration.Tests.ps1 -IgnoreCodeCoverage

### Microsoft 365 DSC Version

dev

other

### The DSC configuration

```powershell
n/a
```

```powershell
n/a - not a fault
```

```powershell
n/a, target node is ms cloud
```

menswearUK:

This is the amended version of TestHarness.psm1 I've been using to target a specific file during development.  I've found it quite useful to specify a single file using the -DscTestsPath  parameter.  Unless there's some other way of doing this that I've been missing all along (perfectly possible) I thought it could be helpful for others too.

fixes #3727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * DEPENDENCIES
   * Updated MSCloudLoginAssistant to version 1.0.121.
+* TestHarness.psm1
+  * Added code around DscTestsPath parameter to target a single test file during development
 
 # 1.23.927.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change log for Microsoft365DSC
 
+* TestHarness.psm1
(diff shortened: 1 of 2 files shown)


NikCharlebois:
:shipit: LGTM

PR #3737, opened 2023-10-03
Add exact match parameter to avoid invalid multiple results

### LINKED ISSUE #3736: Export may found multiple values  for CsGroup.DisplayName and CSGroup.GroupId, then creates invalid export
### Description of the issue

The query use for retrieving displayName and Id may return multiple values instead of the unique one.

In this case the displayname contains an array of string instead of the string.
For the GroupId, it contains an raay of on element and the element is the concatenation of the id but without quote.

```

        TeamsGroupPolicyAssignment "TeamsGroupPolicyAssignment-19"
        {
            ApplicationId         = $ConfigurationData.NonNodeData.ApplicationId;
            CertificateThumbprint = $ConfigurationData.NonNodeData.CertificateThumbprint;
            GroupDisplayName      = @("Management",Management External");
            GroupId               = @(4d458b94-af23-1234-a45c-a67c123a123123ab12c1-2ab3-1ab3-12a3-a1b2ab12345e);
            PolicyName            = "ter-asia-ondemand";
            PolicyType            = "TeamsComplianceRecordingPolicy";
            Priority              = 19;
            TenantId              = $OrganizationName;
        }
```

### Microsoft 365 DSC Version

1.23.927.1

Teams

### The DSC configuration

```powershell
Not fully sure, but it seems this append when a group has the same display name as another one plus something, like:
group1.displayName = "Management"

group2.displayName = "Management external"
```

_No response_

```powershell
Find-CsGroup -SearchQuery "1234567-1234-1234-1234-1234567890ab"| fl

RunspaceId  : 7654321-1234-1234-1234-1234567890ab
Id          : 1234567-1234-1234-1234-1234567890ab
DisplayName : Management

RunspaceId  : 7654321-1234-1234-1234-1234567890ab
Id          : 1234567-1234-1234-1234-1234567890df
DisplayName : Management External
```

Adding the parameter `-ExactMatchOnly $true` solve the pb:
```powershell
Find-CsGroup -SearchQuery "1234567-1234-1234-1234-1234567890ab" -ExactMatchOnly | fl

RunspaceId  : 7654321-1234-1234-1234-1234567890ab
Id          : 1234567-1234-1234-1234-1234567890ab
DisplayName : Management
 External
```

pkernevez:
Export may found multiple values for CsGroup.DisplayName and CSGroup.GroupId, then creates invalid export 

Fixes #3736


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -365,7 +365,7 @@ function Export-TargetResource
         $totalCount = $instances.Length
         foreach ($item in $instances)
         {
-            $Group = Find-CsGroup -SearchQuery $item.GroupId
+            $Group = Find-CsGroup -SearchQuery $item.GroupId -ExactMatchOnly $true
```
NikCharlebois, line 368: I am trying to understand what the outcome would be here. If we have two instances with the same name, we should print a warning letting the user know we won't be exporting them like we do for other resource (e.g., AADApplication) instead of trying to determine which ones to export or skip.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
pkernevez: I update the PR to support this case.

Now we are using the original groupId (unique) and and use the first display name (are they are all exactly equals).
I think we don't need any warning with this behaviour.

WDYT ?

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -365,15 +365,15 @@ function Export-TargetResource
         $totalCount = $instances.Length
         foreach ($item in $instances)
         {
-            $Group = Find-CsGroup -SearchQuery $item.GroupId
+            [array]$Group = Find-CsGroup -SearchQuery $item.GroupId -ExactMatchOnly $true
             if ($null -eq $totalCount)
             {
                 $totalCount = 1
             }
             Write-Host "    |---[$j/$totalCount] GroupPolicyAssignment {$($Group.DisplayName)}" -NoNewline
```
andikrueger, line 373: In the results hash table we use the first item of the array. We should do this here as well.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
```diff
@@ -365,15 +365,15 @@ function Export-TargetResource
...
             $results = @{
-                GroupDisplayName      = $Group.DisplayName
```
andikrueger, line 375: Is there any chance, that the array Group does not hold any elements? If yes, we should add a null check and check for the lenght.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * EXODistributionGroup
   * Fixes the export of group membership to use Identity.
+* TeamsGroupPolicyAssignment
+  * Fixes the export of CsGroup, when the display name of a group is included in another display name. Fix [#3736](https://github.com/microsoft/Microsoft365DSC/issues/3736)
```
andikrueger, line 8: A very minor thing. Could you please put "Fix [..." in a new line?

andikrueger (commented):
Just a few minor remarks

CHANGELOG.md
pkernevez: Yes

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
pkernevez: Of course. I've done it.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsGroupPolicyAssignment/MSFT_TeamsGroupPolicyAssignment.psm1
pkernevez: I wondered if this was the case, but in this case it means that we haven't managed to find the resource with its ID.
And what to put in place ?
Or we throw an error in this case, like:
```
            [array]$Group = Find-CsGroup -SearchQuery $item.GroupId -ExactMatchOnly $true
            if ($null -eq $totalCount)
            {
                $totalCount = 1
            } else if ($totalCount -Eq 0)
            {
                throw "GPA {$($item.GroupId)}: Unable to find a CsGroup whis that ID"
            }

```
What do you prefer

andikrueger: $totalCount is the number of instances and not related to the group count.

Throwing an exception will break the export. It would be better two write a warning/verbose message about an issue to resolve the group.

```powershell
if ($null -eq $group -or $group.length -eq 0)
{
# Write-Something
}
else if($group.lenght -ge 2){
#Write-something about several groups and taking the first one
}
else{
#current process
}
```

I haven't had a look into the return of Find-CsGroup and if there might be more than one return...

pkernevez: If there is more than one, it's not an issue. They all have the same display name as we are using `ExactMatchOnly $true`, any of them is ok, so let's take the first.
For me the only non nominal case is `-eq 0`, that is not supposed to happen (we are retrieving the display name of a group by its UID). 
The only possibility I see is: if the group has been deleted concurrently between the retrieval of the list (L354) and the call to `Find-CsGroup` inside the loop L368.
I don't know what is the strategy for this kind of error handling in the DSC365 project.
Do you have a pointer to documentation about error handling or a piece of code having the same behaviour?

andikrueger: I would be good to go with the easiest solution then and make sure to print out a warning to the log about the issue. This is something that should not fail the whole export.

pkernevez: I'm not fully clear how do you 'print out' a warning following your standard.
I've pushed a new commit with a try.

pkernevez: Are you expected something else ?

andikrueger: Yes, this looks good. Just one minor thing: Could you Write the message to the verbose stream and not the host. Host might break in some unattended instances.

pkernevez: Which one ?
I don't add any Write-Host on my code, I only changed the message of an existing one.
And I added a Write-Error

NikCharlebois (approved):

PR #3769, opened 2023-10-09
M365DSCReport: Also define property dataType, if present, as being primary key on CIM instances

ricmestre:

Some CIM instances, most notably *Assignments* in almost all Intune resources have the property *dataType* but it's not being properly set as primary key in *Get-M365DSCCimInstanceKey* which might cause issues when comparing reports/blueprints.

This PR solves this by adding *dataType* as candidate for being primary key for CIM instances, just like *odataType* is already today.

None.


andikrueger:
Thanks for suggesting this change! Could you please have a look at the conflict within the changelog?

ricmestre:
Done, if this gets in I'll update my other PR..

NikCharlebois (approved):

PR #3794, opened 2023-10-14
Add IntuneDeviceCleanupRule Resource

### LINKED ISSUE #3599: NewResourceName: IntuneDeviceCleanupRules
### Description
Adding support for INtune Clean-up rules described here:
https://learn.microsoft.com/en-us/graph/api/resources/intune-devices-manageddevicecleanupsettings?view=graph-rest-beta

### Proposed properties
enable / disable
deleteDevices after days

FabienTschanz:
Adds the IntuneDeviceCleanupRule as a new resource type.

- Fixes #3599


Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCleanupRule/MSFT_IntuneDeviceCleanupRule.psm1
```diff
@@ -0,0 +1,381 @@
...
+        -Parameters $PSBoundParameters
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    $nullResult = $PSBoundParameters
+    $nullResult.Ensure = 'Absent'
+
+    try
+    {
+        $url = "https://graph.microsoft.com/beta/deviceManagement/managedDeviceCleanupSettings"
```
NikCharlebois, line 72: Let's not hardcode URLs and instead use the $Global:MSCloudLoginProfile.MicrosoftGraph values to generalize this. Otherwise the code doesn't work for sovereign clouds. Also, we should use the cmdlets instead of direct graph calls wherever possible. In this case, we should call (Get-MgBetaDeviceManagement).ManagedDeviceCleanupSettings

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCleanupRule/MSFT_IntuneDeviceCleanupRule.psm1
```diff
@@ -0,0 +1,381 @@
...
+    #region Telemetry
+    $ResourceName = $MyInvocation.MyCommand.ModuleName -replace 'MSFT_', ''
+    $CommandName = $MyInvocation.MyCommand
+    $data = Format-M365DSCTelemetryParameters -ResourceName $ResourceName `
+        -CommandName $CommandName `
+        -Parameters $PSBoundParameters
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    $url = "https://graph.microsoft.com/beta/deviceManagement/managedDeviceCleanupSettings"
```
NikCharlebois, line 171: Same comment as above

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCleanupRule/MSFT_IntuneDeviceCleanupRule.psm1
```diff
@@ -0,0 +1,381 @@
...
+
+        [Parameter()]
+        [System.Int32]
+        [ValidateRange(30, 270)]
+        $DeviceInactivityBeforeRetirementInDays,
+
+        [Parameter()]
+        [System.String]
+        [ValidateSet('Absent', 'Present')]
+        $Ensure = 'Present',
```
NikCharlebois, line 196: What is the purpose of the Ensure property in this resource. The Set method doesn't seem to be looking for it. Since there will only ever be one instance of this, shouldn't we implement the IsSingleInstance pattern instead of the Ensure one? Have a look at how we are doing it for other tenant wide settings, for example TeamsOrgWideAppsSettings

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCleanupRule/MSFT_IntuneDeviceCleanupRule.psm1
FabienTschanz: As suggested I tried the `Get-MgBetaDeviceManagement` cmdlet, but that one does not return a value for the `DeviceInactivityBeforeRetirementInDays` property. It simply returns `$null`. So unfortunately, we can't use that one yet. Might raise an issue for that. 
image

FabienTschanz: I will open an issue a bit later, stay tight. In the meantime, I updated the code to use the `$Global:MSCloudLoginConnectionProfile.MicrosoftGraph.ResourceUrl` property. Tested and verified, works as intended.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCleanupRule/MSFT_IntuneDeviceCleanupRule.psm1
FabienTschanz: Updated as well, see above.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCleanupRule/MSFT_IntuneDeviceCleanupRule.psm1
FabienTschanz: Didn't know there exists a `IsSingleInstance` pattern. I had a look at the other resources and I hope I implemented it correctly. Please review it again, and if there is something missing, let me know.

NikCharlebois (approved):

FabienTschanz:
@NikCharlebois Would you like to fix the changelog? There is a conflict that requires our attention after the merge from the Dev branch.

NikCharlebois (approved):

Issue #3892, opened 2023-11-14
IntuneWindowsAutopilotDeploymentProfileAzureAD{,Hybrid}Joined, IntuneWindowsUpdateForBusinessRingUpdateProfileWindows10: Resources cannot set Assignments property


ricmestre:
### Description of the issue

*IntuneWindowsAutopilotDeploymentProfileAzureADJoined*, *IntuneWindowsAutopilotDeploymentProfileAzureADHybridJoined* and *IntuneWindowsUpdateForBusinessRingUpdateProfileWindows10* resources cannot apply the *Assignments* property, this seems to be due to an incorrect endpoint used in cmdlet *Update-DeviceConfigurationPolicyAssignment* (from M365DSCDRGUtil.psm1), according to https://learn.microsoft.com/en-us/graph/api/intune-enrollment-windowsautopilotdeploymentprofileassignment-update?view=graph-rest-beta it should be "deviceManagement/windowsAutopilotDeviceIdentities" instead of "deviceManagement/windowsAutopilotDeploymentProfiles", nevertheless the rest of the URI is also different and works differently since it must used POST for creating the assignments and PATCH to update them, whereas currently the cmdlet only uses POST.

This means that these resources always show up as not being in desired state since *Assignments* is not correct, I have a diff which I'll create a PR for that solves the problem with the *Id* property but that's a separate issue.

Minimal reproducer provided with the reference blueprint below, no errors will be displayed while configuration is deployed but Event Viewer shows the following below, but please bear in mind that the apps I use for deployment don't have problems with other resources and they already have scope 'DeviceManagementServiceConfig.ReadWrite.All' assigned which should be sufficient to perform this task so this is not a permission issue from my side.

```code
Error updating data:

{ Response status code does not indicate success: Forbidden (Forbidden). } \ at Update-DeviceConfigurationPolicyAssignment, C:\Program Files\WindowsPowerShell\Modules\Microsoft365DSC\1.23.1108.1\modules\M365DSCDRGUtil.psm1: line 1163
 \ at <ScriptBlock>, <No file>: line 1
```

@William-Francillette Could you please have a look?

### Microsoft 365 DSC Version

1.23.1108.1

### Which workloads are affected

other

### The DSC configuration

```powershell
Configuration ConfigureMicrosoft365
{
    Param(
        [Parameter()]
        [PSCredential]$Credential
    )

    Import-DscResource -ModuleName Microsoft365DSC

[...]

PR #3893, opened 2023-11-14
Remove Id from PSBoundParameters in Test-TargetResource

### LINKED ISSUE #3888: Intune: Some resources can be deployed but fail test on Test-TargetResource method
### Description of the issue

The resources below can be deployed to other tenants after exporting from another source tenant, nevertheless the *Test-TargetResource* fails so the resources keep updating every time when *Start-DSCConfiguration* is called.

This is happening since in *Test-TargetResource* the property *Id* is not being removed from being checked just like it was done on #3886, additionally *CertificateThumbprint* and *ManagedIdentity* should also be removed from that check or better yet in this case just call *Remove-M365DSCAuthenticationParameter* (defined in M365DSCUtil.psm1) to remove all authentication parameters.

```powershell
IntuneDeviceConfigurationDeliveryOptimizationPolicyWindows10
IntuneDeviceConfigurationHealthMonitoringConfigurationPolicyWindows10
IntuneDeviceConfigurationIdentityProtectionPolicyWindows10
IntuneWindowsAutopilotDeploymentProfileAzureADHybridJoined
IntuneWindowsAutopilotDeploymentProfileAzureADJoined
```

### Microsoft 365 DSC Version

1.23.1108.1

other

### The DSC configuration

```powershell
N/A
```

```powershell
N/A
```

```powershell
OsArchitecture       : 64-bit
WindowsVersion       : 2009
OsLanguage           : en-US

PSVersion                      5.1.22621.1778
BuildVersion                   10.0.22621.1778
```

ricmestre:
After deploy the resources below they always show up as not in desired state, this is because the *Id* parameter is not being removed from *PSBoundParameters* in Test-TargetResource, by removing it the issue is solved, except that the Autopilot ones also have a problem with *Assignments* property not being correctly set and therefore will not be completely fixed with this PR but I raised a separate issue for that.

IntuneDeviceConfigurationDeliveryOptimizationPolicyWindows10
IntuneDeviceConfigurationHealthMonitoringConfigurationPolicyWindows10
IntuneDeviceConfigurationIdentityProtectionPolicyWindows10
IntuneWindowsAutopilotDeploymentProfileAzureADHybridJoined
IntuneWindowsAutopilotDeploymentProfileAzureADJoined

EDIT: IntuneDeviceConfigurationEndpointProtectionPolicyWindows10 is affected by similar problem but not by not removing *Id* but *CertificateThumbprint*.
EDIT2: IntuneDeviceEnrollmentStatusPageWindows10 has both problems, it needs to have *Id* and all auth methods removed from Test, but also has problems with *Assignments* as reported on #3892 

- Fixes #3888


ricmestre:
@William-Francillette If you have time and this PR is accepted could you please check the DRG in order to have similar changes as I have here?

William-Francillette:
@NikCharlebois, @ykuijs @andikrueger 
Is there any use cases in checking Id in Test-TargetResource? If not, we may update Test-M365DSCParameterState from M365DSCUtil (around line 614) and force skipping Id as we do for the credentials parameters

andikrueger:
I just try to understand why to remove the ID from test-TargetResource. Usually we would compare the ID if it’s passed into the resource as a desired parameter. 

If it is set, we should compare it. 

E.g. someone wants to change the DisplayName (key) of an object. This would only work, if we monitor the ID as well.

Please correct me if I got the purpose wrong or the context of this change.

ricmestre:
@andikrueger If Id is not removed and is kept in the blueprint then when applying the policy to another tenant is not an issue but will complain that the resource is not in desired state because of course the Ids are different between tenants.

This is the same situation if Id is removed, now that it's not mandatory you can remove it therefore since the property is not there it will fallback to DisplayName and will just work, as long as the resource has been updated to do that, I know of some really old resources that if cannot be found by Id it won't fallback to DisplayName but that's another issue for another day.

NikCharlebois:
> @NikCharlebois, @ykuijs @andikrueger Is there any use cases in checking Id in Test-TargetResource? If not, we may update Test-M365DSCParameterState from M365DSCUtil (around line 614) and force skipping Id as we do for the credentials parameters

There are cases where even if the property is named Id (or Identity) it is represented by a friendly display name instead of a GUID. In these cases we need to keep ID as a param in the Test-TargetResource.

ykuijs (approved):
LGTM

PR #3920, opened 2023-11-21
AADRoleSetting export with Filter and sort by DisplayName

### LINKED ISSUE #3919: AADRoleSetting: "-Filter" parameter doesn't work when using Export-M365DSCConfiguration
### Description of the issue

Hello,

If you try to export only some roles using the Export-M365DSCConfiguration (AADRoleSetting + a Filter), the filter is not taken into account and all roles will be exported.

The issue seems coming from the MSFT_AADRoleSetting DSCResource and more particularly from the Export-TargetResource function. In fact, the parameter $Filter is not used in the below Cmdlet :
[array]$roles = Get-MgRoleManagementDirectoryRoleDefinition -ErrorAction Stop 

Maybe a solution would be to simply replace the above command with (there might be better solutions):
[array]$roles = Get-MgRoleManagementDirectoryRoleDefinition -Filter $Filter -ErrorAction Stop

This is my first issue so don't hesitate if you have any questions, I'm not familiar with the process.. :)

### Microsoft 365 DSC Version

1.23.1115.1

Azure Active Directory

### The DSC configuration

```powershell
Example:
$Component = "AADRoleSetting"
$Filters   = @{ AADRoleSetting = "DisplayName eq 'Global Administrator'" }
Export-M365DSCConfiguration  -Components $Component `
                             -TenantId              $TenantId `
                             -ApplicationId         $ApplicationId `
                             -CertificateThumbprint $CertificateThumbprint `
                             -Filters               $Filters              
--> This command should return only the Global Administrator role but it will export all roles (as the filter is not used...)
```

_No response_

_No response_

MKlingner:
AADRoleSetting
  * Export sorted by DisplayName for better comparison
  * Enable Filter property to be used on export

    - Fixes #3919


Modules/Microsoft365DSC/DSCResources/MSFT_AADRoleSetting/MSFT_AADRoleSetting.psm1
```diff
@@ -1379,7 +1380,7 @@ function Export-TargetResource
     try
     {
         $Script:ExportMode = $true
-        [array] $Script:exportedInstances = Get-MgBetaRoleManagementDirectoryRoleDefinition -ErrorAction Stop
+        [array] $Script:exportedInstances = Get-MgBetaRoleManagementDirectoryRoleDefinition -Filter $Filter -Sort DisplayName -ErrorAction Stop
```
andikrueger, line 1383: Not sure what will happen, if filter is empty? Does it make sense to add the Filter only when it is set?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change log for Microsoft365DSC
 
+* AADRoleSetting
+  * Export sorted by DisplayName for better comparison
+  * Enable Filter property to be used on export
+    FIXES [#3919](https://github.com/microsoft/Microsoft365DSC/issues/3919)
```
andikrueger, line 6: Please add a new line after the link.

andikrueger (commented):
Thanks for this contribution. Just two minor comments.

Modules/Microsoft365DSC/DSCResources/MSFT_AADRoleSetting/MSFT_AADRoleSetting.psm1
MKlingner: i've done a recheck and Get-MgBetaRoleManagementDirectoryRoleDefinition with empty Filter returns the same results as without

Issue #4011, opened 2023-12-07
AADConditionalAccessPolicy: Add possibility to specify Include-/ExcludeApplications by name instead of GUID


salbeck-sit:
### Description of the issue

Today we specify IncludeApplications and ExcludeApplications in Conditional Access policies by their GUID. The exception is built-in names such as 'All', 'Office365' and likely others that I haven't had use for so far.
Groups and Users are specified by their name - so why not Apps ?
The update I'm suggesting will allow applications to be specified by their DisplayName in the same way that the Entra portal allows you to do while at the same time keeping the door open to GUIDs being used in order to keep the resource backwards-compatible.
It may be an issue to keep on top of additional built-in names added. I haven't been able to find a list other than what I can infer from https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-conditional-access-cloud-apps
It doesn't make any real difference if 'built-in names' are updated in the resource or not as it is somewhat lenient in regards to 'translating' a name to a GUID. If an SPN doesn't exist, the name is used as-is. This may cause runtime-errors, ie if a name is missspelled, but errors should be easy enough to catch. The resource can also be updated to provide logging of such errors

I have a working copy of the resource and also expanded unit-tests. It'd be great if someone can provide a longer list of the above mentioned built-in names - or provide it as an add-on PR.

### Microsoft 365 DSC Version

DEV

### Which workloads are affected

Azure Active Directory

### Environment Information + PowerShell Version

```powershell
Built and tested using Windows Powershell 5.1 and Pester 5.4.0
```

salbeck-sit:
PS I've also noticed a bug (or rather, VSCode did) in AADCOnditionalAccessPolicy wrt TermsOfUse due to a missspelled variable-name. I haven't included it as a separate issue but it is included in the PR I have ready

andikrueger:
I guess, this was done to mitigate the risk of having applications with the same display name. Does you code handle this as well? User Principal Names for groups and users are unique.

salbeck-sit:
@ricmestre my idea about updating 'built-in names' is that such names are the ones defined in the GUI that doesn't refer to a GUID. Ie the names are logical like 'Office365' or 'All' because they don't refer to a 'real' app but to a group of apps.
That means that these names cannot be represented by a GUID. In regards to the above, that means that on export, these names cannot be coverted to GUIDs  and in configs, they also need to be specified by name only.

[...]

PR #4051, opened 2023-12-15
AADConditionalAccessPolicy - implement #4011

salbeck-sit:
Update AADConditionalAccessPolicy parameters IncludeApps and ExcludeApps to accept ServicePrincipal displaynames in addition to application-id if names are unique. For non-unique names, the application-id must be used instead.
For exports, application-ids are converted to displaynames where unique.
This may cause 'false drift' if or when a previous export contains application-ids and a later export contains names for the same applications

Implements #4011

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1
@@ -264,415 +264,414 @@ function Get-TargetResource
         }
     }
 
-    if ($null -eq $Policy)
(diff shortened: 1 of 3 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADConditionalAccessPolicy/MSFT_AADConditionalAccessPolicy.psm1
```diff
@@ -495,7 +495,7 @@ function Get-TargetResource
                 }
             }
 
-            Write-Verbose -Message 'Get-TargetResource: Processing ExcludeLocations'
+            Write-Verbose 'Get-TargetResource: Processing ExcludeLocations'
```
NikCharlebois, line 498: Let's not remove the -Message parameter for verbose. This is the standard across all resources.

NikCharlebois (changes requested):

PR #4072, opened 2023-12-20
Set ErrorAction to Stop on integration workflows

ricmestre:

Set ErrorAction to Stop on integration workflows on both {Start,Test}-DSCConfiguration cmdlets.

### DIFF
--- a/.github/workflows/Global - Integration - AAD.yml
+++ b/.github/workflows/Global - Integration - AAD.yml
@@ -63,12 +63,14 @@ jobs:
 
           try
           {
-            $Result = Test-DSCConfiguration -Detailed -Verbose
+            $Result = Test-DSCConfiguration -Detailed -Verbose -ErrorAction Stop
           }
           catch
           {
             throw $_
           }
+
+          Write-Host ""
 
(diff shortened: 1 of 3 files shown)


.github/workflows/Global - Integration - AAD.yml
```diff
@@ -63,13 +63,15 @@ jobs:
...
           {
-            $Result = Test-DSCConfiguration -Detailed -Verbose
+            $Result = Test-DSCConfiguration -Detailed -Verbose -ErrorAction Stop
           }
           catch
           {
             throw $_
           }
 
+          Write-Host ""
```
NikCharlebois, line 73: Why the empty Write-Host?

NikCharlebois (changes requested):

.github/workflows/Global - Integration - AAD.yml
ricmestre: To leave an empty line between whatever comes out of Test-DSCConfiguration and the results below, at least for me it's easier to look at the results that way otherwise they are stuck together.

Issue #4091, opened 2023-12-22
Compilation errors occurred while processing configuration 'M365TenantConfig'.


DscEngineer:
### Description of the issue

I've been working on DSC for past few months and already deployed Exchange, Teams, SPO but in Intune I am getting this error while trying to create MOF file. Please help me fix this as I have a lot of data in this policy and I can not create it Manually.

```
IntuneDeviceConfigurationEndpointProtectionPolicyWindows10\M365TenantConfig.ps1:229 char:17
+                 MSFT_MicrosoftGraphwindowsFirewallRule{
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Write-Error], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnsupportedValueForProperty,Microsoft365DSC\MSFT_MicrosoftGraphWindowsFirewallRule

Microsoft365DSC\MSFT_MicrosoftGraphwindowsFirewallRule : At least one of the values 'remoteAccess,wireless,lan' is not supported or valid for property 'InterfaceTypes' on class 'MSFT_MicrosoftGraphWindowsFirewallRule'. Please specify only supported values: notConfigured, remoteAccess, wireless, lan.

Compilation errors occurred while processing configuration 'M365TenantConfig'. Please review the errors reported in error stream and modify your configuration code appropriately.
At C:\Windows\system32\WindowsPowerShell\v1.0\Modules\PSDesiredStateConfiguration\PSDesiredStateConfiguration.psm1:3917 char:5
```

### Microsoft 365 DSC Version

1.23.1213.1

### Which workloads are affected

other

### The DSC configuration

```powershell
IntuneDeviceConfigurationEndpointProtectionPolicyWindows10
```

DscEngineer:
am able to fix this but it still did not resolve the issue I was expecting it will. I was trying to configure Endpoint security | Firewall Configuration settings where I have 100 of rules configured and I cannot find any component for DSC on the portal. Can someone please share the same as it is very important setting and it is impossible to do manual configuration of 100+ firewall rules.

ykuijs:
Hi @DscEngineer, could you please share the configuration you are using? I think there might be an issue in there.

William-Francillette:
@DscEngineer, @ykuijs 
I had a look and there is a disparity between the schema and actual API which is the cause of the issue
[...]

PR #4122, opened 2024-01-04
SCDLPComplianceRule: Fix data type

### LINKED ISSUE #3463: SCDLPComplianceRule - AccessScope - Cannot convert value "System.Collections.Generic.List1[System.String]"

**SCDLPComplianceRule - AccessScope - Cannot convert value**
I am deploying one  DLPComplianceRule. When specifying `AccessScope` it always fails with the same error. When removing this key, the configuration succeeds.

##[error]Cannot process argument transformation on parameter 'AccessScope'. Cannot convert value 
"System.Collections.Generic.List`1[System.String]" to type 
"System.Nullable`1[Microsoft.Office.CompliancePolicy.Tasks.AccessScope]". Error: "Unable to match the identifier name 
System.Collections.Generic.List`1[System.String] to a valid enumerator name. Specify one of the following enumerator 
names and try again: InOrganization, NotInOrganization, None"
```powershell
Configuration M365Configuration {
    param
    (
        [System.String]$ApplicationId,

        [System.String]$TenantId,
        
        [System.String]$CertificateThumbprint
    )

    Import-DscResource -ModuleName Microsoft365DSC

    node localhost
    {
        SCDLPCompliancePolicy 'ConfigureCompliancePolicy'
        {
            Name                  = "General Data Protection Regulation (GDPR)"
            Comment               = "Test Policy"
            Priority              = 0
            ExchangeLocation      = "All"
            Ensure                = "Present"
            ApplicationId         = $ApplicationId;
            CertificateThumbprint = $CertificateThumbprint;
            TenantId              = $TenantId;
        }
        SCDLPComplianceRule 'ConfigureDLPComplianceRule'
        {
            Name                                = 'Low volume EU Sensitive content found'
            Policy                              = 'General Data Protection Regulation (GDPR)'
            AccessScope                         = 'InOrganization'
            BlockAccess                         = $True
            BlockAccessScope                    = 'All'
            ContentContainsSensitiveInformation = MSFT_SCDLPContainsSensitiveInformation
            {
                SensitiveInformation = @(
                    MSFT_SCDLPSensitiveInformation
                    {
                        name           = 'EU Debit Card Number'
                        id             = '0e9b3178-9678-47dd-a509-37222ca96b42'
                        maxconfidence  = '100'
                        minconfidence  = '75'
                        classifiertype = 'Content'
                        mincount       = '1'
                        maxcount       = '9'
                    }
                )
            }
            Disabled                            = $False
            DocumentIsPasswordProtected         = $False
            DocumentIsUnsupported               = $False
            ExceptIfDocumentIsPasswordProtected = $False
            ExceptIfDocumentIsUnsupported       = $False
            ExceptIfHasSenderOverride           = $False
            ExceptIfProcessingLimitExceeded     = $False
            GenerateIncidentReport              = @('SiteAdmin')
            HasSenderOverride                   = $False
            IncidentReportContent               = @('DocumentLastModifier', 'Detections', 'Severity', 'DetectionDetails', 'OriginalContent')
            NotifyUser                          = @('LastModifier')
            ProcessingLimitExceeded             = $False
            RemoveRMSTemplate                   = $False
            ReportSeverityLevel                 = 'Low'
            StopPolicyProcessing                = $False
            Ensure                              = 'Present'
            ApplicationId                       = $ApplicationId;
            CertificateThumbprint               = $CertificateThumbprint;
            TenantId                            = $TenantId;
        }
    }
}
```

Windows Server 2022

1.23.628.1

malauter:
Fix data type of AccessScope

- Fixes #3463


ricmestre:
Isn't this considered a breaking change?

malauter:
> Isn't this considered a breaking change?

Usually changing a data type is a breaking change, but in this case the AccessScope attribute never worked before. Not sure if we should still consider it as a breaking change in this case.

ricmestre:
It's the same case I found while fixing *IntuneDeviceConfigurationPolicyMacOs* and it's being discussed again on similar issue in #4091

ykuijs:
It isn't 100% the same case as you are mentioning @ricmestre. There you can specify multiple values for a parameter, but the question is how to input them: As a single comma separated string or as a collection.

In this case the resource has a collection configured, but the New-DlpComplianceRule cmdlet only accepts a single value. That does not match and will never work.

Formally this change is indeed a breaking change, but it fixes a situation that never worked in the first place. So a bit torn in what we should be doing here.

@NikCharlebois What is you opinion here? Should we merge and fix the issue now or wait until the next breaking change release?

ricmestre:
@ykuijs In *IntuneDeviceConfigurationPolicyMacOs* the resource also never worked unless *UpdateDelayPolicy* wasn't used as parameter, that's because Graph API expects a String and the resource was setup with that parameter as String[] so if you used that param the policy was never created because the API always returned a bogus error message. The obvious solution would be to change the data type as it's being done here, but I changed it differently in order to not be considered a breaking change, as mentioned elsewhere it's still not working correctly but I have a PR opened to fix it properly.

I'm not saying I'm against this PR, in fact I'd prefer to have this in without right now without being considered a breaking change.

NikCharlebois (approved):

NikCharlebois:
We need to keep in mind that most configuration files were first generated via an export operation, which would only return a string value at the moment. While I agree that this is considered a breaking change based on our principles, trying to set the value never actually worked, and most configurations will continue to compile without any errors since the majority of the values are already just a single string. I am ok with merging this change. 

Even for #4091, it's something I would be ok with including. @ykuijs and I are having a parallel chat about allowing breaking changes to be merged before major releases when they fix something that is broken anyway:

"We should be careful with this. In #4122, the monitoring scenario works fine, it's the Set that is broken. 70% of customers use M365DSC for monitoring only. So fixing Set with a risk (even if very tiny) to break Test is sometimes not worth it in my opinion
In the present case, most values are single string anyway, so changing from String[] to String will also have a very minimal impact"

PR #4123, opened 2024-01-04
EXORecipientPermission Initial Release

raandree:

EXORecipientPermission Initial Release.

None.

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_EXORecipientPermission/MSFT_EXORecipientPermission.schema.mof
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_EXORecipientPermission/MSFT_EXORecipientPermission.schema.mof
@@ -1,8 +1,8 @@
 [ClassVersion("1.0.0.0"), FriendlyName("EXORecipientPermission")]
 class MSFT_EXORecipientPermission : OMI_BaseResource
 {
-    [Write, Description("The mailbox the permission should be given on.")] String Identity;
-    [Write, Description("The account to give the permission to.")] String Trustee;
+    [Key, Description("The mailbox the permission should be given on.")] String Identity;
+    [Key, Description("The account to give the permission to.")] String Trustee;
     [Write, Description("The access rights granted to the account. Only 'SendAs' is supported.")] String AccessRights;
 
     [Write, Description("Present ensures the group exists, absent ensures it is removed"), ValueMap{"Present","Absent"}, Values{"Present","Absent"}] string Ensure;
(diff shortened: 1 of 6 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_EXORecipientPermission/MSFT_EXORecipientPermission.schema.mof
```diff
@@ -0,0 +1,16 @@
+[ClassVersion("1.0.0.0"), FriendlyName("EXORecipientPermission")]
+class MSFT_EXORecipientPermission : OMI_BaseResource
+{
+    [Key, Description("The mailbox the permission should be given on.")] String Identity;
+    [Key, Description("The account to give the permission to.")] String Trustee;
+    [Write, Description("The access rights granted to the account. Only 'SendAs' is supported.")] String AccessRights;
```
ykuijs, line 6: If only SendAs is supported, shouldn't we hardcode that in the resource? Doesn't make sense to request input from the user, but require a specific value.

Modules/Microsoft365DSC/DSCResources/MSFT_EXORecipientPermission/MSFT_EXORecipientPermission.schema.mof
```diff
@@ -0,0 +1,16 @@
+[ClassVersion("1.0.0.0"), FriendlyName("EXORecipientPermission")]
+class MSFT_EXORecipientPermission : OMI_BaseResource
+{
+    [Key, Description("The mailbox the permission should be given on.")] String Identity;
+    [Key, Description("The account to give the permission to.")] String Trustee;
```
ykuijs, line 5: Would it be an idea to make this parameter a collection? That way we can add one resource per Identity to a config instead of one resource per Trustee per Identity. That would dramatically reduce the number of resources required in the config to configure a larger number of Trustees.

Modules/Microsoft365DSC/DSCResources/MSFT_EXORecipientPermission/readme.md
```diff
@@ -0,0 +1,5 @@
+# EXORecipientPermission
+
+## Description
+
+This resource allows users to retrieve Office 365 Recipient Permission.
```
ykuijs, line 5: Permission should be plural: Permissions

ykuijs (changes requested):
Few comments and I am missing an Changelog entry and unit tests.

NikCharlebois:
@raandree can you take a look at @ykuijs' review and update? Thanks

Modules/Microsoft365DSC/DSCResources/MSFT_EXORecipientPermission/MSFT_EXORecipientPermission.schema.mof
raandree: Right, done that.

PR #4149, opened 2024-01-15
EXOTransportRule: Change data type of Priority from String to Int

### LINKED ISSUE #4136: [EXOTransportRule] Priority parameter is String instead of Int
### Description of the issue

The Priority parameter in the EXOTransportRule resource has been defined in the schema as a String. When reviewing the New-TransportRule cmdlet, there the Priority property is defined as an Int32.

It doesn't sound logical that these are of different types and it would make sense to align these. That would be a breaking change.

Resource schema: https://github.com/microsoft/Microsoft365DSC/blob/e4f6c3eb6c5b41f54a8fdcbac5bdbba6158a281f/Modules/Microsoft365DSC/DSCResources/MSFT_EXOTransportRule/MSFT_EXOTransportRule.schema.mof#L131
New-TransportRule cmdlet: https://learn.microsoft.com/en-us/powershell/module/exchange/new-transportrule?view=exchange-ps

### Microsoft 365 DSC Version

Dev

Exchange Online

### The DSC configuration

```powershell
N/A
```

```powershell
N/A
```

```powershell
N/A
```

malauter:
This is a breaking change!

Change data type of Priority from String to Int

- Fixes #4136


NikCharlebois:
Is the fact that it is incorrectly set as a string causing any issues?

ykuijs:
It isn't breaking anything right away, but it is incorrect. We are now relying on PowerShell converting the value to an Int. Users could also enter a different value than a number, which will break the deployment. By changing this value to an Int, the resource matches the actual situation and we are preventing potential issues. It is however a breaking change, so should be released with the next breaking change release.

Issue #3935, opened 2023-11-23
AADGroup -Filter -ConsistencyLevel -CountVariable 


ifinch:
### Description of the issue

While testing, we found that we cannot search for the onPremisesSyncEnabled attribute like below

$filter =  @{AADGroup = "OnPremisesSyncEnabled eq null" }
export-M365DSCConfiguration -Credential $creds -Components $componentName -FileName $fileName -path  $path -ConfigurationName $configName -Filters $filter

•	onPremisesSyncEnabled eq Null
•	onPremisesSyncEnabled eq ‘Null’
•	onPremisesSyncEnabled ne true
•	onPremisesSyncEnabled eq $null

The only way we have been able to get the filter out through powershell is through the following run

get-mggroup -filter "OnPremisesSyncEnabled ne true" -CountVariable CountVar -ConsistencyLevel eventual

### Microsoft 365 DSC Version

1.23.1115.1

### Which workloads are affected

Azure Active Directory

jagatjb:
Be it AADGroup or O365Group or any other components, If filter option can be used for all the conditions, then it will be a gamechanger for us.

bartvermeersch:
This is indeed only a quick patch I'm afraid. We would need startswith to collect all groups starting with a prefix.
If you don't want to use consistentcyLevel/count for every filter, would it be a solution to implement two separate parameters -Filters and -AdvancedFilters ?

bartvermeersch:
A similar issue for AADUser: [2430](https://github.com/microsoft/Microsoft365DSC/issues/2430)

hao-public:
We need this as well. We have 25K+ groups from our on-premises AD and need to filter them out. We only want to see the Cloud enabled groups! It takes 13+ hours to retrieve the groups for us right now - this is not scalable.

FabienTschanz:
Filtering based on onPremisesSyncEnabled is supported. ConsistencyLevel and CountVariable are set when exporting AADGroup resources. Filter support was greatly enhanced recently.

PR #4164, opened 2024-01-15
Update MSFT_AADGroup.psm1

ifinch:
line 1050
changed from if ($Filter -like "*endsWith*") to if ($Filter -like "*endsWith*" -or $Filter -like "*onPremisesSyncEnabled eq null*" -or $Filter -like "*onPremisesSyncEnabled ne true*")

it resolves the issue in #3935


ricmestre:
I added that condition to the $Filter to itch a scratch at the time but in order to actually solve this (also for other resources that support filtering) someone needs to go through the list in https://learn.microsoft.com/en-us/graph/aad-advanced-queries?tabs=http#device-properties and implement the condition correctly.

Looking at that table it shows that for your use case you only need to define *ConsistencyLevel* and *$count* if comparing *onPremisesSyncEnabled* to null, it's not actually required if comparing to true or false, I've just tested it in my tenant and can verify this so your patch is half incorrect.

The best way would be to have the MSGraph SDK decide which properties need *ConsistencyLevel* and *$count* and add them automatically, this has been requested several years ago but it's still not in place, see https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/513.

ifinch:
changed if ($Filter -like "endsWith" -or $Filter -like "onPremisesSyncEnabled eq null" -or $Filter -like "onPremisesSyncEnabled ne 
true") to be if ($Filter -like "endsWith" -or $Filter -like "onPremisesSyncEnabled eq null")  removing "onPremisesSyncEnabled ne true"

ifinch:
added the block to check allowed advanced Group properties from the table https://learn.microsoft.com/en-us/graph/aad-advanced-queries?tabs=http#group-properties that support eq null.

start line 1050, ends 1078

# Define the list of attributes
        $attributesToCheck = @(
            "description",
            "displayName",
            "hasMembersWithLicenseErrors",
            "mail",
            "mailNickname",
            "onPremisesSecurityIdentifier",
            "onPremisesSyncEnabled",
            "preferredLanguage"
        )

        # Initialize a flag to indicate whether any attribute matches the condition
        $matchConditionFound = $false

        # Check each attribute in the list
        foreach ($attribute in $attributesToCheck) {
            if ($Filter -like "*$attribute eq null*") {
                $matchConditionFound = $true
                break
            }
        }

        # If any attribute matches, add parameters to $ExportParameters
        if ($matchConditionFound) {
            $ExportParameters.Add('CountVariable', 'count')
            $ExportParameters.Add('ConsistencyLevel', 'eventual')
        }

Modules/Microsoft365DSC/DSCResources/MSFT_AADGroup/MSFT_AADGroup.psm1
```diff
@@ -1047,10 +1047,36 @@ function Export-TargetResource
...
+        # Check each attribute in the list
+        foreach ($attribute in $attributesToCheck) {
+            if ($Filter -like "*$attribute eq null*") {
+                $matchConditionFound = $true
+                break
+            }
+        }
+
+        # If any attribute matches, add parameters to $ExportParameters
+        if ($matchConditionFound) {
```
ricmestre, line 1075: Now you're missing *endsWith* here, this is not an attribute but a function that you can call on the query which also needs to be checked and if you remove it then it stops working.

ifinch: changed the condition to look as below to take into account endsWith
so it is now

if ($matchConditionFound -or $Filter -like "*endsWith*"))

ricmestre:
@NikCharlebois I'm not 100% sure but from the tests I've seen always including *ConsistencyLevel* and *CountVariable* in the cmdlets while filtering doesn't cause any issues, I might be wrong here though but if that's the case then adding them always if *$Filter* is not null would be better then have to chase whatever is mentioned in [0]. As an example they even omit the function *endsWith* that I added so I don't know exactly to which properties it applies so I had to apply the big hammer to the condition which now I asked @ifinch to change.

Any insights?

[0] https://learn.microsoft.com/en-us/graph/aad-advanced-queries?tabs=http#device-properties

NikCharlebois:
Can we please add an entry to the changelog? Thanks

NikCharlebois:
> _ConsistencyLevel_

Not sure this is really something we want to push at scale since it can change the current behavior in certain scenarios:
https://devblogs.microsoft.com/microsoft365dev/build-advanced-queries-with-count-filter-search-and-orderby/#what-is-eventual-consistency

NikCharlebois:
Can you update your branch? Not sure why unit tests are failing in your PR. Thanks

ifinch:
Done

ifinch:
It is pending anything on my side?

PR #4195, opened 2024-01-19
Only do one request instead of n+1 when exporting TeamsCallQueue

### LINKED ISSUE #4192: When exporting CsCallQueue (it's also right for a lot of other resources), we are doing n+1 request instaead of 1
### Description of the issue

The export have the following pattern:
1. Create a list of what we wan to export
2. For each of them read the detail of the row
3. export the row

For CsCalQueue (and most of the time for other resources)
part 1, does one request to retrieve the list of n resources
part2, does n request to retrieve the detail

But both part are using same API, meaning that after part1 we already have the data. part 2 is useless.

In my case n=150, it means that the export could speed by ~100 (part1 cost more than other requests ).

### Microsoft 365 DSC Version

1.23.110.1

Teams

### The DSC configuration

_No response_

_No response_

_No response_

pkernevez:
Only do one request instead of n+1 when exporting TeamsCallQueue
This is not specific at this resource, but this PR only fix it for it.

Fixes and Close #4192


NikCharlebois:
Sorry, I am having a hard time understanding how this is different than what we had previsouly?

ricmestre:
@nikcharlebois Currently this resource is calling *Get-CsCallQueue*, during *Export-TargetResource*, which gets all call queues, then it iterates per each one of them and calls *Get-CsCallQueue* again inside *Get-TargetResource* although this time with a filter by name to figure out its settings.

It seems what @pkernevez is trying to do here is to keep the initial call to *Get-CsCallQueue* for exporting all queues but avoiding the subsequent calls inside *Get-TargetResource*, in fact bypassing getting inside the Get completely, and just return the settings through the cmdlet *Convert-Resource* he introduced.

Didn't test it tough, don't have any tenants with examples right now.

Modules/Microsoft365DSC/DSCResources/MSFT_TeamsCallQueue/MSFT_TeamsCallQueue.psm1
```diff
@@ -907,7 +940,8 @@ function Export-TargetResource
                 TenantId              = $TenantId
                 CertificateThumbprint = $CertificateThumbprint
             }
-            $Results = Get-TargetResource @Params
+            # $Results = Get-TargetResource @Params
```
andikrueger, line 943: Could you please remove this line.

andikrueger (changes requested):
This looks like a good runtime improvement for exporting all instances of this resource. The redundant calls to Get-TargetResource are mitigated this way. It even would allow the implementation of $Filter for this resource, if I see it correctly.

NikCharlebois:
Looks like the caching pattern that is being used across most resources, where we cache all instances in the $Script:exportedInstances variable is not fully implemented here. I would rather make sure we keep the patterns aligned across resources than creating yet a new one. Have a look at how we are doing it for the AADUser resource for example:

https://github.com/microsoft/Microsoft365DSC/blob/Dev/Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1#L184

pkernevez:
I didn't knew this existing pattern ( I didn't found it before when reading the code).
My target (reducing the number of calls) could be archived by following it, so I'm going to push a new version following it.

pkernevez:
I revert previous changes and use the export cache pattern, it was partially in place : the cache existed but was not used.

pkernevez:
Are the changes ok for you  ?

andikrueger (approved):
LGTM

Issue #4082, opened 2023-12-21
IntuneDeviceEnrollmentPlatformRestriction: Assignments not working


ricmestre:
### Description of the issue

I created a policy manually which I then exported, config found below, then I removed it and tried to recreate but the assignment cannot be applied since it fails with *Bad request* on *Update-DeviceConfigurationPolicyAssignment* (failure not visible on CLI, only in Event Viewer).

The policy still gets created but then deploying the config again will result in error (hard failure in CLI, deployment stops) because then it tries to update it a few lines below in the code where it has another call to *Update-DeviceConfigurationPolicyAssignment*, which would have the same problem as above, but in this case it even fails for another reason which is calling it with *$Identity* which can be from another tenant (or from an already deleted policy in this case) so that chunk needs to changed to *$currentCategory.Identity* instead.

### Microsoft 365 DSC Version

1.23.1220.1

### Which workloads are affected

other

### The DSC configuration

```powershell
IntuneDeviceEnrollmentPlatformRestriction "IntuneDeviceEnrollmentPlatformRestriction-IntuneDeviceEnrollmentPlatformRestriction_1"
        {
            Assignments                       = @(
                MSFT_DeviceManagementConfigurationPolicyAssignments{
                    deviceAndAppManagementAssignmentFilterType = 'none'
                    dataType = '#microsoft.graph.groupAssignmentTarget'
                    deviceAndAppManagementAssignmentFilterId = '00000000-0000-0000-0000-000000000000'
                    groupId = 'b0b8fd3f-af2a-453b-be57-80182d599f02'
                }
            );
            Credential                        = $CredsCredential;
            DeviceEnrollmentConfigurationType = "singlePlatformRestriction";
            DisplayName                       = "IntuneDeviceEnrollmentPlatformRestriction_1";
            Ensure                            = "Present";
            Identity                          = "7857eee0-7b91-48e4-a72c-f66f36c4a483_SinglePlatformRestriction";
            WindowsRestriction                = MSFT_DeviceEnrollmentPlatformRestriction{
                platformBlocked = $False
                personalDeviceEnrollmentBlocked = $False
            };
        }
```

### Verbose logs showing the problem

```powershell
Error updating data:

[...]

PR #4247, opened 2024-01-25
Customize Intune policy assignments

FabienTschanz:
Adds a new parameter to the `Update-DeviceConfigurationPolicyAssignment`-Cmdlet to enable assignments for all Intune policies, despite different root key names.  
Normally, the root key is `assignments`. In some cases, that name is changed to e.g. `enrollmentConfigurationAssignments`. The current implementation does not work with those different root key names. With the change, you can add the `-RootIdentifier` parameter with the root key name so that the POST and assign works for those resources as well. 

None.


ricmestre:
@nikcharlebois @andikrueger Hi, can this please be merged?

It'd still need a couple of fixes afterwards but the issue with assignments on *IntuneDeviceEnrollmentPlatformRestriction* that I reported on #4082 could be fixed with this in since the root identifier is hard coded right now to *assignments* but for this resource instead it needs to be *enrollmentConfigurationAssignments* as mentioned in https://learn.microsoft.com/en-us/graph/api/intune-shared-deviceenrollmentconfiguration-assign?view=graph-rest-beta

FabienTschanz:
@NikCharlebois @andikrueger Can you have a look and merge it please? There is the pull request #4362  waiting that has this one as a requirement. And it's already open for a quite a long time too.

ykuijs (approved):
LGTM

Issue #4261, opened 2024-01-26
TeamsEmergencyCallRoutingPolicy: Cannot remove


ricmestre:
### Description of the issue

Setting Ensure to Absent outputs that the resource has been removed but it actually doesn't, this is due to using parameter *Confirm:$false* in cmdlet *Remove-CsTeamsEmergencyCallRoutingPolicy* which then just returns without doing anything.

Fix here is just to remove the parameter from the cmdlet since it's not interactive and doesn't need it.

### Microsoft 365 DSC Version

1.23.124.1

### Which workloads are affected

Teams

### The DSC configuration

```powershell
TeamsEmergencyCallRoutingPolicy "TeamsEmergencyCallRoutingPolicy-TeamsEmergencyCallRoutingPolicy_1"
        {
            AllowEnhancedEmergencyServices = $False;
            ApplicationId                  = $TeamsApplicationId;
            CertificateThumbprint          = $TeamsCertThumbprint;
            Ensure                         = "Absent";
            Identity                       = "TeamsEmergencyCallRoutingPolicy_1";
            TenantId                       = $OrganizationName;
        }
```

### Verbose logs showing the problem

```powershell
N/A
```

### Environment Information + PowerShell Version

```powershell
OsName               : Microsoft Windows 11 Enterprise
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250
OsLanguage           : en-US
OsMuiLanguages       : {en-US, en-GB}

Name                           Value
----                           -----
PSVersion                      5.1.22621.1778
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.22621.1778
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

PR #4280, opened 2024-02-02
Release 1.24.131.1

NikCharlebois:
* EXOAvailabilityAddressSpace
  * Added support for the TargetServiceEpr and TargetTenantId parameters.
  * Fixed the logic to retrieve existing instance by Forest Name.
* EXODistributionGroup
  * The Get function now retrieves the ModeratedBy and ManagedBy properties
    by the users' UPN instead of their GUID.
* EXOHostedContentFilterRule
  * Changed logic to retrieve the Rules by name. Using the Policy's name instead.
* EXOIntraOrganizationConnector
  * Fixes the DiscoveryEndpoint value from the Get method to include trailing
    forward slash.
* EXOMalwareFilterRule
  * Fixed an issue retrieving the right value for the Enabled property
* EXOOMEConfiguration
  * Fixes an error in the Get method where the ExternalMailExpiryInDays property
    wasn't properly returned.
* EXOSafeLinksPolicy
  * Deprecated the UseTranslatedNotificationText property
* TeamsEmergencyCallRoutingPolicy
  * Fix deletion of resource
    FIXES [#4261](https://github.com/microsoft/Microsoft365DSC/issues/4261)
* TEAMS
  * Added support for ManagedIdentity Authentication across Teams resources.
* DEPENDENCIES
  * Updated MSCloudLoginAssistant dependencies to version 1.1.10.
* MISC
  * Change the way to Export encoding is done so that it no longer relies
    on the Get-DSCResource function.

* N/A


salbeck-sit:
Module manifest version 1.24.131.1 still includes NestedModules = ...,  'Modules/EncodingHelpers/M365DSCStringEncoding.psm1' causing import to fail.

PR #4326, opened 2024-02-14
New unit test to check for wrongfully assigned ReadWrite-permissions

andikrueger:
Right now, there a few resources that would request write Permission even though only an export of the tenant is needed. This PR tries to add a quality gate to the settings.json files to only allow read permission for Microsoft graph in the “read” case.


ykuijs (approved):
LGTM

PR #4334, opened 2024-02-15
M365DSCUtil: Fix #4333

### LINKED ISSUE #4333: IntuneSettingCatalogCustomPolicyWindows10: Export gets ResourceInstanceName with the Id instead of Name
### Description of the issue

Now that *IntuneSettingCatalogCustomPolicyWindows10* is finally working with ConvertTo-DSCObject I was able to test it, and found a small problem when getting it through my internal tests which is when the export is done the ResourceInstanceName ends up with the Id instead of its Name like most other policies (which use DisplayName instead of Name as key and therefore show up correctly).

The issue is with the order of the conditions in the function *Get-M365DSCExportContentForResource* and fixing this is just a matter of hoisting the condition looking for the key *Name* right to just before the condition for the key *Id*.

### Microsoft 365 DSC Version

1.24.207.2

other

### The DSC configuration

```powershell
IntuneSettingCatalogCustomPolicyWindows10 "IntuneSettingCatalogCustomPolicyWindows10-fc8be7c8-b4ac-4821-8430-751cabc4138e"
        {
            Assignments          = @(
                }
                }
            );
            Description          = "IntuneSettingCatalogCustomPolicyWindows10_1";
            Id                   = "fc8be7c8-b4ac-4821-8430-751cabc4138e";
            Name                 = "IntuneSettingCatalogCustomPolicyWindows10_1";
            Platforms            = "windows10";
            Settings             = @(
                MSFT_MicrosoftGraphdeviceManagementConfigurationSetting{
                    SettingInstance = MSFT_MicrosoftGraphDeviceManagementConfigurationSettingInstance{
                        choiceSettingValue = MSFT_MicrosoftGraphDeviceManagementConfigurationChoiceSettingValue{
                            Value = 'device_vendor_msft_policy_config_updatev83diff~policy~cat_edgeupdate~cat_applications_pol_defaultcreatedesktopshortcut_0'
                        }
                        SettingDefinitionId = 'device_vendor_msft_policy_config_updatev83diff~policy~cat_edgeupdate~cat_applications_pol_defaultcreatedesktopshortcut'
                        odataType = '#microsoft.graph.deviceManagementConfigurationChoiceSettingInstance'
                    }
                }
            );
            Technologies         = "mdm";
        }
```


```powershell
N/A
```


```powershell
OsName               : Microsoft Windows 11 Enterprise                                                                                                                                                             OsOperatingSystemSKU : EnterpriseEdition                                                                                                                                                                           OsArchitecture       : 64-bit                                                                                                                                                                                      WindowsVersion       : 2009                                                                                                                                                                                        WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250                                                                                                                                                     OsLanguage           : en-US                                                                                                                                                                                       OsMuiLanguages       : {en-US, en-GB}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                  Name                           Value                                                                                                                                                                               ----                           -----                                                                                                                                                                               PSVersion                      5.1.22621.1778                                                                                                                                                                      PSEdition                      Desktop                                                                                                                                                                             PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                             BuildVersion                   10.0.22621.1778                                                                                                                                                                     CLRVersion                     4.0.30319.42000                                                                                                                                                                     WSManStackVersion              3.0                                                                                                                                                                                 PSRemotingProtocolVersion      2.3                                                                                                                                                                                 SerializationVersion           1.1.0.1
```

ricmestre:
If a resource has *Name* as key then it should be exported to the blueprint with it in its *ResourceInstanceName* title instead of appearing with *Id*, this is already done when the key is *DisplayName* but for *Name* is not working due to having the condition after *Id*.

EDIT: Just checked and there are several resources which have *Name* as Key or Required, hopefully this won't break any setup since Keys are supposed to be unique so having the *ResourceInstanceName* with the *Id* or *Name* should not conflict with other resources of the same type regardless of the key being used.

- Fixes #4333


Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3326,6 +3326,10 @@ function Get-M365DSCExportContentForResource
     {
         $primaryKey = $Results.DisplayName
     }
+    elseif ($Results.ContainsKey('Name'))
```
andikrueger, line 3329: Can we make sure, that the Name is a mandatory parameter? I'm wondering if there are any other resource that might use Name a a non-key-parameter.

andikrueger (commented):
Please have a look at my comment. I haven't run a test but am guessing there might be challenges with other resources.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
ricmestre: Yes, there's a few resources that use property *Name* and it's not mandatory, but then again the same already happens with *DisplayName* and could happen with any of the other properties.

The only way I see it working properly would be to add an extra parameter to the function where you could define the key but would involve changing all resources.

ricmestre: If this is not accepted I can live with that, it's not big of a problem for me but like I said the way the code is today it can cause problems where it's not supposed to happen since it relies on the ordering of the conditions instead of the actual knowledge of what's a key parameter, or not, of each resource.

andikrueger: Would it be possible to import the resource in this step and get the keys on the fly?

andikrueger: If I’m not mistaken, we do have this kind of functionality within the code. 🔎

ricmestre: We could always do something like this, with error handling of course but it works.

```powershell
# $ResourceName is passed as a parameter of Get-M365DSCExportContentForResource
$ResourceName = "IntuneSettingCatalogCustomPolicyWindows10"
(Get-DscResource -Name $ResourceName).Properties.Where({$_.IsMandatory}) | Select-Object -Property Name
```

```
Name
----
Name
```

andikrueger: Great find. This would be more generic also for future use.

ricmestre: Taking out the usual suspects for keys I still have these resources below that have more than one key and are not even covered by what's in the code today. From these with multiple keys we need to select a single one of them as part of the heuristics, which will be almost the same as what's there already but with *Get-DscResource* helping out.

```code
Name                           Value                                                                                                                                                                               ----                           -----                                                                                                                                                                               Name                           AADRoleEligibilityScheduleRequest                                                                                                                                                   Keys                           {Principal, RoleDefinition}                                                                                                                                                         Name                           M365DSCRuleEvaluation                                                                                                                                                               Keys                           {ResourceName, RuleDefinition}                                                                                                                                                      Name                           SCComplianceSearchAction                                                                                                                                                            Keys                           {Action, SearchName}                                                                                                                                                                Name                           SPOPropertyBag                                                                                                                                                                      Keys                           {Key, Url, Value}                                                                                                                                                                   Name                           SPOSiteAuditSettings                                                                                                                                                                Keys                           {AuditFlags, Url}                                                                                                                                                                   Name                           SPOSiteDesignRights                                                                                                                                                                 Keys                           {Rights, SiteDesignTitle}                                                                                                                                                           Name                           SPOStorageEntity                                                                                                                                                                    Keys                           {Key, SiteUrl}                                                                                                                                                                      Name                           TeamsGroupPolicyAssignment                                                                                                                                                          Keys                           {GroupDisplayName, PolicyType}                                                                                                                                                      Name                           TeamsUser                                                                                                                                                                           Keys                           {TeamName, User}
```

andikrueger: I would join them together to a longer string to stay unique.

ricmestre: Cool that is better for me than joining with dashes since I also rely on splitting this ResourceInstanceName on blueprints in order to do some stuff.

Will have a look into doing this, I'll update the PR once I have it tested.

ricmestre: @andikrueger I'm making a change here for the keys of the resources and now during a full Intune export I have 2 resources that complain about duplicate resources with the same key even when I go back using the original code!!!

I thought I was going crazy but I found out why, this is because of Graph 2.13.1 since it adds the *ResponseHeaders* as an additional object now on some queries and these 2 resources have a condition looking if the $getValue is an array which then throws an exception.

MSFT_IntuneDeviceEnrollmentStatusPageWindows10/MSFT_IntuneDeviceEnrollmentStatusPageWindows10.psm1:152:        if($getValue -is [Array])
MSFT_IntuneWindowsAutopilotDeploymentProfileAzureADJoined/MSFT_IntuneWindowsAutopilotDeploymentProfileAzureADJoined.psm1:130:        if($getValue -is [Array])

As I explained on https://github.com/microsoft/Microsoft365DSC/issues/4341 this is a problem with Graph quality on these past few releases, I really can't use these 2 policies until the issue with Graph is resolved so how can I proceed in not only solving #4333 but these additional issues?

ricmestre: Hopefully this week's release includes Graph 2.14.1 which solves the ReturnHeaderValues issue, until then I'll keep this PR as draft.

andikrueger: Please have a look at this PR #4348

ricmestre: Thank you I'm already aware but unfortunately I don't have time to look at this right at this moment.

ricmestre:
@andikrueger I had a little bit of time to look into it again, could you please check?

I'm calling Get-DscResource on the whole module to cache the information, so it will take longer to export the first resource but then all others will use the cached info otherwise it'd take longer to call Get-DscResource everytime per each resource.

Then I added as heuristics the usual mandatory keys plus Title since I think can be unique amongst other mandatory keys, all others if there are many will be joined together or if it's a single key then it will use that one, all remaining code afterwards was kept.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3361,38 +3361,50 @@ function Get-M365DSCExportContentForResource
     $Results = Format-M365DSCString -Properties $Results `
         -ResourceName $ResourceName
 
+    if ($Script:AllM365DscResources.Count -eq 0)
+    {
+        $Script:AllM365DscResources = Get-DscResource -Module 'Microsoft365Dsc'
+    }
+
     $primaryKey = ''
```
andikrueger, line 3369: This should be an array, or am I mistaken?

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3361,38 +3361,50 @@ function Get-M365DSCExportContentForResource
...
-    if ($Results.ContainsKey('IsSingleInstance'))
+    $Resource = $Script:AllM365DscResources.Where({ $_.Name -eq  $ResourceName })
+    $Keys = $Resource.Properties.Where({ $_.IsMandatory }) | `
+        Select-Object -ExpandProperty Name
+    if ($Keys.Contains('IsSingleInstance'))
     {
         $primaryKey = ''
```
andikrueger, line 3375: If we want to add the object to the array, we should use +=

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3361,38 +3361,50 @@ function Get-M365DSCExportContentForResource
...
 
+    if ($Script:AllM365DscResources.Count -eq 0)
+    {
+        $Script:AllM365DscResources = Get-DscResource -Module 'Microsoft365Dsc'
+    }
+
     $primaryKey = ''
-    if ($Results.ContainsKey('IsSingleInstance'))
+    $Resource = $Script:AllM365DscResources.Where({ $_.Name -eq  $ResourceName })
+    $Keys = $Resource.Properties.Where({ $_.IsMandatory }) | `
```
andikrueger, line 3371: As an alternative approach we could remove IsSingleInstance here and iterate over all keys and grab the values of the $Result list this:

PseudoCode:

$keys.Remove("IsSingleInstance")
$primaryKeys = @()
foreach($key in $keys){$primaryKeys += $Results.$Key}

andikrueger (changes requested):
Thanks for looking into it and the change. Just some additional ideas and suggestions.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
ricmestre: $primaryKey is not an array right now and with my code I don't change that, it's still a regular string. The intention here is to select the best primary key since it may have more than one and $primaryKey should only include one of them, see next comment.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
ricmestre: For resources that are IsSingleInstance the $primaryKey is IsSingleInstance itself but we want to end up with ResourceInstanceName equal to the name of the Resource itself, so this wouldn't work. what I additionally do is for resources that have multiple mandatory keys but are not the usual ones I join them into a single string so you'd get something like "ResourceName-ValueOfKey1ValueOfKey2ValueOfKey3", and for all others just "ResourceName-ValueOfKey1".

andikrueger: Thanks for the clarification.  Just read lines 3395 to 3407. This makes sense. Please ignore my comments then. :)

andikrueger: One question left, will we have unique Names for all Resources with this approach?

ricmestre: That's a side effect of what I wanted to achieve initially but yes that's the intention, with the code that it's there right now there's no 100% guarantee that happens.

I don't think the current unit tests are useful to test this change but I exported the whole Intune workload on 3 separate tenants, using this code, without issues and all resources got unique ResourceInstanceNames.

andikrueger: ... This would only apply in a scenario of a well setup tenant with unique names. All resources, that require multiple keys are only unique, if we use all of the keys. Or will this over-complicate things?

ricmestre: Ideally we will always want each resource to be unique or we'll have problems, so except for the IsSingleInstance case then yes in order to be totally unique would be to join all keys. Nevertheless for now I think this would be already better compared to what's there now, so if it's approved let it settle for a few weeks and if no one complains then we can change to join all mandatory keys except for resources that are IsSingleInstance since as previously mentioned those ones should end up with ResourceInstanceName = ResourceName.

andikrueger: Just did another review of the code and this change will be the best option and does not require any further changes, as there is  a further check for uniqueness in L3421+ which will mitigate possible duplicates and we would still end up with the best name and a numeric value in the end.

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
```diff
@@ -3361,38 +3361,50 @@ function Get-M365DSCExportContentForResource
...
     {
-        $primaryKey = $Results.Usage
+        $primaryKey = $Results.Id
+    }
+
+    if ([String]::IsNullOrEmpty($primaryKey))
+    {
+        if ($Keys.Count -gt 1)
+        {
+            $primaryKey = $Results.$Keys -Join ''
```
andikrueger, line 3402: This will create $null as $primaryKey. In this case, we should go with an foreach ($Key in $keys){$primaryKey += $Results.$key}

andikrueger (changes requested):
Please have a look at L3402

Modules/Microsoft365DSC/Modules/M365DSCUtil.psm1
ricmestre: You're right, it doesn't like to deserialize a string array. I'll update the whole chunk with the foreach.

andikrueger (approved):
LGTM

ricmestre:
@andikrueger Forgot about resources that are IsSingleInstance should end up with ResourceInstanceName = ResourceName so I updated the condition to only iterate through the Keys for resources that aren't IsSingleInstance.

andikrueger:
Thanks for having a second look at it.

andikrueger (approved):

PR #4360, opened 2024-02-22
Fix Windows SCEP Certificate Key Usage

FabienTschanz:
This pull request addresses an issue in the `IntuneDeviceConfigurationScepCertificatePolicyWindows10` resource, where the `KeyUsage` property was defined as a single String with a ValidateSet, whereas the exported configuration of that property could actually be a String with multiple entries, separated by a comma as the delimiter, e.g. `keyEncipherment,digitalSignature`.

I implemented option 3 from #4091 that describes to use a string collection and join them together in the Get and Set method to work around that issue and still have a pretty decent user experience. 

This is most likely considered a breaking change, but since only the export of the policy ever worked and as soon as one wanted to compile the `M365TenantConfig.ps1`, an error would occur since the configuration `keyEncipherment,digitalSignature` is not valid value when supplied. 

None

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationSCEPCertificatePolicyWindows10/MSFT_IntuneDeviceConfigurationScepCertificatePolicyWindows10.schema.mof
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationSCEPCertificatePolicyWindows10/MSFT_IntuneDeviceConfigurationScepCertificatePolicyWindows10.schema.mof
@@ -27,7 +27,7 @@ class MSFT_IntuneDeviceConfigurationScepCertificatePolicyWindows10 : OMI_BaseRes
     [Write, Description("Target store certificate. Possible values are: user, machine."), ValueMap{"user","machine"}, Values{"user","machine"}] String CertificateStore;
     [Write, Description("SCEP Hash Algorithm. Possible values are: sha1, sha2."), ValueMap{"sha1","sha2"}, Values{"sha1","sha2"}] String HashAlgorithm;
     [Write, Description("SCEP Key Size. Possible values are: size1024, size2048, size4096."), ValueMap{"size1024","size2048","size4096"}, Values{"size1024","size2048","size4096"}] String KeySize;
-    [Write, Description("SCEP Key Usage. Possible values are: keyEncipherment, digitalSignature.")] String KeyUsage[];
+    [Write, Description("SCEP Key Usage. Possible values are: keyEncipherment, digitalSignature."), ValueMap{"keyEncipherment","digitalSignature"}, Values{"keyEncipherment","digitalSignature"}] String KeyUsage[];
     [Write, Description("SCEP Server Url(s).")] String ScepServerUrls[];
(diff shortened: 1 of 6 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationSCEPCertificatePolicyWindows10/MSFT_IntuneDeviceConfigurationScepCertificatePolicyWindows10.schema.mof
```diff
@@ -27,7 +27,7 @@ class MSFT_IntuneDeviceConfigurationScepCertificatePolicyWindows10 : OMI_BaseRes
     [Write, Description("Target store certificate. Possible values are: user, machine."), ValueMap{"user","machine"}, Values{"user","machine"}] String CertificateStore;
     [Write, Description("SCEP Hash Algorithm. Possible values are: sha1, sha2."), ValueMap{"sha1","sha2"}, Values{"sha1","sha2"}] String HashAlgorithm;
     [Write, Description("SCEP Key Size. Possible values are: size1024, size2048, size4096."), ValueMap{"size1024","size2048","size4096"}, Values{"size1024","size2048","size4096"}] String KeySize;
-    [Write, Description("SCEP Key Usage. Possible values are: keyEncipherment, digitalSignature."), ValueMap{"keyEncipherment","digitalSignature"}, Values{"keyEncipherment","digitalSignature"}] String KeyUsage;
+    [Write, Description("SCEP Key Usage. Possible values are: keyEncipherment, digitalSignature.")] String KeyUsage[];
```
ykuijs, line 30: If the only accepted values are the two mentioned in the description, shouldn't we keep the ValueMap/Values properties? That means we are enforcing the input values, even though you can enter multiple values.

This is also done here: https://github.com/microsoft/Microsoft365DSC/blob/b3438f91f633c54907758f32b9980a6399f89542/Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantAccessPolicy/MSFT_AADCrossTenantAccessPolicy.schema.mof#L6

ykuijs (changes requested):
Added one comment, please review.

And about the "Breaking change" comment: Don't think that is the case here. We are moving from a single string to an array. If an existing config has a single string and the input expects an array, PowerShell automatically converts the string into an array. So everything should work as expected.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationSCEPCertificatePolicyWindows10/MSFT_IntuneDeviceConfigurationScepCertificatePolicyWindows10.schema.mof
FabienTschanz: Didn't know that was possible, thought the ValueMap only applies to String parameters and not arrays of Strings. I readded the validation back and checked, still works fine. Thanks.

ykuijs (approved):
LGTM

PR #4372, opened 2024-02-26
AAD workload: Show current values of resources in Test-TargetResource

ricmestre:

Looking at the AAD integration tests pipeline it looks like it's failing the test in AADGroup, but since it only shows the target values it's difficult to understand exactly why the test fails, it can be seen earlier in the logs that it's using a restored group instead of creating a new one, but it's still not clear without having access to the current values of the resource.

It seems that showing the current values wasn't only missing in AADGroup but for several other resources as well from the AAD workload so I added it to all that had it missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,39 @@
 
 # UNRELEASED
 
+* AADApplication
+  * Show current values of resource in Test-TargetResource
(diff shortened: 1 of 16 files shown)


andikrueger (approved):
LGTM

PR #4420, opened 2024-03-07
TeamsGroupPolicyAssignment: Skip assignments that have orphaned/deleted groups or without display name

### LINKED ISSUE #4407: TeamsGroupPolicyAssignment: Export stops if it finds an assignment with an orphaned/deleted group
### Description of the issue

While with a customer performing an export of their policies we found a problem with one of their TeamsGroupPolicyAssignments where the assignment was done to an already deleted group and therefore the export of TeamsGroupPolicyAssignment resources stopped, fortunately only skipped one additional assignment that wasn't exported but it means that with this problem it could basically skip dozens or hundreds of assignments if a single problem with an assignment is found.

Fix is to skip the assignment to the next one if group is not found.

### Microsoft 365 DSC Version

1.24.228.1

Teams

### The DSC configuration

```powershell
N/A
```

```powershell
N/A
```

```powershell
OsArchitecture       : 64-bit
WindowsVersion       : 2009
OsLanguage           : en-US
OsMuiLanguages       : {en-US, en-GB}

PSVersion                      5.1.22621.1778
BuildVersion                   10.0.22621.1778
```

ricmestre:

If an assignment has a group that has been deleted, and/or orphaned, then skip it instead of throwing an error otherwise all assignments won't be saved into the blueprint even for the ones that were successfully exported up until the error occurred.

- Fixes #4407


NikCharlebois:
We recently ran into a similar issue with AADConditionalAccessPolicy, where we were simply skipping invalid groups instead of throwing an error. The customer made a typo in the groups name, which resulted in the group being skipped and the policy got updated with 0 group in the ExcludedGroup policy which locked them out of their system. I believe we should use it as a rule of thumb to always error out when an invalid User or Group is specified as a value. Thoughts?

NikCharlebois (approved):

NikCharlebois:
After much thinking on this topic, the consensus should be that if an invalid group is specified, that the Set-targetResource should always throw the error, because otherwise it risks bringing the environment in an invalid state. From an Export perspective, it is fine to skip over those in my opinion. I will merge the PR.

PR #4430, opened 2024-03-08
AADUser advanced query filter support

### LINKED ISSUE #2430: AADUser: Export Filtering doesn't work as expected

I am looking for a way to export only users that are not synchronized. Thereby I came across the filters.
First tests here works without problems:
```powershell
@{AADUser = "StartsWith(UserPrincipalName, 'start')"}
@{AADUser = "UserPrincipalName eq 'xxx@xyz.com')"}
```
But if I try to use other filters it dosn't work:
```powershell
@{AADUser = "EndsWith(UserPrincipalName, 'xyz.com')"}
@{AADUser = "UserPrincipalName like '*@xyz.com')"}
or
@{AADGroup = "SecurityEnabled eq '$True'"}
```

Is the any documentation how to use the filters? Thank you very much!

Windows Server 2019

ModuleVersion '1.22.1019.1'

ifinch:
advanced query filter support

-Fixes #2430

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
@@ -914,7 +914,128 @@ function Export-TargetResource
             Property    = $propertiesToRetrieve
             ErrorAction = 'Stop'
         }
-        if ($Filter -like '*endsWith*')
+        $queryTypes = @{
+                        'eq'        = @('assignedPlans/any(a:a/capabilityStatus)',
+                                        'assignedPlans/any(a:a/service)',
+                                        'assignedPlans/any(a:a/servicePlanId)',
+                                        'authorizationInfo/certificateUserIds/any(p:p)',
+                                        'businessPhones/any(p:p)',
+                                        'companyName',
+                                        'createdObjects/any(c:c/id)',
+                                        'employeeHireDate',
+                                        'employeeOrgData/costCenter',
+                                        'employeeOrgData/division',
+                                        'employeeType',
+                                        'faxNumber',
+                                        'mobilePhone',
+                                        'officeLocation',
+                                        'onPremisesExtensionAttributes/extensionAttribute1',
+                                        'onPremisesExtensionAttributes/extensionAttribute10',
+                                        'onPremisesExtensionAttributes/extensionAttribute11',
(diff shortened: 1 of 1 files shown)


ifinch:
Hey @bartvermeersch ,

Those ones are precented in the advanced query doc
https://learn.microsoft.com/en-us/graph/aad-advanced-queries?tabs=http#user-properties

extension_4750531c75d84e8698ecd21bbd5da9f5_fruitPreference 
extension_4750531c75d84e8698ecd21bbd5da9f5_fruitPreference/any(p:p)
extension_4750531c75d84e8698ecd21bbd5da9f5_test

However, there is difference how MgUser vs MsBetaUsers and as MgUser used, some of the extensions should be cleaned up including those you pointed out.

ifinch:
Hi @NikCharlebois , anything extra needed to follow up on this removing from On-Hold state?

Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
```diff
@@ -914,7 +914,125 @@ function Export-TargetResource
...
+                            'usageLocation',
+                            'userType'
+                            )
+            }
+
+        # Initialize a flag to indicate whether the filter conditions match the attribute support
+        $allConditionsMatched = $true
+
+        # Check each condition in the filter against the support list
+        foreach ($condition in $Filter.Split(' '))
```
NikCharlebois, line 1021: We should check to make sure $Filter is not null first.

NikCharlebois (changes requested):

NikCharlebois:
My only concern here is that we will need to manage a hardcoded list of possible filters.

Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
ifinch: done

ifinch:
> My only concern here is that we will need to manage a hardcoded list of possible filters.

completely agreed, but not that many options we have now to make differently AFAIK

ifinch:
Hi @NikCharlebois any ETA on this to be merged?

ifinch (commented):
removed printing "Filter is null" on export with a Write-Host

NikCharlebois (approved):

PR #4432, opened 2024-03-08
TeamsVoiceRoute: Fix policy removal and also comparison in Test-TargetResource

ricmestre:

As already done on other Teams resources the *Confirm* switch cannot be used when calling the cmdlet to remove the policy, the cmdlets are not interactive so they don't need the switch and it actually makes the cmdlet fail.

While here also fix the comparison in Test-TargetResource.

Note: M365DSC doesn't support creating voice gateways yet so if they are referenced in TeamsVoiceRoute and not created an exception will be thrown.

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_TeamsVoiceRoute/MSFT_TeamsVoiceRoute.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_TeamsVoiceRoute/MSFT_TeamsVoiceRoute.psm1
@@ -316,16 +316,31 @@ function Test-TargetResource
     Write-Verbose -Message "Testing configuration of Voice Route {$Identity}"
 
     $CurrentValues = Get-TargetResource @PSBoundParameters
+    $ValuesToCheck = ([Hashtable]$PSBoundParameters).Clone()
+    $ValuesToCheck = Remove-M365DSCAuthenticationParameter -BoundParameters $ValuesToCheck
+
+    if ($CurrentValues.Ensure -ne $PSBoundParameters.Ensure)
+    {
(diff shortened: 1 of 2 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_TeamsVoiceRoute/MSFT_TeamsVoiceRoute.psm1
```diff
@@ -208,34 +209,27 @@ function Set-TargetResource
...
             $notFoundGatewayList += $gateway
         }
     }
 
-    if ($notFoundUsageList)
+    if ($notFoundGatewayList)
     {
         $notFoundGateways = $notFoundGatewayList -join ','
-        throw "Please create the Voice Gateway object(s) ($notFoundGateways) using `"TeamsVoiceRoute`""
+        throw "Please create the Voice Gateway object(s) ($notFoundGateways) using cmdlet `"New-CsOnlinePSTNGateway`""
```
andikrueger, line 221: Could you outline this change? Does the current recommendation not work?

ricmestre: No it doesn't, TeamsVoiceRoute cannot create voice gateways. You either create it with that cmdlet or a new resource needs to be added to M365DSC that does that.

andikrueger: Thanks for the info.

PR #4463, opened 2024-03-19
Check that Test-DscConfiguration and Get-DscConfiguration also function as expected in integration tests

Borgquite:
At present the integration tests do not appear to do either of these things:
- Check if resources are truly idempotent (if you run Test-DscConfiguration once a resource has been applied without errors, it should always return 'true'.
- Check if Get-TargetResource is working correctly without errors
This pull request is an attempt to resolve it. I fear it may throw up quite a number of errors in existing modules.

### DIFF
--- a/Tests/Integration/M365DSCTestEngine.psm1
+++ b/Tests/Integration/M365DSCTestEngine.psm1
@@ -97,6 +97,13 @@ function New-M365DSCIntegrationTest
     {
         Master -ConfigurationData $ConfigurationData -Credscredential $Credential
         Start-DscConfiguration Master -Wait -Force -Verbose -ErrorAction Stop
+
+        $TestDscConfigurationResult = Test-DscConfiguration Master -Verbose
+        if ($TestDscConfigurationResult.InDesiredState -ne $true) {
+            throw "Resources not in desired state: $($TestDscConfigurationResult.ResourcesNotInDesiredState.ResourceId)"
+        }
+
+        $GetDscConfigurationResult = Get-DscConfiguration -Verbose
     }
     catch
     {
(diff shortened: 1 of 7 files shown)


Borgquite:
I don't have a full testing M365 tenant for testing so hard for me to test this with the full range of resources. It is possible that I need to put a Start-Sleep between the Set and Test process to allow Microsoft 365 changes to replicate - any thoughts as to how long that should be?

Tests/Integration/Microsoft365DSC/M365DSCIntegration.AAD.Create.Tests.ps1
```diff
@@ -332,6 +332,13 @@
     {
         Master -ConfigurationData $ConfigurationData -Credscredential $Credential
         Start-DscConfiguration Master -Wait -Force -Verbose -ErrorAction Stop
+
+        $TestDscConfigurationResult = Test-DscConfiguration Master -Verbose
+        if ($TestDscConfigurationResult.InDesiredState -ne $true) {
```
NikCharlebois, line 337: These files are automatically generated and should never be modified manually

Tests/Integration/M365DSCTestEngine.psm1
```diff
@@ -97,6 +97,13 @@ function New-M365DSCIntegrationTest
     {
         Master -ConfigurationData $ConfigurationData -Credscredential $Credential
         Start-DscConfiguration Master -Wait -Force -Verbose -ErrorAction Stop
+
+        $TestDscConfigurationResult = Test-DscConfiguration Master -Verbose
```
NikCharlebois, line 101: This is already handled by the workflow. Have a look in the YAML file, we are doing a Test-DSCConfiguration for every integration test already.

NikCharlebois (commented):
Not sure we need this PR as most of it is already handled by the workflows' logic.

Tests/Integration/Microsoft365DSC/M365DSCIntegration.AAD.Create.Tests.ps1
Borgquite: Understood - will sort this out

Tests/Integration/M365DSCTestEngine.psm1
Borgquite: Great, thanks I didn't see that. Give me a little bit, I'll see if I can implement the same checks there...

PR #4464, opened 2024-03-19
Schema Update

andikrueger:

This will fix some issues within the schema definition. The schema was created with a regex based schema generator.


ricmestre:
@andikrueger The schema was generated with New-M365DSCSchemaDefinition so if there are issues on it they need to be solved in the corresponding schema.mof files :)

andikrueger:
@ricmestre: @NikCharlebois and I had a discussion about the tooling and basically built two solutions. I just included the regex based approach in here for your review.

ricmestre:
Cool! With regex in place is much better than relying on having the strings on the exact and correct place they're expected, but I still raised a new PR to fix those schemas your patch caught.

NikCharlebois (approved):

Issue #4437, opened 2024-03-11
AADAdministrativeUnit: Cannot leave 'Ensure' as default


Borgquite:
### Description of the issue

When creating an AADAdministrativeUnit resource, if the 'Ensure' field is not set, the resource is created correctly, but running Test-DscConfiguration afterwards still return 'False'. This also means that multiple runs of Start-DscConfiguration result in many AADAdministrativeUnits being created - in other words, it's not idempotent.

The apparent intention for the resource is that 'Ensure' defaults to 'Present'.

Setting Ensure = 'Present' manually resolves the issue but would rather this was not necessary.

### Microsoft 365 DSC Version

1.24.228.1

### Which workloads are affected

Azure Active Directory

### The DSC configuration

```powershell
$Microsoft365DSCApplicationId = Get-AutomationVariable -Name 'Microsoft365DSCApplicationId'
$Microsoft365DSCCertificateThumbprint = Get-AutomationVariable -Name 'Microsoft365DSCCertificateThumbprint'
$Microsoft365DSCOrganization = Get-AutomationVariable -Name 'Microsoft365DSCOrganization'

Configuration Example
{
    Import-DscResource -ModuleName Microsoft365DSC -ModuleVersion 1.24.228.1

    node localhost
    {
        AADAdministrativeUnit 'TestUnit'
        {
            Id                            = 'Test-Unit'
            DisplayName                   = 'Test-Unit'
            Description                   = 'Test-Unit1'
            ApplicationId                 = $Microsoft365DSCApplicationId
            CertificateThumbprint         = $Microsoft365DSCCertificateThumbprint
            TenantId                      = $Microsoft365DSCOrganization
        }
    }
}

$cd = @{
    AllNodes = @(
        @{
            NodeName = 'localhost'
            PSDscAllowPlainTextPassword = $true
        }
    )
}

Example -ConfigurationData $cd
```

### Verbose logs showing the problem

```powershell
Running the configuration:

Start-DscConfiguration -Path .\Example\ -Wait -Verbose -Force                                                                        
VERBOSE: Perform operation 'Invoke CimMethod' with following parameters, ''methodName' = SendConfigurationApply,'className' = MSFT_DSCLocalConfigurationManager,'namespaceName' =                 root/Microsoft/Windows/DesiredStateConfiguration'.                                                                                                                                                
VERBOSE: An LCM method call arrived from computer <redacted> with user sid <redacted>.                                                                        
[...]

Issue #2775, opened 2023-01-12
AADAdministrativeUnit: ScopedRoleMembers parameter Get-TargetResource returns 'Could not infer CimType from the provided .NET object.'


Borgquite:
#### Details of the scenario you tried and the problem that is occurring

Trying to run Get-DscConfiguration against an AADAdministrativeUnit which has Scoped Role Members defined results in errors.

- Create an AU via the Azure AD interface called 'TestUnit'
- Under 'Roles and administrators', assign an Administrative Role to the AU
- Run Get-DscConfiguration -Verbose against this configuration

The Get-DscConfiguration process will fail with an error (see below). There are more detailed errors in the 'Desired State Configuration' 'Operational' event log - the underlying error appears to be HResult -2147024809 'Could not infer CimType from the provided .NET object.'

Appears to be something wrong with the ScopedRoleMembers for AADAdministrativeUnit? If you remove the Administrative Role assigned to the AU, then run Get-DscConfiguration, it works fine.

#### Verbose logs showing the problem

PS C:\Users\user.name\Documents> Get-DscConfiguration -verbose                                                                                     
VERBOSE: An LCM method call arrived from computer COMPUTERNAME with user sid REDACTED.                                                     
VERBOSE: [COMPUTERNAME]: LCM:  [ Start  Get      ]                                                                                                                               
VERBOSE: [COMPUTERNAME]:                            [[AADUser]TestUser] Getting configuration of Office 365 User test.user@contoso.com                                            
VERBOSE: [COMPUTERNAME]:                            [[AADUser]TestUser] Getting Office 365 User test.user@contoso.com                                                             
VERBOSE: [COMPUTERNAME]:                            [[AADUser]TestUser] Found User test.user@contoso.com                                                                          
VERBOSE: [COMPUTERNAME]: LCM:  [ End    Get      ]  [[AADUser]TestUser]  in 3.4290 seconds.                                                                                      
VERBOSE: [COMPUTERNAME]:                            [[AADAdministrativeUnit]TestUnit] Found AU with id {REDACTED}, DisplayName {Test-Unit}           
VERBOSE: [COMPUTERNAME]:                            [[AADAdministrativeUnit]TestUnit] GET https://graph.microsoft.com/v1.0/directoryObjects/REDACTED  
with 0-byte payload                                                                                                                                                           
[...]

Issue #3458, opened 2023-07-11
Proposal/Request for improvement for automated deployment testing/assertion


ricmestre:
Given the multiple issues I've raised with resources not working in deployments would it be possible for you to create an environment where to test deployments of all resources of the projects? And then call Assert-M365DSCBlueprint to check back if there any drifts, this would ensure that not only if the resource was deployed correctly but also if the drifts checking is working properly or if it needs to be corrected. Ideally each resource should be tested with all the authentication methods that it supports.

On my side I'm currently only testing deployments with resources that we will be using (customers may be using completely different resources and therefore not tested) and I'm doing all this manually which means it has a really narrow scope of which resources may or may not work on a real deployment scenario and it's also extremely time consuming.

I'm perfectly aware this is not an easy feat since you'd have to ensure that the target tenant(s) would need to return to a clean state at the end of each test so that new tests run correctly, maybe you'd need several tenants in order to run several tests in parallel, have these tests run on public or private repo? It also wouldn't be able to test every single value the resource supports and all their combinations between them, but at least would be a start to have 1 or 2 examples per resource that could be tested, maybe even use the ones already included in the Unit Testing?

At the end of the day this would help me immensely since at the moment I'm basically adding conditions such as the below to essentially bypass resources that I know will fail the deployment (noticed by manual testing) but of course this is a hack since I won't be able to rule out every single one of the failing resources because I won't have time to test them all. This also means that something might go utterly wrong in a deployment in a customer's tenant (not only might it fail to deploy but deploy resources partially) and I won't know prior that the resource(s) that caused it was/were not working because I didn't test it.

```powershell
<#
    Workaround to bypass IntuneApplicationControlPolicyWindows10
    policies until issue
    https://github.com/microsoft/Microsoft365DSC/pull/3437 is not
    fixed
#>
if ($Resource.ResourceName -eq "IntuneApplicationControlPolicyWindows10") {
    continue
}
```

Really appreciate you take this in consideration!

ricmestre:
@andikrueger Very well put! See below some added context to the conversation.

Resources that are known to take a considerable amount of time to apply in the tenant should be left out of the testing.

When I said that it should test all authentication methods of all resources I meant that it should test the ones each resource supports, of course some of them don't support Service Principals (unfortunately) so that shouldn't be tested.

[...]

Issue #3194, opened 2023-04-20
AADAdministrativeUnit: SHOWSTOPPER - Removes all Members and ScopedRoleMembers during resource update when parameter is undefined


Borgquite:
I have just had manual AU assignments for 433 users and 721 groups wiped out on my tenant, because of this behaviour in Microsoft365DSC.

Sidenote: I've now raised a 10 [bugs](https://github.com/microsoft/Microsoft365DSC/issues?q=is%3Aissue+author%3ABorgquite+label%3ABug) (a couple still to be assigned the flag) and raised a number of times that (compared to other DSC modules I've used, see below) this module appears to lack sufficient unit testing, or integration testing. This bug demonstrates that point. I have no words right now, and maybe it's my fault somehow, but right now I can't think of anything more that I could have done which would have warned me about this, except for a line-by-line review of the code. I'm angry because I've trusted a module which I know is insufficiently tested, and I'm angry because the tests which should have caught this bug (which has existed for months) appear to be just... missing in action.

#### Details of the scenario you tried and the problem that is occurring

- Create an AU without specifying Members or ScopedRoleMembers parameters e.g. sample configuration below. _Please note that this is a perfectly normal thing to do and means I expect Microsoft365DSC to **leave those parameters** alone._
- Run Start-DscConfiguration - AU is created
- Use the Azure AD console to assign Users, Groups and Devices to the AU
- Run Test-DscConfiguration - no changes required - _so according to the Test functionality, the Set-TargetResource should NOT modify Members or ScopedRoleMembers because they are not defined - **this is the correct behaviour - and apart from doing a line-by-line review of your code, this is the only safety net to ensure I'm safe to run this on a production environment**._
- Update the Description of the AU in DSC and re-apply configuration
- MICROSOFT365DSC GOES AHEAD AND REMOVES EACH AND EVERY SINGLE MEMBER AND SCOPEDROLEMEMBER FROM THE AFFECTED ADMINISTRATIVE UNIT (see logs)

#### Verbose logs showing the problem

```
VERBOSE: Perform operation 'Invoke CimMethod' with following parameters, ''methodName' = SendConfigurationApply,'className' = MSFT_DSCLocalConfigurationManager,'namespaceName' =    root/Microsoft/Windows/DesiredStateConfiguration'.                                                                                                                                   
VERBOSE: An LCM method call arrived from computer COMPUTERNAME with user sid REDACTED.                                                           
[...]

PR #4470, opened 2024-03-20
Update integration tests to also check Get-TargetResource

Borgquite:
Have seen a few bugs myself where Get-TargetResource is failing (#2775) and sometimes this causes unexpected behaviour in the rest of the resource (#4437)

Hopefully this adds a simple sanity check using Get-DSCConfiguration to the integration tests to catch these before release

### DIFF
--- a/.github/workflows/Global - Integration - AAD.yml
+++ b/.github/workflows/Global - Integration - AAD.yml
@@ -88,6 +88,15 @@ jobs:
           {
             Write-Host "All resources in the Tenant are in the Desired State"
           }
+
+          try
+          {
+            $Result = Get-DSCConfiguration -Verbose -ErrorAction Stop
+          }
+          catch
+          {
+            throw $_
+          }
       - name: Generate {Update} Integration Tests from Examples
         shell: powershell
         run: |
@@ -145,6 +154,15 @@ jobs:
           else
           {
             Write-Host "All resources in the Tenant are in the Desired State"
+          }
+
+          try
+          {
+            $Result = Get-DSCConfiguration -Verbose -ErrorAction Stop
+          }
+          catch
+          {
+            throw $_
           }
       - name: Generate {Remove} Integration Tests from Examples
         shell: powershell
@@ -205,3 +223,11 @@ jobs:
               Write-Host "All resources in the Tenant are in the Desired State"
             }
 
+            try
+            {
+              $Result = Get-DSCConfiguration -Verbose -ErrorAction Stop
+            }
+            catch
+            {
+              throw $_
+            }


ricmestre:
But why? Test-DSCConfiguration calls the associated Test-TargetResource function of the resources in the config which in turn already call Get-TargetResource for each one of them.

Borgquite:
@ricmestre If I'm honest, it's a desperate attempt to try to improve the integration tests on this resource, a struggle that I'm aware you also share! (#3458). At the moment I'm dreading the times when I have to update Microsoft365DSC - let alone start to try to a new resource - because there's absolutely no guarantee that it's not going to break in a new, and unexpected way.

Sometimes, that new and unexpected way is, for example, going to wipe out entire sections of my Microsoft 365 tenant without warning me (#3194).

I'd describe myself as 'intermediate' at DSC, perhaps slightly less than 'intermediate' when it comes to module writing. The reasonf for this is that I tried to get my head around #4437 yesterday, and it turned out that the cause was a try {} catch loop that had been incorrectly duplicated - an error that wasn't noticed by the author, or the approver, or the existing integration or unit tests. I figure that if that sort of error can fall through the gaps in this project, it's worth testing Get-TargetResource itself, as I *think* an incorrect try {} catch around Get-TargetResource in Test-TargetResource would possibly fall through the gaps.

But mainly I'm just grasping around for something, anything, that would stop my headaches with this vital but sorely inadequately tested module.

ricmestre:
I understand since I found myself in the same position too many times and I'm not even using it in production yet, but in this case this particular PR won't solve much, in fact it's a duplication of work that isn't necessary.

Currently when something is merged into the repo it will run these workflows you changed, you can have a look into "Actions" (https://github.com/microsoft/Microsoft365DSC/actions) and look for "Global - Integration - {AAD,EXO,INTUNE}" and their output, there are resources that cause errors and impact the whole workflow, you can look at those.

Also, the tests these workflows call are present in folder "Modules\Microsoft365DSC\Examples\Resources" so you can look at what's being tested.

Borgquite:
@ricmestre As mention, I wouldn't use or update any resource here without extensive testing of the entire CRUD process for each variable, and I've learned the hard way that I need to do a line-by-line code review of each change before deploying any update. Even then I am missing bugs (like #4437 where I'm fairly sure before the changes, Get-TargetResource was just returning whatever was passed to it without even checking Microsoft Graph, therefore breaking Test-TargetResource).

If I run the following script, then the second catch {} doesn't fire. Doesn't that mean there is value in this patch?

```
# Test-TargetResource
try {
    write-host "hello"
    try {
        # Buggy Get-TargetResource here
        throw "hi"
        
    }
    catch
    {
        write-host "caught"
    }
}
catch
{
    write-host "not caught"
}
```

On the integration tests, it doesn't look like they fire on every merge, only some? For example, this one https://github.com/microsoft/Microsoft365DSC/actions/runs/8344518121 doesn't appear to have any integration test runs associated with it? Am I missing something?

ricmestre:
@NikCharlebois @andikrueger Could you please check this?

NikCharlebois (approved):

NikCharlebois:
I will merge the changes because there is indeed value in them. These "Global -" tests are set to be replaced by the new suite of integration tests as time permits. Currently the tests have been put in place for AAD, and partially for EXO.

Borgquite:
@NikCharlebois Thanks so much - appreciated!

Please could you let us know if there's a strong reason why the integration tests don't run on pull requests? (Looks like only pushes?)

PR #4479, opened 2024-03-22
[O365OrgSettings] Fixed incorrect permissions in settings.json

ykuijs:
For the O365OrgSettings resource, the Read permissions had Application.Read.All, but the Get method uses a "New" cmdlet that requires ReadWrite permissions. Corrected this in the settings.json.

It also corrects the resourcename property used in several settings.json files, that I found to be incorrect.


andikrueger (approved):
LGTM

PR #4485, opened 2024-03-22
Alternative solution for Advanced query filter for AADUser (and possibly other resources)

bartvermeersch:
In all other graph related frameworks, the user is responsible of specifying if he/she wants or needs to use the advanced filter capabilities. I would stick to that and not trying to replicate the internal graph logic by **adding a switch parameter** for when you want to use the advanced filter capabilities.

An alternative for https://github.com/microsoft/Microsoft365DSC/pull/4430

This PR is a Proof of Concept.

- Fixes [#2430](https://github.com/microsoft/Microsoft365DSC/issues/2430)

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
@@ -866,6 +866,10 @@ function Export-TargetResource
         $Filter,
 
         [Parameter()]
+        [Switch]
+        $AdvancedQueryFilter,
+
+        [Parameter()]
         [System.Management.Automation.PSCredential]
         $Credential,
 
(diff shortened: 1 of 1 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1
```diff
@@ -865,6 +865,10 @@ function Export-TargetResource
         [System.String]
         $Filter,
 
+        [Parameter()]
+        [Switch]
+        $AdvancedQueryFilter,
```
NikCharlebois, line 870: We shouldn't be modifying the signature of individual resources. There is no way to access this property since the Export is currently driven by the Export-M365DSCConfiguration cmdlet which doesn't expose that AdvancedQueryFilter parameter. Handling this properly would require significant changes in the overall Export logic.

bartvermeersch: OK, too bad. Duplicating the graph business logic for advanced queries into M365DSC won't be sustainable either I'm afraid.

PR #4490, opened 2024-03-25
AADAdministrativeUnit: Updated test & examples

Borgquite:
Have updated the examples for AADAdministrativeUnit to include setting Visibility and ScopedRoleMembers

This should make it easier to see how to get this to work, and allow this functionality to be tested in the integration tests

### DIFF
--- a/Modules/Microsoft365DSC/Examples/Resources/AADAdministrativeUnit/1-Create.ps1
+++ b/Modules/Microsoft365DSC/Examples/Resources/AADAdministrativeUnit/1-Create.ps1
@@ -27,7 +27,7 @@ Configuration Example
                 MSFT_MicrosoftGraphScopedRoleMembership
                 {
                     RoleName       = 'User Administrator'
-                    RoleMemberInfo = MSFT_MicrosoftGraphIdentity
+                    RoleMemberInfo = MSFT_MicrosoftGraphMember
                     {
                         Identity = "admin@$Domain"
                         Type     = "User"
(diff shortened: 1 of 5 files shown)


ricmestre:
@borgquite The document you're changing is automatically generated, the section specific to the examples can be tweaked by modifying the files present in folder *Modules\Microsoft365DSC\Examples\Resources* so you need to modify the examples for *AADAdministrativeUnit* found in that folder.

Borgquite:
@ricmestre Thanks for picking that up! Updated.

Modules/Microsoft365DSC/Examples/Resources/AADAdministrativeUnit/1-Create.ps1
```diff
@@ -23,7 +23,17 @@ Configuration Example
             MembershipRule                = "(user.country -eq `"Canada`")"
             MembershipRuleProcessingState = 'On'
             MembershipType                = 'Dynamic'
-            Ensure                        = 'Present'
```
NikCharlebois, line 26: Why are we removing the Ensure parameter?

NikCharlebois (changes requested):

Modules/Microsoft365DSC/Examples/Resources/AADAdministrativeUnit/1-Create.ps1
Borgquite: Hey - I was hoping to catch issues like #4437 where a breaking bug was only uncovered because I left Ensure undefined.
Since #4475 this shouldn't be necessary on any resource anyway
It does seem to be fairly normal to be able to omit 'Ensure' in DSC resources. If you'd prefer consistency, I could remove Ensure from all other 'create' / 'update' examples - or add it back here if you prefer. (Of course this would prevent other bugs like the one in #4437 being unearthed.

NikCharlebois: Ok thanks for the clarifications. Let's leave it as is for now then.

PR #4504, opened 2024-03-28
IntuneAntivirusPolicyWindows10SettingCatalog: Add missing properties

ricmestre:

While doing an export and trying to assert it I noticed that *ConvertTo-DSCObject* complained about a few properties on *IntuneAntivirusPolicyWindows10SettingCatalog* that were in the blueprint but not in the schema of the resource, this PR fixes this by adding those missing properties.

Description and allowed values taken from https://learn.microsoft.com/en-us/windows/client-management/mdm/defender-csp

*allowdatagramprocessingonwinserver*
*allownetworkprotectiondownlevel*
*disablednsovertcpparsing*
*disablehttpparsing*


ykuijs (approved):
LGTM

PR #4510, opened 2024-04-02
Adds Intune logo to Delta reports 

coaxke:
Adds Intune logo to `Get-IconPath` function when producing DSC Delta reports that contain Intune resources

When producing delta reports, the logo for Intune is currently an empty img src. This update simply points to Intune.jpg asset on the official Microsoft365dsc images static content path.

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
@@ -374,6 +374,10 @@ function Get-IconPath
     {
         return 'http://microsoft365dsc.com/Images/Teams.jpg'
     }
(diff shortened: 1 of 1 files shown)


coaxke:
If you'd be open to it, I'd happily make the change (in another PR) to have the function return the base64 image which could be embedded directly into the module and referenced similar to

```html
<img src="data:image/png;base64 iVBORw0KGgoAAAANSUhEUgAAABkAA[...]== alt="Intune Logo" />
```
 This would allow for the consumption of reports in environments that control outbound Internet access to specific-endpoints & remove the need for you to host these images.

NikCharlebois (approved):

Issue #4189, opened 2024-01-19
SPOTenantSettings: Missing support for property TenantDefaultTimezone


salbeck-sit:
### Description of the issue

The resource returns information that is available through Get-PnPTenant. The corresponding Graph-cmdlet Get-MgAdminSharepointSetting returns quite a number of the same properties and additionally TenantDefaultTimezone. The latter would be very nice to set first thing in order to not having to do it from every SPO-site being created.
I'm not sure if any other resources mix permissions/cmdlets like this - ie calling New-M365DSCConnection twice with different workloads - but I'd think that this would be the easy way. The PNP-module could be used to provide validation for the timezone via 'Get-PnpWeb -Includes RegionalSettings.TimeZones' (see https://www.sharepointdiary.com/2017/06/sharepoint-online-change-time-zone-using-powershell.html) - but another option is to just let the Update-cmdlet throw an error (and log it).

### Microsoft 365 DSC Version

DEV

### Which workloads are affected

SharePoint Online

andikrueger:
Are all current properties part of the MSGraph cmdLet?  Changing to the graph cmdLet would also allow for a better permission management in regards to access rights for SPO. https://learn.microsoft.com/en-us/graph/api/sharepointsettings-update?view=graph-rest-1.0&tabs=http

salbeck-sit:
Unfortunately not. There is some overlap but nowhere near enough for a direct replacement

PR #4546, opened 2024-04-11
SPOTenantSettings: Add property TenantDefaultTimezone

salbeck-sit:
This PR adds property TenantDefaultTimezone to SPOTenantSettings

Implements #4189

Note that the resource now connects to two workloads: PNP and MicrosoftGraph. Technically this shouldn't be an issue but as it's not normal, please advise. 
Changelog will be updated shortly


salbeck-sit:
Not entirely sure what is wrong. I'll have another look

salbeck-sit:
What a mess. I run tests on another machine and in this particular case something went very wrong when I assumed that I had copied everything and then managed to need two attempts to straighten it out

salbeck-sit:
reverted changelog because the commit blocked updates

NikCharlebois (approved):

PR #4637, opened 2024-05-04
Embeds Report Images

coaxke:

This PR embeds images displayed in delta reports as base64 encoded images

When delta reports are generated images are sourced from dependencies directory of module, base64 encoded and then embedded into report HTML.  This allows images to be rendered in environments that have restricted internet access and are no-longer delivered via HTTP.

image
image


coaxke:

PR #4679, opened 2024-05-15
Fix Properties of (Dynamic) Administrative Unit

bartvermeersch:
Some properties of Dynamic Administrative Units have moved from within Additionalproperties to the root object level when performing Get-MgBetaDirectoryAdministrativeUnit

image


bartvermeersch:
If I'm not mistaken in my analysis/findings the test also need to be updated?

https://github.com/microsoft/Microsoft365DSC/blob/473cd2ad6763cc8767cecd14c5596017eed6e0f2/Tests/Unit/Microsoft365DSC/Microsoft365DSC.AADAdministrativeUnit.Tests.ps1#L239-L242

NikCharlebois:
Can we please update the branch with the latest code from DEV? It appears the unit tests are failing. Thanks

bartvermeersch:
I think the unit tests were failing because they contain the same issue, but I want to have your confirmation. 

membershipType, membershipRule, membershipRuleProcessingState
 are no longer children of "AdditionalProperties" but are direct children of the Administrative Unit object in beta graph.

ykuijs:
@bartvermeersch You have just updated the code itself. The unit tests fail because those are not updated.

When unit tests run, they run on in an environment where there is no M365 environment available. So in order to be able to test the code, we need to fake (mock) the M365 cmdlets in such a way that they will return the data in the same format as the original cmdlet will.

Since the original cmdlet has changed, we now also need to update the mocked version of that cmdlet. As you see here, the mock is not yet updated: https://github.com/microsoft/Microsoft365DSC/blob/c24165e09604af93e4e041835cb88b042a22eb77/Tests/Unit/Microsoft365DSC/Microsoft365DSC.AADAdministrativeUnit.Tests.ps1#L233-L245

Please check the unit test file and make sure the required changes are also implemented there.

ykuijs (approved):
LGTM

PR #4736, opened 2024-06-05
Add JSON conversion to app configuration device policy

FabienTschanz:
Adds the JSON conversion to the `PayloadJson` property for better readability of the settings to the `IntuneAppConfigurationDevicePolicy` resource.

Example of the new version: 

```
Id                   = "23ab5a34-dc1e-4a53-84cc-fcb91a93a880";
PackageId            = "app:com.azure.authenticator";
PayloadJson          = "{`"kind`":`"androidenterprise#managedConfiguration`",`"productId`":`"app:com.azure.authenticator`",`"managedProperty`":[{`"key`":`"sharedDeviceRegistrationToken`",`"valueString`":`"asdf`"},{`"key`":`"sharedDeviceTenantId`",`"valueString`":`"asdf`"},{`"key`":`"sharedDeviceRegistrationPrefillUpn`",`"valueString`":`"{{IMEI}}`"},{`"key`":`"sharedDeviceMode`",`"valueBool`":false}]}";
ProfileApplicability = "default";
```

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * IntuneAppConfigurationDevicePolicy
   * Add assignment group display name and fix compilation
     FIXES [#4724](https://github.com/microsoft/Microsoft365DSC/issues/4724)
+  * Add conversion from `payloadJson` to actual JSON.
 * M365DSCResourceGenerator
   * Add support for generating Intune settings catalog policies
 * M365DSCDRGUtil
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppConfigurationDevicePolicy/MSFT_IntuneAppConfigurationDevicePolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppConfigurationDevicePolicy/MSFT_IntuneAppConfigurationDevicePolicy.psm1
@@ -200,11 +200,17 @@ function Get-TargetResource
             }
         }
 
+        $payloadJson = $null
+        if (-not [System.String]::IsNullOrEmpty($getValue.AdditionalProperties.payloadJson))
+        {
+            $payloadJson = [System.Text.Encoding]::ASCII.GetString([System.Convert]::FromBase64String($getValue.AdditionalProperties.payloadJson))
+        }
+
         $results = @{
             #region resource generator code
             ConnectedAppsEnabled  = $getValue.AdditionalProperties.connectedAppsEnabled
             PackageId             = $getValue.AdditionalProperties.packageId
-            PayloadJson           = $getValue.AdditionalProperties.payloadJson
+            PayloadJson           = $payloadJson
             PermissionActions     = $complexPermissionActions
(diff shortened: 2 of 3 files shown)


ricmestre:
Can't test it right now but it's pretty much in line with what I would expect, I'll give you more feedback later but it looks good to me :)

FabienTschanz:
@ricmestre Awesome, I checked it for myself and changed some values, and it all worked as intended. I sure hope you don't find a bug in that little code😄

andikrueger:
Just out of curiosity: will this also work with report generation (DSCParser) or are creating a new issue on this end?

I don’t know if JSON within a configuration is really something easy to handle. Any chance to save the json as hash table which would require less escaping?

ricmestre:
@FabienTschanz Short story, export and deployment/change afterwards is working, I've just tested it. Thank you!

@andikrueger Hi the long story is the reason of asking @FabienTschanz to convert the property content to JSON is that the persons that will use this don't know what JWT is (in fact not even JSON, but it's easier to explain than JWT). Next step about the escaping is that I also rely on having DSCParser working correctly, otherwise I cannot use the change in this PR, so I've also tested *ConvertTo-DSCObject* and it "works" and what do I mean by that?

Well the JSON content appears unescaped without backticks which is very good for my use case, the problem is that it splits the content in two strings, the first one empty and second one with all its content, see below so something must be changed in order to have it only in one string.

I also don't know why they chose the API to return the settings from iOS in an hashtable and for Android a JWT base64 string instead, and sure it would look better that way in its own CIM instance but that is more complex to do than just converting between base64/JWT<->JSON

image

FabienTschanz:
@ricmestre Thanks for the feedback. I will have a look at the `ConvertTo-DSCObject` function later in the evening why it spits out an array with an empty element at first. After that, I will create a report @andikrueger. Will keep you posted.

FabienTschanz:
@ricmestre Quick question: How does your configuration look like? I only get an array back from the `ConvertTo-DSCObject` function if I feed it a configuration with two policies, one iOS and one Android. If there is only an Android policy, it simply outputs the JSON as a normal string and not an array.

@andikrueger Generating a report with `New-M365DSCReportFromConfiguration` works without any issues. Outputs all the configuration and the JSON in a proper and correct way. For me it all looks good.

ricmestre:
@FabienTschanz To be perfectly honest with you I really don't know what the problem is, I just tried to replicate it and now I can't because it works, but this is because I can't use the same versions of the modules associated with M365DSC, there are so many problems that people for some reason don't encounter and I have to "surf" and either downgrade or upgrade the dependencies until I can make it work... It might look like I'm looking for problems, but for some reason I always find problems everywhere! For some reason now I can't replicate the issue, I'm using DSCParser 2.0.0.4 because of an issue introduced on 2.0.0.5, and upgraded to ReverseDSC 2.0.0.20, either way in terms of code to me this looks good to me.

NikCharlebois (approved):

PR #4742, opened 2024-06-07
O365 org settings fix parameter defaults to false if not specified

### LINKED ISSUE #4741: O365OrgSettings: Fix: Boolean Parameters default to false if not specified in configuration
### Description of the issue

If the parameters are not defined in the configuration file, they would default to $false instead of not being taken into consideration.

### Microsoft 365 DSC Version

1.24.605.1

Office 365 Admin

### The DSC configuration

_No response_

_No response_

_No response_

afloca:
Check that the Parameters are present in the configuration instead of defaulting to false if not specified

- Fixes #4741


FabienTschanz:
Shouldn't a setting go back to its default value if I unset it? Don't know what the corresponding default values for those settings are though. I'm wondering because for the settings catalog in Intune, that works exactly that way. But I'm not familiar with the O365 settings.

afloca:
> Shouldn't a setting go back to its default value if I unset it? Don't know what the corresponding default values for those settings are though. I'm wondering because for the settings catalog in Intune, that works exactly that way. But I'm not familiar with the O365 settings.

For most of the settings the default value is enabled, however the expectation is when specifying the desired configuration state on a set of parameters, not specified parameters should not be checked or set to a value (in this case false). 
In my case while trying to set the CortanaEnabled Setting, I was getting a 403 Error for the PlannerAllowCalendarSharing that was not specified in the configuration at all - this doesent solve the 403 for Planner, but allows usage of the rest of the settings independently.

ykuijs (approved):
LGTM

ykuijs:
@afloca, can you please update your branch with dev? After that we can merge this PR.

afloca:
@ykuijs Done

ykuijs (approved):
LGTM

PR #4768, opened 2024-06-15
Fix Intune device remedation assignment

### LINKED ISSUE #4726: IntuneDeviceRemediation: Assignments are exported without display name
### Description of the issue

Just like *IntuneAppConfigurationDevicePolicy* the assignments are not being exported with display name, so even though I'm able to apply the scripts into another target tenant they don't have the assignments due to different Ids.

@FabienTschanz Could you please also check this one? EDIT: Looks like you used the same code on the other resources you created so they all share the same problem.

### Microsoft 365 DSC Version

1.24.529.1

Intune

### The DSC configuration

```powershell
IntuneDeviceRemediation "IntuneDeviceRemediation-WMI Health"
        {
            Assignments              = @(
                    deviceAndAppManagementAssignmentFilterId = '00000000-0000-0000-0000-000000000000'
                    groupId = '5817c5d6-bb3b-4c75-83d7-938173627409'
                }
            );
            Credential               = $Credscredential;
            Description              = "Checks WMI health";
            DetectionScriptContent   = "REDACTED";
            DeviceHealthScriptType   = "deviceHealthScript";
            DisplayName              = "WMI Health";
            EnforceSignatureCheck    = $False;
            Ensure                   = "Present";
            Id                       = "1140f4fd-57c0-4ca5-8e45-82dd7132ddc1";
            Publisher                = "REDACTED";
            RemediationScriptContent = "";
            RoleScopeTagIds          = @("0");
            RunAs32Bit               = $True;
            RunAsAccount             = "system";
        }
```

```powershell
N/A
```

```powershell
N/A
```

FabienTschanz:
This pull request fixes an issue in the `IntuneDeviceRemediation` resource where the assignments were incorrectly exported. This would normally be a breaking change because the types etc. changed, but since the resource is very new and the assignments never have worked, I would still prefer to have that fixed than having to wait until the next breaking release.

- Fixes #4726


ricmestre:
@FabienTschanz Just tested your diff, you're almost there but there's still a few issues.

My remediation scripts were created by other members several years ago and they set them up to run hourly which don't have a time associated to the schedule, only an interval, but in your diff you always try to get the time with ```Get-Date``` which will fail since time is ```null``` in this case. On line 236 just before assigning the values to ```$assignmentResult``` add the following then change the assignment of sub-property ```Time``` to ```$Time``` a few lines down below:

```powershell
            if (-not [System.String]::IsNullOrEmpty($assignment.RunSchedule.AdditionalProperties.time))
            {
                $Time = Get-Date -Format 'HH:mm:ss' -Date $assignment.RunSchedule.AdditionalProperties.time
            }
            else
            {
                $Time = $null
            }
```

After I added the patch above I'm able to export but I've noticed that the Assignments are not properly indented, see below.

```powershell
            Assignments              = @(
                MSFT_IntuneDeviceRemediationPolicyAssignments{
                    RunSchedule = MSFT_IntuneDeviceRemediationRunSchedule{
                Interval = 1
                DataType = '#microsoft.graph.deviceHealthScriptHourlySchedule'
            }
                    RunRemediationScript = $False
                    Assignment = 'MSFT_DeviceManagementConfigurationPolicyAssignments{
                deviceAndAppManagementAssignmentFilterId = '00000000-0000-0000-0000-000000000000'
                groupDisplayName = 'REDACTED'
                groupId = 'REDACTED'
            }
                }
            );
```

And if you notice carefully you'll see an extraneous ```'``` just before ```MSFT_DeviceManagementConfigurationPolicyAssignments``` which makes the blueprint to fail to compile to MOF, you have some code to cope with this in ```Export-TargetResource``` but seems not all cases are taken care of.

FabienTschanz:
@ricmestre Thanks for the hint with the missing time. I just stumbled upon a better way to export the resources that I am just testing. Stay put, sorry for the effort you have to put in.

FabienTschanz:
@ricmestre Should now all be done. If you want to you can test again, it worked on my end including the correct indentation.

ricmestre:
@FabienTschanz The assignments are being added now, but ```Test-DSCConfiguration``` still fails the test and according to the output since the ```Assignments``` property still appears in ```Target values``` it means the comparison is not being done correctly in ```#Compare Cim instances``` chunk.

Another thing I've only noticed now is that you are not exporting Microsoft's own remediation scripts in ```Export-TargetResource```, is there a reason for that? The tenant from where I'm exporting these scripts is old and contains 2 of those scripts whereas I've a few other tenants where they are not present and I'd like to have them there as well.

```powershell
VERBOSE: [REDACTED]:                            [[IntuneDeviceRemediation]IntuneDeviceRemediation-REDACTED] Current Values: AccessTokens=$null
ApplicationId=***
ApplicationSecret=$null
Assignments=({Assignment={dataType=#microsoft.graph.groupAssignmentTarget
deviceAndAppManagementAssignmentFilterId=00000000-0000-0000-0000-000000000000
deviceAndAppManagementAssignmentFilterType=none
groupDisplayName=REDACTED
groupId=REDACTED}
RunRemediationScript=False
RunSchedule={DataType=$null
Date=$null
Interval=$null
Time=$null
UseUtc=$null}})
CertificateThumbprint=***
Credential=***
Description=
DetectionScriptContent=REDACTED
DetectionScriptParameters=()
DeviceHealthScriptType=deviceHealthScript
DisplayName=REDACTED
EnforceSignatureCheck=False
Ensure=Present
Id=REDACTED
ManagedIdentity=False
Publisher=REDACTED
RemediationScriptContent=
RemediationScriptParameters=()
RoleScopeTagIds=(0)
RunAs32Bit=False
RunAsAccount=system
TenantId=***
VERBOSE: [REDACTED]:                            [[IntuneDeviceRemediation]IntuneDeviceRemediation-REDACTED] Target Values:
Assignments=({RunRemediationScript=False; RunSchedule=; Assignment=MSFT_DeviceManagementConfigurationPolicyAssignments})
Description=
DetectionScriptContent=REDACTED
DeviceHealthScriptType=deviceHealthScript
DisplayName=REDACTED
EnforceSignatureCheck=False
Publisher=REDACTED
RemediationScriptContent=
RoleScopeTagIds=(0)
RunAs32Bit=False
RunAsAccount=system
VERBOSE: [REDACTED]:                            [[IntuneDeviceRemediation]IntuneDeviceRemediation-REDACTED] Test-TargetResource returned False
```

ricmestre:
By the way, it looks like one of the scripts actually passed the test, it's the only one that uses all devices assignments, all others which use a specific custom group fail the test.

FabienTschanz:
@ricmestre I'm not exporting the Microsoft scripts because they are global scripts. I can't edit the description, name, publisher, content of the scripts etc. The only thing I could change is the `RunAs32Bit`, `RunAsAccount` and the Assignments. Since some tenants have them and some do not, I wasn't per se exporting them too and supporting the more complex logic. But if that's something you would like to have, I can add that functionality too.

ricmestre:
I think we should export them anyway but only if those scripts can be re-created elsewhere, if not present, as global scripts as well, that property seems to be read-only though. I'm saying this because if they are supposed to be set as global in order to not have their content changed  then you could add logic to after being created only allow updating ```RunAs32Bit```, ```RunAsAccount``` and ```Assignments``` and write a warning for everything else.

In any case that should be addressed separately in another issue/PR, for now there's still that issue with the ```Assignments``` which I've just explained earlier.

FabienTschanz:
I wonder where the problem is. Can you provide me with a configuration example from Intune (screenshot maybe?) and from a .ps1?

ricmestre:
IntuneDeviceRemediation "IntuneDeviceRemediation-example"
        {
            Assignments              = @(
                MSFT_IntuneDeviceRemediationPolicyAssignments{

                    RunSchedule = MSFT_IntuneDeviceRemediationRunSchedule{
                        Interval = 1
                        DataType = '#microsoft.graph.deviceHealthScriptHourlySchedule'
                    }
                    RunRemediationScript = $False

                    Assignment = MSFT_DeviceManagementConfigurationPolicyAssignments{
                        deviceAndAppManagementAssignmentFilterId = '00000000-0000-0000-0000-000000000000'
                        groupDisplayName = 'example'
                        groupId = '1932487d-92f4-4c8e-a5fc-afce702ac597'
                    }

                }
            );
            Credential               = $Credscredential;
            Description              = "example";
            DetectionScriptContent   = "V3JpdGUtT3V0cHV0ICJIZWxsbyBXb3JsZCIK";
            DeviceHealthScriptType   = "deviceHealthScript";
            DisplayName              = "example";
            EnforceSignatureCheck    = $False;
            Ensure                   = "Present";
            Id                       = "bf180bbb-8464-4fb3-80d3-31afd3969044";
            Publisher                = "example";
            RemediationScriptContent = "";
            RoleScopeTagIds          = @("0");
            RunAs32Bit               = $True;
            RunAsAccount             = "system";
        }

FabienTschanz:
Thanks a lot, just tested the configuration and it works as expected 😢 Did you update the `MSFT_IntuneDeviceRemediation.psm1` and `M365DSCDRGUtil.psm1` with all of the content that was in my last commit? 

Are there any errors in the event log (Applications and Services Logs --> M365DSC)? The assignment update could potentially fail and not throw an error in the console, that would then only be visible in the event log.

ricmestre:
Yes I've updated both files, and as I mentioned before the assignments are actually being assigned what is failing is ```Test-DscConfiguration``` afterwards which is not making the comparison on the assignments correctly and it fails the test which also means that if I call ```Start-DscConfiguration``` again the test always fails and it will try to update the remediation script every time even if the assignments are actually already correct.

FabienTschanz:
I do understand the `Test-DscConfiguration` is failing, but it doesn't do that in my case with your configuration. Could it be that the groupId you are using is one that does not exist and because we are using the `groupDisplayName`, that will be found and the id updated? That would make sense, since I can't use `Compare-M365DSCIntunePolicyAssignment` which would handle that case, therefore I do the comparison with the entire object using `Compare-M365DSCComplexObject`.

ricmestre:
Ok I completely forgot M365DSC being dumb about the assignments comparison, after updating ```groupId``` to an existing one the test worked.

So I think it's good to go, but probably a thing to improve is when making comparisons and whenever a name is available along with the id of the object, regardless of being assignments or something else, it should compare the id first and then fall back the comparison to the name just like we already do when Getting/Setting several resources. FYI I have developed a monitoring tool which basically I have to exclude the Assignments property from being assessed across tenants because it will always report as drift since the Ids will be different.

FabienTschanz:
Puh, glad we found the issue laying in the configuration. To make sure that edge case is protected, I will update my code to check for the `groupDisplayName` if a `groupId` is found and if there is no name, I will use that id for the check. That way we can avoid a failing test on the assignments.

In your monitoring tool, are there any resources that work with the assignments check? Like the `IntuneAppConfigurationPolicy` or `IntuneAppConfigurationDevicePolicy`? They have a check for the assignments with `Compare-M365DSCIntunePolicyAssignment` that should work for cross-tenant configurations.

But now that we have discovered that there is the possibility of a mismatch, I should check if #4767 has issues when we export the `groupDisplayName`...

ricmestre:
I've excluded the property from being assessed since the beginning a couple of years ago, I didn't bother looking if any of the resources were now using ```Compare-M365DSCIntunePolicyAssignment``` recently.

Let me have a quick look and I'll let you know.

ricmestre:
Just tested ```IntuneAppConfigurationDevicePolicy``` and the comparison worked for another tenant, I didn't have any ```IntuneAppConfigurationPolicy``` handy but since it uses ```Compare-M365DSCIntunePolicyAssignment``` it will also work, although the code on that one is actually not quite right since it calls ```Compare-M365DSCComplexObject``` even if it's for ```Assignments``` and then calls ```Compare-M365DSCIntunePolicyAssignment``` so the object will be compared twice. It just needs to be updated to behave as you did on ```IntuneAppConfigurationDevicePolicy``` which uses if/else for that situation.

I guess it's a matter of adding that code to the DRG and update all resources, not in one go since not all of them have the same code and it's quite tricky to update those.

FabienTschanz:
Awesome, thanks a lot for the check 👍

FabienTschanz:
@NikCharlebois Ready for review.

ricmestre:
@FabienTschanz Weren't you looking into comparing the ```groupId``` and if it wasn't correct then fallback into comparing ```groupDisplayName``` instead?

FabienTschanz:
@ricmestre I found a way which works for all Intune resources. Implemented it over in #4767, see the last commit in `M365DSCDRGUtil`. In this PR, since the comparison doesn't work in multi-tenant scenarios, I won't be adding that functionality since it will be added without anything more to do with #4767.

ricmestre:
Just tested your changes in #4767 and the comparison works as intended, just a minor note that you might now be aware of, @NikCharlebois at one point said that each resource should be decoupled and have its own ```Assignments``` CIM instance instead of always being ```MSFT_DeviceManagementConfigurationPolicyAssignments``` so that if you need to update one instance you're not forced to change ALL resources that have that instance like I had to do to add property ```groupDisplayName```.

With that being said, at that time I also modified the resources ```IntuneAccountProtection*``` to each have their own instance for the assignments so your change in #4767 won't call ```Compare-M365DSCIntunePolicyAssignment``` because your if clause is expecting ```MSFT_DeviceManagementConfigurationPolicyAssignments``` where in this case it should also look at ```MSFT_IntuneAccountProtection*Assignments``` so maybe just change the if to

```powershell
if ($Source.$key.GetType().FullName -like '*CimInstance' -and ($Source.$key.CimClass.CimClassName -eq 'MSFT_DeviceManagementConfigurationPolicyAssignments' -or $Source.$key.CimClass.CimClassName -like 'MSFT_Intune*Assignments'))
```

FabienTschanz:
@ricmestre Affirmative, thanks for the hint. I wasn't aware of that discussion. 

Edit: I should probably thank you once more for the absolutely amazing job you do for this project. I'm really impressed by the time and effort you invest and the huge help you are for me, especially when testing the new and updated resources. Thanks a bunch.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceRemediation/MSFT_IntuneDeviceRemediation.schema.mof
```diff
@@ -35,8 +54,8 @@ class MSFT_IntuneDeviceRemediation : OMI_BaseResource
     [Write, Description("Indicate whether PowerShell script(s) should run as 32-bit")] Boolean RunAs32Bit;
     [Write, Description("Indicates the type of execution context. Possible values are: system, user."), ValueMap{"system","user"}, Values{"system","user"}] String RunAsAccount;
     [Key, Description("The unique identifier for an entity. Read-only.")] String Id;
-    [Write, Description("Represents the assignment to the Intune policy."), EmbeddedInstance("MSFT_DeviceManagementConfigurationPolicyAssignments")] String Assignments[];
-    [Write, Description("Present ensures the policy exists, absent ensures it is removed.")] string Ensure;
+    [Write, Description("Represents the assignment to the Intune policy."), EmbeddedInstance("MSFT_IntuneDeviceRemediationPolicyAssignments")] String Assignments[];
```
NikCharlebois, line 57: I believe this would be a breaking change in the resource. Existing configurations that include Assignments[] as DeviceManagementConfigurationPolicyAssignments would be broken by this change.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceRemediation/MSFT_IntuneDeviceRemediation.schema.mof
FabienTschanz: I understand, but that's unfortunately a necessity for this change. But the resource never actually worked with the assignments previously, so the configuration could be exported (with wrong assignments), and never correctly deployed. In my opinion, since the resource is relatively new and because of the fact that the assignments were broken, there should be the possibility to allow this change to be deployed.

ricmestre: @NikCharlebois Yes, Fabien created this resource just 1 month ago and had this problem I reported with the display name not being present, or let's call it didn't have this feature, but actually was also missing the run schedule so the deployed scripts wouldn't be of much use anyway.

NikCharlebois: I agree with your observation. Since the resource never actually work, we don't risk breaking it further. I will review and merge in today's release. Thanks for all your help Fabien, really appreciated.

NikCharlebois (approved):

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceRemediation/MSFT_IntuneDeviceRemediation.schema.mof
FabienTschanz: Thanks a bunch for the understanding, glad I can help.

PR #4823, opened 2024-07-01
EXOMailTips: Remove property Ensure

ricmestre:

This resource is of type ```IsSingleInstance``` therefore the property ```Ensure``` can be removed, no code needs to be adapted (except for removing the parameter) since the property is not being used anywhere.


FabienTschanz:
If we are preparing for a breaking release, I think we should include this one too.

NikCharlebois (approved):

PR #4834, opened 2024-07-03
Migrate Intune ASR resource to new settings catalog cmdlets

FabienTschanz:
This pull request migrates the Intune Attack Surface Reduction Rules resource to the new settings catalog cmdlets and fixes a couple of issues with `groupSettingCollection` and `choiceSetting`. 

None.


NikCharlebois:
Similar conversation as last release, but these would result in breaking changes if we remove parameters.

FabienTschanz:
@NikCharlebois Actually there shouldn't be any breaking changes, the removed parameter `AttackSurfaceReductionRules` does not have a corresponding entry in the MOF file and otherwise there are only new parameters being added. No parameters are being removed that are in use. Or did I miss something completely?

Edit: Just checked again, there is nothing being removed from the MOF file. This shouldn't be treated as a breaking change in my opinion. If `AttackSurfaceReductionRules` would have had an entry in the MOF, then I would agree, but not in this case.

NikCharlebois (approved):

NikCharlebois:
Good catch on the param not being in the schema.mof file. Completed review and will merge in this week's release. Thanks again for the help

PR #4841, opened 2024-07-04
Fix Intune policy assignment and complex object comparison

### LINKED ISSUE #4830: Intune: Assignment comparison not working for several resources
### Description of the issue

After upgrading to latest release I found out that at least these resources mentioned below now fail ```Test-DscConfiguration``` in my integration tests pipeline even if I'm still using the same blueprint as previous release where it was working. This seems due to changes in the way that comparisons are done in complex objects, this was changed to cope better with Intune Assignments property but clearly now creates other problems.

IntuneDeviceConfigurationCustomPolicyWindows10
IntuneDeviceConfigurationKioskPolicyWindows10
IntuneDeviceConfigurationNetworkBoundaryPolicyWindows10
IntuneDeviceConfigurationPolicyAndroidDeviceOwner
IntuneDeviceConfigurationPolicyMacOS
IntuneDeviceConfigurationVpnPolicyWindows10
IntuneDeviceEnrollmentStatusPageWindows10
IntuneSettingCatalogASRRulesPolicyWindows10
IntuneSettingCatalogCustomPolicyWindows10
IntuneWindowsAutopilotDeploymentProfileAzureADHybridJoined
IntuneWindowsAutopilotDeploymentProfileAzureADJoined

@FabienTschanz Looks like the tests I did that were successful the other day only succeeded by chance (e.g. this one is working ```IntuneDiskEncryptionMacOS```), the assignments are actually never compared as assignments or even as complex objects for that matter, that is because before even reaching that code the CIMInstance is split into ```$keys``` (M365DSCDRGUtil.psm1:L643) which are essentially the strings inside it, ```groupDisplayName```, ```groupId```, etc, so the object never matches that ```if``` block found in M365DSCDRGUtil.psm1:L683.

Another reason for failing is that not all resources in Intune return the property ```deviceAndAppManagementAssignmentFilterType``` when it's set to ```none``` so if you include it in the blueprint then it will complain that the objects are different. Same goes for returning ```deviceAndAppManagementAssignmentFilterId``` in an all 0's GUID, some that actually return the type as none not always return their Id so if you include it you will also have the same problem, or if you don't include it but the target resource in Intune expects it even if it's all 0's because the type was set to none. Important to note that this even changes every now and then, if I take an export today of the assignments of a certain Intune resource may be different of what it includes tomorrow, this is not a problem with the code but actually the API so the code must cope with this so that the same blueprint can be re-applied several times and ```Test-DscConfiguration``` always returns the same regardless.

And yet another problem, removing some comments to display the verbose statements found this on M365DSCDRGUtil.psm1:L741 ```Configuration drift - simple object key: dataType Source {#microsoft.graph.groupAssignmentTarget} Target {#microsoft.graph.exclusionGroupAssignmentTarget}``` which means it's trying to compare the assignment target with the exclusion target which shouldn't happen.

### Microsoft 365 DSC Version

1.24.626.1

other

### The DSC configuration

```powershell
IntuneWindowsAutopilotDeploymentProfileAzureADJoined "IntuneWindowsAutopilotDeploymentProfileAzureADJoined-IntuneWindowsAutopilotDeploymentProfileAzureADJoined_1"
        {
            Assignments                = @(
                    groupDisplayName = 'DummyGroupExclude'
                }
                    groupDisplayName = 'DummyGroupInclude'
                }
            );
            Credential                 = $Credscredential;
            Description                = "Windows Autopilot_Azure AD Join_Standard";
            DeviceNameTemplate         = "CMD%SERIAL%";
            DeviceType                 = "windowsPc";
            DisplayName                = "IntuneWindowsAutopilotDeploymentProfileAzureADJoined_1";
            EnableWhiteGlove           = $False;
            Ensure                     = "Present";
            ExtractHardwareHash        = $True;
            Id                         = "ef518b1f-fd62-4069-aca8-5344823b3f6d";
            Language                   = "os-default";
            OutOfBoxExperienceSettings = MSFT_MicrosoftGraphoutOfBoxExperienceSettings1{
                HideEULA = $True
                HideEscapeLink = $True
                HidePrivacySettings = $True
                DeviceUsageType = 'singleUser'
                SkipKeyboardSelectionPage = $True
                UserType = 'standard'
            };
        }
```

```powershell
Test-DSCConfiguration returns false
```

```powershell
Win11/PS5.1
```

FabienTschanz:
This pull request addresses an issue with Intune policy assignments, where multiple assignments are specified and not treated the same as a single assignment. Additionally, the assignment comparison was improved to handle with `00000000-0000-0000-0000-000000000000` filter ids and when the filter type is either not available or `none`.

The complex object comparison in `Compare-M365DSCComplexObject` was updated to reference the previous logic where one would call `Convert-M365DSCDRGComplexTypeToHashtable` to convert the CIM instance to a hashtable, which removed all null values from the object. Now only properties that are explicitly set to null will be checked, and other "implicit" null properties (meaning not specified) will be skipped.

@ricmestre Do you mind taking a look at the proposed changes and run them against your tests?

- Fixes #4830


ricmestre:
Unfortunately this doesn't solve the issue. Take the example below, after turning on verbose around it looks like it's returning back from the comparison early on because it's trying to compare the ```Description``` field of first object with second one (blueprint vs tenant of course). This happens on L740 then after returning from the recursive call back to L646 so the assignments are not even compared since it fails here.

```Configuration drift - simple object key: Description Source {Block Duplicated Edge Shortcuts - Custom ADMX} Target {Disable Edge Desktop Shortcut Creation}```

then

```Configuration drift - The complex array items are not identical```

```powershell
        IntuneDeviceConfigurationCustomPolicyWindows10 "IntuneDeviceConfigurationCustomPolicyWindows10-IntuneDeviceConfigurationCustomPolicyWindows10_1"
        {
            Assignments           = @(
                    groupDisplayName = 'DummyGroupExclude'
                }
                    groupDisplayName = 'DummyGroupInclude'
                }
            );
            Credential            = $CredsCredential;
            Description           = "IntuneDeviceConfigurationCustomPolicyWindows10_1";
            DisplayName           = "IntuneDeviceConfigurationCustomPolicyWindows10_1";
            Id                    = "ea99430e-6cec-4ee8-855c-fb7a67f0df1e";
            OmaSettings           = @(
                MSFT_MicrosoftGraphomaSetting{
                    Description = 'Block Duplicated Edge Shortcuts - Custom ADMX'
                    OmaUri = './Device/Vendor/MSFT/Policy/ConfigOperations/ADMXInstall/Windows/Policy/WindowsCustomizationsAdmx'
                    Value = '1'
                    odataType = '#microsoft.graph.omaSettingString'
                    DisplayName = 'Block Duplicated Edge Shortcuts - Custom ADMX'
                }
                MSFT_MicrosoftGraphomaSetting{
                    Description = 'Disable Edge Desktop Shortcut Creation'
                    OmaUri = './Device/Vendor/MSFT/Policy/Config/Windows~Policy~C_Edge/DisableEdgeDesktopShortcutCreation'
                    Value = '1'
                    odataType = '#microsoft.graph.omaSettingString'
                    DisplayName = 'Disable Edge Desktop Shortcut Creation'
                }
            );
            SupportsScopeTags     = $True;
        }
```

FabienTschanz:
@ricmestre Thanks for the feedback. That seems to be unrelated to the Intune assignments, but I'll look into it. Hope to have good news soon.

ricmestre:
Bear in mind that although unrelated to the assignments this was working before, on version 1.24.619.1 I don't have this problem, the comparison on the example I gave you works without issues.

FabienTschanz:
@ricmestre I think I found the issue from your case, but I will have a look at it again later. Pushed the current state of the fix.

ricmestre:
Just to let you know this version also doesn't solve this.

But I missed earlier that it also complained about ```IsEncrypted```, if the property being compared is not present then it shouldn't be compared do you agree?

```Configuration drift - key: IsEncrypted Source {null} Target {False}```

**Comment by FabienTschanz** [comment]
@ricmestre That handling didn't actually change. It's more that `Get-M365DSCDRGComplexTypeToHashtable` swallowed all null properties when converting the complex object to the hashtable, which explains why your blueprint now returns false because all properties, including the null values, are being compared. 

In my opinion, that's the state how it's supposed to be, because a property could be set to `$null` willingly or just be left. If we skip those values, a configuration change from a value to null would be skipped, which isn't the desired behavior.

I would rather suggest a change where one could specify the properties to exclude for a given type, which would then skip the comparison of those properties. That way, we could specify the `IsEncrypted` property. 

The reason why this property now pops up is because previously, the converted hashtable was without null values, and now I'm using the CIM class definition and comparing everything in that instance. 


**Edit**: Updated the code to only compare CIM properties that were effectively passed and not only because the exist in the CIM instance definition. Hope that fixes the issue, the custom policy works now.

**Comment by FabienTschanz** [comment]
@ricmestre The logic was updated yesterday to exclude null properties except if they were explicitly specified in the configuration. Otherwise, null properties in the source will be skipped. That's the only difference I could find compared to 1.24.619.1. 

I will have a look at the ASR rules next.

**Comment by ricmestre** [comment]
@FabienTschanz As far as comparison goes there's 2 that are still not working even with the latest patch, ```IntuneAppConfigurationPolicy``` which has issues comparing not only ```Assignments``` but also ```CustomSettings``` and ```IntuneDeviceConfigurationPlatformScriptMacOS``` which has issues as well but only on ```Assignments```. Test configurations for both below:

```powershell
        IntuneAppConfigurationPolicy "IntuneAppConfigurationPolicy-IntuneAppConfigurationPolicy_1"
        {
            Assignments           = @(
                    groupDisplayName = 'DummyGroupExclude'
                }
                    groupDisplayName = 'DummyGroupInclude'
                }
            );
            Credential            = $CredsCredential;
            CustomSettings        = @(
                MSFT_IntuneAppConfigurationPolicyCustomSetting{
                    name = 'com.microsoft.intune.mam.managedbrowser.AppProxyRedirection'
                    value = 'false'
                }
                MSFT_IntuneAppConfigurationPolicyCustomSetting{
                    name = 'com.microsoft.intune.mam.managedbrowser.AllowTransitionOnBlock'
                    value = 'true'
                }
                MSFT_IntuneAppConfigurationPolicyCustomSetting{
                    name = 'com.microsoft.outlook.Contacts.LocalSyncEnabled'
                    value = 'true'
                }
                MSFT_IntuneAppConfigurationPolicyCustomSetting{
                    name = 'com.microsoft.outlook.Contacts.LocalSyncEnabled.UserChangeAllowed'
                    value = 'true'
                }
            );
            DisplayName           = "IntuneAppConfigurationPolicy_1";
            Id                    = "44754094-9c69-4ee0-b587-d8af4efd4a7e";
        }
```

and 

```powershell
        IntuneDeviceConfigurationPlatformScriptMacOS "IntuneDeviceConfigurationPlatformScriptMacOS-IntuneDeviceConfigurationPlatformScriptMacOS_1"
        {
            Assignments                 = @(
                    groupDisplayName = 'DummyGroupExclude'
                }
                    groupDisplayName = 'DummyGroupInclude'
                }
            );
            Credential                  = $Credscredential;
            DisplayName                 = "IntuneDeviceConfigurationPlatformScriptMacOS_1";
            Ensure                      = "Present";
            BlockExecutionNotifications = $False;
            Description                 = "";
            ExecutionFrequency          = "00:00:00";
            FileName                    = "shellscript.sh";
            Id                          = "845175f7-5039-4779-9ad8-0cc750ef7214";
            RetryCount                  = 0;
            RoleScopeTagIds             = @("0");
            RunAsAccount                = "user";
            ScriptContent               = "ZWNobyAtZSAiSGVsbG8gV29ybGQiCg==";
        }
```

FabienTschanz:
@ricmestre Thanks, I'll take a look at it.

FabienTschanz:
@ricmestre I found the issue with the MacOS platform scripts, the assignments weren't correctly exported when filtered by display name (when the id wasn't found). That's now fixed with this PR. 

But I can't find an issue with the app configuration policy. That works fine in my environment, no matter how much I test. Tested with changing the value and name of the custom setting, added an additional entry etc. Same goes for the assignments, doesn't if I change the id, group assignment type etc. it always works. Can you find out which part of the settings and assignments is affected, so that I can try and reproduce the issue?

ricmestre:
The fix for ```IntuneDeviceConfigurationPlatformScriptMacOS``` worked.

Regarding ```IntuneAppConfigurationPolicy``` the policy is being created correctly, the problem is the comparison. Even if I create the policy, re-export it to a new blueprint and compare it with ```Test-DscConfiguration``` it always fails. You can check which comparison(s) is/are failing by adding ```Verbose``` switch to all calls to ```Compare-M365DSCComplexObject``` in the DSC resource and then inside ```M365DSCDRGUtil.psm1```.

FabienTschanz:
I understood that the comparison isn‘t working and that‘s what I tried to do, in my case it does the correct thing and only fails if there is an actual difference. If everything is correct, it returns true as expected. When I deleted, recreated and then ran the comparison, it succeeded and no discrepancy was found. I will try again.

ricmestre:
Still using the config I gave earlier it always fails for me, see below:

image

ricmestre:
Adding Verbose shows the following drifts on ```CustomSettings```

image

FabienTschanz:
@ricmestre Finally... Seems like my local installation was faulty, that's why it worked. After reinstalling and reapplying the changes from this PR I got the same error as you. Found the culprit, I am using the properties of the complex object for possible null detection, and a Microsoft Graph model is neither a CIM instance nor a hashtable, therefore the keys weren't populated correctly. That's now fixed by converting only those models to a hashtable, and then continuing with the comparison. 

Please check for your case, it should work now.

ricmestre:
Yep, it does work. I'll run my full Intune integration tests again and I'll get back to you with the results but it should be ok this time.

ricmestre:
So good news, with all patches including from #4845 all my Intune tests (New/Update/Remove) are working!!! :D 🚀

ykuijs (approved):
LGTM

PR #4872, opened 2024-07-13
Add Intune Windows quality update profile

### LINKED ISSUE #2659: IntuneQualityUpdateWindows10: New resource proposal
### Description

This resource would be used to control quality updates for Windows 10 and later

### Proposed properties

All

None

FabienTschanz:
This pull request adds the `IntuneWindowsUpdateForBusinessQualityUpdateProfileWindows10` as a DSC resource to manage the Windows quality updates in Intune. 

**Note**: Currently in draft because create and delete testing is pending due to missing Windows 10/11 Enterprise E3 license on the test tenant. 

- Fixes #2659


ricmestre:
It will be a pain also testing this in my automated tests since they run in a dev tenant without the required licensing but anyway...

So I've tested it in my live tenant and there's a backend issue, I created the policy manually, exported it, removed it and re-created it without issues but if I re-run the deployment it will always create a new policy. This is because the cmdlet ```Get-MgBetaDeviceManagementWindowsQualityUpdateProfile``` doesn't support filtering, at least not directly and requires someone to report this to the Graph SDK team, see at the bottom. The workaround is to in ```Get-TargetResource``` if the policy is not found by ```Id``` then retrieve all policies and filter them by ```DisplayName``` locally instead of sending the filter to the API.

```
Get-MgBetaDeviceManagementWindowsQualityUpdateProfile : The query specified in the URI is not valid. Query option 'Filter' is not allowed. To allow it, set the 'AllowedQueryOptions' property on EnableQueryAttribute or QueryValidationSettings.
Status: 400 (BadRequest)
```

FabienTschanz:
@ricmestre Should have spotted that one because I replaced the filter in the export method with another way to filter. Updated it now, I will test it tomorrow in my company test tenant.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneWindowsUpdateForBusinessQualityUpdateProfileWindows10/MSFT_IntuneWindowsUpdateForBusinessQualityUpdateProfileWindows10.psm1
```diff
@@ -94,9 +94,9 @@ function Get-TargetResource
 
             if (-not [System.String]::IsNullOrEmpty($DisplayName))
             {
-                $getValue = Get-MgBetaDeviceManagementWindowsQualityUpdateProfile `
-                    -Filter "DisplayName eq '$DisplayName'" `
-                    -ErrorAction SilentlyContinue
+                $getValue = Get-MgBetaDeviceManagementWindowsQualityUpdateProfile | Where-Object -FilterScript {
```
ricmestre: You want to add ```-All``` here to retrieve all profiles first and maybe also keep ```-ErrorAction SilentlyContinue``` otherwise if the cmdlet fails you go straight to the next catch block instead of returning ```$nullResult```.

FabienTschanz: Thanks for being the good eyed watcher. Just updated it, totally makes sense.

FabienTschanz:
Checked in my company test tenant, looks good.

NikCharlebois (approved):

PR #4881, opened 2024-07-15
correct location of generated cert

### LINKED ISSUE #4880: Update-M365DSCAzureAdApplication creates certificate in wrong store
`Update-M365DSCAzureAdApplication` when used with the `-CreateSelfSignedCertificate` switch, creates a certificate in the CurrentUser store.
Unfortunately, the other scripts in the module assume certificates specified with `-CertificateThumbprint` are in the LocalMachine store.
Passing the thumbprint as `..\..\LocalMachine\My\abced...` does not work either.

I suggest we change the location of the generated cert to match the expectation and make it easier for new users to correctly authenticate with an app ID and certificate.

Additionally, this should be combined with `-certificatePath` to export the cert to a file. However, the export is only the public key which cannot be used for authentication. Is the intention here that the export should be a pfx?

ketumbra:
The DSC scripts assume the certificate (specified by thumbprint) is in the localmachine store, not current user. Corrections to help output regarding use of both CertificatePath and CreateSelfSigned params.

This PR addresses the above issues by generating the cert in the LocalMachine store and correcting the parameter comments.

- Fixes #4880

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
@@ -1246,7 +1246,7 @@ If specified, a new secret will be created for the application. -CreateNewSecret
 The path to the certificate to be uploaded for the app registration. If using with -CreateSelfSignedCertificate - a file with this name will be created and uploaded (file must not exist). Otherwise the file must already exist. Cannot be used with -CreateNewSecret simultaneously.
 
 .Parameter CreateSelfSignedCertificate
-If specified, a self-signed certificate will be created for the application. -CreateSelfSignedCertificate or -CertificatePath can be used, not both.
+If specified, a self-signed certificate will be created for the application.
 
 .Parameter AdminConsent
 If specified, admin consent will be granted for the application.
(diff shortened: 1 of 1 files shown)


Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
```diff
@@ -1714,7 +1714,7 @@ function Update-M365DSCAzureAdApplication
                     if ($CreateSelfSignedCertificate)
                     {
                         Write-LogEntry '    CreateSelfSignedCertificate specified, generating new Self Signed Certificate'
-                        $cerCert = New-SelfSignedCertificate -CertStoreLocation 'Cert:\CurrentUser\My' `
+                        $cerCert = New-SelfSignedCertificate -CertStoreLocation 'Cert:\LocalMachine\My' `
```
NikCharlebois, line 1717: This part depends on what the intent is. If the user wants to do an Export, than the CurrentUser store is where the certificate should be placed. If it's a logic path using the LCM (e.g., Start/Test-DSCConfiguration) then LocalMachine is required.

ketumbra: but currently it does neither an export-pfx nor places it in a store for subsequent use. So as a new user, I cannot use this scriptlet to configure a service principal for me.

PR #4898, opened 2024-07-17
write failures to error output

ketumbra:
Change failures from verbose to error output

During application of a MOF, not all resources were created. Finding out why required searching through the verbose output. Errors should be output using the error stream so they are highlighted and easily available.

- None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+* MSFT_AADConditionalAccessPolicy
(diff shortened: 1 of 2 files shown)


NikCharlebois (changes requested):
Thanks, can you also add an entry in the Changelog.md file to describe the change?

ketumbra:

ketumbra:
Sure thing. Done.

NikCharlebois (approved):

PR #4919, opened 2024-07-23
Add Intune EDR Linux Profile

FabienTschanz:
This Pull Request adds the `IntuneEndpointDetectionAndResponsePolicyLinux`, the Linux variant to `IntuneEndpointDetectionAndResponsePolicyWindows10`. 

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * IntuneAttackSurfaceReductionRulesPolicyWindows10ConfigManager
   * Migrate to new settings catalog cmdlets.
     FIXES [#3966](https://github.com/microsoft/Microsoft365DSC/issues/3966)
+* IntuneEndpointDetectionAndResponsePolicyLinux
+  * Initial release.
 * IntuneWindowsUpdateForBusinessFeatureUpdateProfileWindows10
   * Introduces new properties and updates the handling of the  
     start and end dates.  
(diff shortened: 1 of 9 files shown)


ricmestre:
While testing this one I saw in the logs during ```Test-TargetResource``` that all Target Values were set to $null, this is because instead of $PSBoundParameters like used everywhere else you used $ValuesToCheck so I caught that.

```powershell
Write-Verbose -Message "Target Values: $(Convert-M365DscHashtableToString -Hashtable $ValuesToCheck)"
```
```powershell
[[IntuneEndpointDetectionAndResponsePolicyLinux]IntuneEndpointDetectionAndResponsePolicyLinux-IntuneEndpointDetectionAndResponsePolicy Linux_1] Target Values:
Description=$null
DisplayName=$null
RoleScopeTagIds=$null
tags_item_key=$null
tags_item_value=$null
```

This is caused by the following chunk where you're trying to add key/values to $null if they're not null but you're not populating them before that with $PSBoundParameters.Clone() which means that all resources where you added this are not performing correct testing.

```powershell
    [Hashtable]$ValuesToCheck = @{}
    $MyInvocation.MyCommand.Parameters.GetEnumerator() | ForEach-Object {
        if ($_.Key -notlike '*Variable' -or $_.Key -notin @('Verbose', 'Debug', 'ErrorAction', 'WarningAction', 'InformationAction'))
        {
            if ($null -ne $CurrentValues[$_.Key] -or $null -ne $PSBoundParameters[$_.Key])
            {
                $ValuesToCheck.Add($_.Key, $null)
                if (-not $PSBoundParameters.ContainsKey($_.Key))
                {
                    $PSBoundParameters.Add($_.Key, $null)
                }
            }
        }
    }
```

EDIT: You just need the keys there, not the values, forget it. I've seen the latest commit on the verbose message so it's ok. I'll test it a little bit further.

FabienTschanz:
@ricmestre Thanks for the hint. I changed the output from `$ValuesToCheck` to `$PSBoundParameters`. 
The comparison should still work, this is only a console output issue. The comparison is actually done by the `PSBoundParameters` with the `$ValuesToCheck.Keys` as the keys to check. No issues should occur while comparing.

ricmestre:
It doesn't like to be created on my dev tenant (probably licensing issues) but it works without issues on a prod tenant.

ricmestre:
And just like that now it also works in the dev tenant, what do you know... I'll try the other similar resource for MacOS now.

NikCharlebois (approved):

PR #4957, opened 2024-08-08
ExoAuthenticationPolicyAssignment: Return all users when exporting

### LINKED ISSUE #4956: ExoAuthenticationPolicyAssignment: Only returning 1000 users for each authentication policy when exporting
When exporting the authentication policy assignments, we are maxing out at 1000 users per authentication policy. This is due to a default limit in the Get-User cmdlet.

adhodgson1:
When exporting Exchange authentication policy assignments, only 1000 users are returned with each authentication policy. This PR adds the unlimited result size parameter to the Get-User call while exporting the resource.

Fixes #4956

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # UNRELEASED
 
+* EXOAuthenticationPolicyAssignment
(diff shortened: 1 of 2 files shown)


ykuijs (approved):
LGTM

PR #4976, opened 2024-08-21
IntuneDeviceCompliancePolicyWindows10: Fix extraction of property `TpmRequired`

ricmestre:

Currently property *TpmRequired* is not being exported, this is due to a typo which this PR fixes.

At the same time we also need to change the casing for the PSBoundParameters where this property is used due to how function *Get-M365DSCIntuneDeviceCompliancePolicyWindows10AdditionalProperties* works which turns only the first letter to lower case instead of the whole first word.


ykuijs (approved):
LGTM

PR #5056, opened 2024-09-17
Add Intune Firewall Policy for Windows 10

### LINKED ISSUE #3033: IntuneFWMicrosoftDefenderFirewallPolicyWindows10: New resource proposal
### Description

This resource would be used to export/apply Firewall policies for Microsoft Defender Firewall

### Proposed properties

All

The cmdlets to deal with this resource in Graph are {New,Get,Update}-MgDeviceManagementIntent{Setting} and are currently already used in resource MSFT_IntuneASRRulesPolicyWindows10

### LINKED ISSUE #5055: Update-IntuneDeviceConfigurationPolicy - Does not throw an exception if an error occurs
### Description of the issue

Currently, the cmdlet `Update-IntuneDeviceConfigurationPolicy` does not throw if an error occurs during execution. Instead, it simply logs the error to the event channel and returns null to the caller. 

It currently looks like the following:
```powershell
try
    {
        Invoke-MgGraphRequest -Method PUT -Uri $Uri -Body $body -ErrorAction Stop
    }
    catch
    {
        New-M365DSCLogEntry -Message 'Error updating data:' `
            -Exception $_ `
            -TenantId $TenantId `
            -Credential $Credential

        return $null
    }
```

As a consumer of the cmdlet, I would like to know if an error happened and for me, it would be the best if the execution is stopped and the exception is thrown to the higher levels to have them handle the issue. So the updated catch clause should look like the following in my opinion: 

```powershell
catch
    {
        New-M365DSCLogEntry -Message 'Error updating data:' `
            -Exception $_ `
            -TenantId $TenantId `
            -Credential $Credential

        throw
    }
```

Instead of returning null, I believe it's better if we rethrow the exception to stop the execution flow and let the caller of the cmdlet handle the exception. What do you think?

For visibility: @ricmestre 

### Microsoft 365 DSC Version

DEV

Intune

### The DSC configuration

_No response_

_No response_

_No response_

FabienTschanz:
This PR adds support for the Intune resource `IntuneFirewallPolicyWindows10`. This is the first in a series of resources for the Firewall blade in the Endpoint security section of Intune. 

Additionally, it updates the resource generator and M365DSCDRGUtil module for more complex property handling, including properties that share the same name but are under separate instance templates (which has not happened before). To manage that and still generate unique names, it uses recursive traversing of the `OffsetUri`, which indicates how deep the current property is in the settings tree. 

- Fixes #3033
- Fixes #5055


ricmestre:
Works without issues, I tried to create a couple of policies based off of some I already had created on another production tenant, made some changes to them and finally removed them.

Since you also changed M365DSCDRGUtil I'll run my whole integration pipeline to ensure it doesn't impact anything else, but that takes a couple of hours, I'll let you know the result.

FabienTschanz:
Thanks a lot @ricmestre. The biggest concern I have is backwards compatibility, so that was my main focus with the changes to the M365DSCDRGUtil. Since the updated unit tests of the settings catalog policies all have a quite good settings coverage and types, I don't necessarily expect any issues, but I can't be certain. 

Developing for something new while retaining backwards compatibility without breaking anything is very hard... Especially if the settings catalog keeps getting more and more complex. I'm absolutely no fan of it, but that's life. 

**Edit:** Almost everything in the `IntuneFirewallPolicyWindows10` resource was auto-generated by the resource generator. I do think we can make quick progress on the new resources 😄

ricmestre:
Settings catalog policies are always a pain in the rear, for me specially the main one *IntuneSettingCatalogCustomPolicyWindows10* with "infinite" recursive nested levels...

Anyway, all other policies I have tests for still work as expected so it's good to go for me.

FabienTschanz:
@NikCharlebois Ready for review.

FabienTschanz:
Unit tests failed because of a failed git task. Tests were not executed.

FabienTschanz:
@NikCharlebois Can you please take a look at this one as well? Then I can prepare my two draft PRs for review. Thank you.

NikCharlebois (approved):

PR #5075, opened 2024-09-20
Add Intune ASR App and Browser Isolation Policy for Windows 10 ConfigMgr

### LINKED ISSUE #3028: IntuneASRAppAndBrowserIsolationPolicyWindows10: New resource proposal
### Description

This resource would be used to export/apply Attack Surface Reduction policies for App and Browser Isolation

### Proposed properties

All

The cmdlets to deal with this resource in Graph are {New,Get,Update}-MgDeviceManagementIntent{Setting} and are currently already used in resource MSFT_IntuneASRRulesPolicyWindows10

FabienTschanz:
This PR adds the Intune resource `IntuneAppAndBrowserIsolationPolicyWindows10ConfigMgr`, which resides under the Endpoint Security --> Attack surface reduction blade. 

None.


ricmestre:
@FabienTschanz Please bear in mind that my proposal #3028 was for the regular policy and not the one for SCCM

FabienTschanz:
@ricmestre Thanks and yes, I thought that would be the case. There is something a bit fishy there, the one for Windows says that Application Guard is going to be deprecated, but for the SCCM one, it doesn't say that. So my thoughts were to create the SCCM one and close the issue, but now I'll update the changelog and the description to not include your issue. 

I will create a separate one for your issue. 

image

ricmestre:
This is really an old one, I didn't have that deprecation notice at the time 😄 

I just looked at it and also see that notice, but it would be nice to mention there which policy, or policies, replaces it... A quick look and I think that those settings can be found inside the resource *IntuneDeviceConfigurationEndpointProtectionPolicyWindows10* since they all start with *ApplicationGuard* and also look like they setup the same settings as *App And Browser Isolation*.

I guess that it's still needed, if it's not much effort creating it with DRG then adding a notice that it will be deprecated in the near future would be much appreciated. That way if we have any customers with it we could improve their setup by proposing to move to the "new" policy.

FabienTschanz:
Personally, I wouldn't add a deprecation notice yet, because if we add one here, we theoretically must add another one on the "new" resource as well for those settings... Microsoft will tell us if they deprecate the entire policy, so we could leave it at the moment and wait for their move.

FabienTschanz:
Unit tests are failing because of `AADIdentityB2XUserFlow` create and update examples.

FabienTschanz:
@NikCharlebois This PR is ready for review, can you take a look at it? Thank you.

FabienTschanz:
Still waiting for a review @NikCharlebois 😃 Would be awesome if I can remove this resource from my todo list.

NikCharlebois (approved):

PR #5083, opened 2024-09-23
Adding MdcSubscriptionDefenderPlan

MatanShabtay:
Adding MdcSubscriptionDefenderPlan DSC resource.
Which allows to enable or disable Microsoft Defender plans for a subscription in Microsoft Defender for Cloud.
For more information about the available Defender plans, sub plans and plan extensions refer to Defender for Cloud onboarding API documentation.
https://learn.microsoft.com/en-us/rest/api/defenderforcloud/pricings/update?view=rest-defenderforcloud-2024-01-01&tabs=HTTP

Important things to note in this resource:
1. Since this tool runs in tenant scope while Azure Powershell runs in subscription scope, Export-TargetResource is being implemented by querying Azure Resource Graph on the tenant scope
2. To allow users to actually export configuration and use it on another tenant, the subscription name is used as a key, and not the subscription ID (similar to AzureSubscription DSC resource which was used for inspiration)
3. The underlying Defender for Cloud commands (Set-AzSecurityPricing and Get-AzSecurityPricing doesnt accept a subscription ID as parameters, the subscription ID being used is the subscription of the logged-on Az Powershell session, therefore the subscription ID of the session is being set in Get-TargetResource and Set-TargetResource using Set-AzContext
4. Need to make sure that all Az Powershell is a dependency of this tool? or a specific Az Powershell modules? If its specific modules and **not** all Az Powershell, then Az.Security and Az.ResourceGraph should be included as well (how do I add them?)

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_MdcSubscriptionDefenderPlan/settings.json
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_MdcSubscriptionDefenderPlan/settings.json
@@ -1,0 +1,20 @@
+﻿{
+    "resourceName": "MdcSubscriptionDefenderPlan",
+    "description": "Enables or disables Microsoft Defender plans for a subscription in Microsoft Defender for Cloud.",
+    "roles": {
+        "read": [],
+        "update": []
+    },
+    "permissions": {
+        "graph": {
+            "delegated": {
+                "read": [],
+                "update": []
+            },
+            "application": {
+                "read": [],
+                "update": []
+            }
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Change log for Microsoft365DSC
 
 # UNRELEASED
-
+* MdcSubscriptionDefenderPlan
+  * Initial release
 * AADPasswordRuleSettings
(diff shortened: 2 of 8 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_MdcSubscriptionDefenderPlan/settings.json
```diff
@@ -0,0 +1,20 @@
...
+    "roles": {
+        "read": [],
+        "update": []
+    },
+    "permissions": {
+        "graph": {
+            "delegated": {
+                "read": [],
+                "update": []
+            },
```
chshum, line 13: do you need to add relevnat permissions here?

chshum (approved):
:shipit:

Modules/Microsoft365DSC/DSCResources/MSFT_MdcSubscriptionDefenderPlan/settings.json
MatanShabtay: Since we are calling Azure APIs, we need to specify Azure required permissions, which is currently not supported

NikCharlebois (approved):

PR #5087, opened 2024-09-24
Microsoft365 DSC for Intune MobileApps MacOS Lob Apps.

Kajalp1079:
* IntuneMobileAppsMacOSLobApp
  * Initial release

N/A


Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,580 @@
...
+            $instance = $Script:exportedInstances | Where-Object -FilterScript {$_.Id -eq $Id}
+        }
+        else
+        {
+            $instance = Get-IntuneMobileAppsMacOSLobApp -MobileAppId $Id -ErrorAction Stop
+        }
+        if ($null -eq $instance)
+        {
+            return $nullResult
+        }
```
FabienTschanz: There is no check for `DisplayName`, even though it is present in the parameters. For cross-tenant configuration, support for that is highly requested.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,580 @@
...
+        [System.String]
+        $Id,
+
+        [Parameter()]
+        [PSObject]
+        $BodyParameter,
+
+        [Parameter()]
+        [System.String]
+        $ResponseHeadersVariable,
```
FabienTschanz: Why do we export the response headers?

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,580 @@
...
+
+        [Parameter()]
+        [System.Collections.Hashtable]
+        $AdditionalProperties,
```
FabienTschanz: What are those additional properties? We normally don't want any hashtable but rather named properties for better readability.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
```diff
@@ -0,0 +1,32 @@
+[ClassVersion("1.0.0"), FriendlyName("IntuneMobileAppsMacOSLobApp")]
+class MSFT_IntuneMobileAppsMacOSLobApp : OMI_BaseResource
+{
+    [Key, Description("The unique identifier for an entity. Read-only. Inherited from mobileApp object.")] String Id;
+
+    [Write, Description("Optional Response Headers Variable.")] String ResponseHeadersVariable;
+    [Write, Description("Additional Parameters.")] Hashtable AdditionalProperties; #TODOK
+
+    [Write, Description("The description of the app. Inherited from mobileApp.")] String Description;
+    [Write, Description("Name of the device health script publisher")] String Developer;
```
FabienTschanz: Should probably not be `device health script publisher`, rather `The developer of the app.`

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
```diff
@@ -0,0 +1,32 @@
...
+    [Write, Description("The admin provided or imported title of the app. Inherited from mobileApp.")] String DisplayName;
+    [Write, Description("Name of the device health script publisher")] String InformationUrl;
```
FabienTschanz: Same as for `Developer`, incorrect text.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
```diff
@@ -0,0 +1,32 @@
...
+    [Write, Description("The value indicating whether the app is marked as featured by the admin. Inherited from mobileApp.")] String IsFeatured;
+    [Write, Description("The large icon to be displayed in the app details and used for upload of the icon. Inherited from mobileApp.")] String LargeIcon; #TODOK
```
FabienTschanz: `LargeIcon` is mentioned as `mimeContent`, which is (from my understanding) just a base64 encoded version of the bytes of the icon. We should probably do it that way too.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
```diff
@@ -0,0 +1,32 @@
...
+    [Write, Description("Notes for the app. Inherited from mobileApp.")] String Notes;
+    [Write, Description("The owner of the app. Inherited from mobileApp.")] String Owner;
+    [Write, Description("The privacy statement Url. Inherited from mobileApp.")] String PrivacyInformationUrl;
+    [Write, Description("The publisher of the app. Inherited from mobileApp.")] String Publisher;
+    [Write, Description("The publishing state for the app. The app cannot be assigned unless the app is published. Inherited from mobileApp. Possible values are: notPublished, processing, published.")] MobileAppPublishingState PublishingState; #TODOK
```
FabienTschanz: If there are only specific values mentioned, they must be used in a `ValueMap` and in the parameter definition as a `ValidateSet`. Although that property is most likely read-only...

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
```diff
@@ -0,0 +1,32 @@
...
+
+    [Write, Description("The list of group assignments for this mobile app. To construct, see NOTES section for ASSIGNMENTS properties and create a hash table.")] String Assignments[]; #TODOK
+    [Write, Description("The list of categories for this app. To construct, see NOTES section for CATEGORIES properties and create a hash table.")] String Categories[]; #TODOK
```
FabienTschanz: Could be exported as only a list of strings with the display names of the respective categories assigned to the app.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
FabienTschanz, line 1: Only had a quick look at the properties on the graph documentation, don't know how accurate that is to the information we can extract from the command line you used. Wouldn't it be better if we used the Graph version of it? If yes, then quite a few properties are missing.

https://learn.microsoft.com/en-us/graph/api/resources/intune-apps-macoslobapp?view=graph-rest-1.0

FabienTschanz (changes requested):
Thank you for the submission. But is this PR supposed to be finished? The way I'm reading it there are missing examples and many `TODOK` comments. If the PR is not ready, either mark it as Draft: or don't open a PR yet until the functionality is ready.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
Kajalp1079: Can you please point me to a good example of code that shows how to deal with similar MIME content?

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
Kajalp1079: Can you please share a link of code example about how to handle a list of strings while the type is enum?

Kajalp1079:
> Thank you for the submission. But is this PR supposed to be finished? The way I'm reading it there are missing examples and many `TODOK` comments. If the PR is not ready, either mark it as Draft: or don't open a PR yet until the functionality is ready.

Yes, that's correct, I am blocked as the UTs are not working locally. I mentioned it in more details in my other PR for Intune App categories. I will appreciate your reply there if possible. Thank you.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
FabienTschanz: We don't yet have something that similar, but something like the following would do it: 

```powershell
$mobileApp = Get-MgBetaDeviceAppManagementMobileApp -MobileAppId <ID>
$base64 = [System.Convert]::ToBase64String($mobileApp.LargeIcon.Value)
```

Now, I think that the `mimeContent` must be a separate CIM object specified in the MOF template. Then, in the get and export cmdlets, we need to convert that to the corresponding type. For that, you can take a look at how the assignments are implemented, these are always a separate CIM type that needs to be converted.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
FabienTschanz: I didn't mean the `PublishingState`, I meant the `Categories` property. Don't want any confusion there. Instead of saving the categories with its Id and the display name, I believe it's "better" for us users of the module to only have the display name exported, then looked up by the resource and handled accordingly.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,566 @@
...
+        [Parameter()]
+        [System.String]
+        $Id,
+
+        [Parameter()]
+        [PSObject]
+        $BodyParameter,
+
+        [Parameter()]
+        [PSObject[]]
```
FabienTschanz: These are not of type `PSObject`, but rather `Microsoft.Management.Infrastructure.CimInstance[]`.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,566 @@
...
+    $nullResult.Ensure = 'Absent'
+    try
+    {
+        if ($null -ne $Script:exportedInstances -and $Script:ExportMode)
+        {
+            $instance = $Script:exportedInstances | Where-Object -FilterScript {$_.Id -eq $Id}
+        }
+        else
+        {
+            $instance = Get-MgDeviceAppManagementMobileApp -MobileAppId $Id -ErrorAction Stop
```
FabienTschanz: I recommend using the Beta version of the command. The beta module is already included in the list of modules, but the normal one isn't. Conveniently, the Intune portal always uses the beta access too.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,566 @@
...
+        }
+
+        if ($null -eq $instance)
+        {
+            Write-Verbose -Message "No Mobile app with Id {$Id} was found. Search with DisplayName."
+            $policy = Get-MgBetaDeviceAppManagementiOSManagedAppProtection -Filter "displayName eq '$DisplayName'" -ErrorAction SilentlyContinue
```
FabienTschanz:
```suggestion
            $policy = Get-MgBetaDeviceAppManagementMobileApp -Filter "displayName eq '$DisplayName'" -ErrorAction SilentlyContinue
```

Note: The `LargeIcon` property is not fetched when searching by display name. It is only available if the call is done against a specific mobile app id.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
```diff
@@ -0,0 +1,566 @@
...
+    param
+    (
+        #region resource parameters
+        [Parameter()]
+        [System.String]
+        $Id,
+
+        [Parameter()]
+        [PSObject]
+        $BodyParameter,
```
FabienTschanz: `BodyParameter` is a parameter of the cmdlet but not a property of the object.

FabienTschanz (commented):
If you want to take a look at how the Intune portal does it, simply open the Browser Developer Tools using F12 (or another key that opens it) and look at the network calls being done when you change / add / remove something. This way, you can see how everything works together.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.psm1
Kajalp1079: Oh, thanks for pointing out that LargeIcon is fetched only with Id. Then should I even bother to include it in this PR? We are not supposed to use the 'Id' to be tenant agnostic, right?

FabienTschanz: @Kajalp1079 It should be included. Please don't misunderstand it when I say "only available when fetched with Id". That doesn't mean that you have to make the Id a key parameter or something like that, it simply means that the contents of `LargeIcon` are only available to you when you call `Get-MgBetaDeviceAppManagementMobileApp -MobileAppId <XYZ>`. If you don't call it specifically like this, e.g. with `-Filter "displayName eq '<XYZ'"`, then `LargeIcon` will be an empty object. 

That means the following: If there is no policy with the provided Id, look it up by display name. If there is a hit on it, then call the policy directly to retrieve the contents of `LargeIcon`, and continue with the execution flow.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
Kajalp1079: ok, I will look into this.
In that case, as a user does it make sense to have to multiple scripts (.pms1's, mof's, examples, tests etc ...) for different types of Apps like Android, Windows, iOS, MacOS etc.? I noticed that they use the 4 common cmdlets as below:
New-MgBetaDeviceAppManagementMobileApp, Get-MgBetaDeviceAppManagementMobileApp, Update-MgBetaDeviceAppManagementMobileApp, Remove-MgBetaDeviceAppManagementMobileApp

Given the same cmdlets, I don't see any reason to create multiple files for different app types but I would like to know your thoughts.

FabienTschanz: As it couldn't be said better: It depends 🥲 Depends on what you see fit. Do you think you can manage all the different app types with one single resource? With all types, e.g. MSI, LoB, Store etc? Then yes, you can. But you have to be aware of the nuances each resource might have, different properties, different fetches, content upload etc. If I'm honest with you, I don't think it's a good way to have everything in one single resource. What you can probably do is combine some things into the same resource. I'll let you decide how to handle it.

Kajalp1079: Ok, that makes sense that handling  different properties, different fetches, content upload etc. maybe different for different apps types. But I was wondering why the Intune Apps' PS cmdlets (New-MgBetaDeviceAppManagementMobileApp, Get-MgBetaDeviceAppManagementMobileApp, Update-MgBetaDeviceAppManagementMobileApp, Remove-MgBetaDeviceAppManagementMobileApp)  do not have those different properties mentioned in the docs here? https://learn.microsoft.com/en-us/powershell/module/microsoft.graph.beta.devices.corporatemanagement/new-mgbetadeviceappmanagementmobileapp?view=graph-powershell-beta

While I can see the different properties in the Graph API docs as below:

1. Windows (Win32) apps: https://learn.microsoft.com/en-us/graph/api/intune-apps-win32lobapp-create?view=graph-rest-1.0&tabs=http - like it has: applicableArchitectures, installCommandLine, minimumCpuSpeedInMHz and many other properties that other apps types do not have.
2. macOS app: https://learn.microsoft.com/en-us/graph/api/intune-apps-macoslobapp-create?view=graph-rest-1.0&tabs=http - it has: bundleId, md5HashChunkSize etc... that other app types do not have.

Do you know why the PS cmdlet supports only the common params? If want to handle the params/properties specific to app type then how to do it? I will need to use Invoke-MgGraphRequest instead of the PS cmdlets that I am currently using in this .psm1 script?

FabienTschanz: It's not that the PowerShell cmdlet "only" supports the common parameters. I rather think that in the `-BodyParameter` you could specify all properties necessary for the desired application type to be created (or with `-AdditionalProperties`). So I don't think that you would need to use `Invoke-MgGraphRequest` there, but I also didn't test it, so my answer is more a guess than a fact. 

If there are significant differences, it's by far easier to create separate resources for the app types and have all the necessary properties, including the `@odata.type` handling there, instead of including all properties of all types, then "dynamically" get the values, and if not all are provided, throw an error. That would be a horrendous user experience.

From my point of view, go with separate resources where necessary. If there are some that you might be able to combine, go for it. Otherwise, don't.

Kajalp1079: I tried to add just one additional property specific to MacOS Lob app 'IgnoreVersionDetection', it's a Boolean property.
But it's not getting exported! I keep getting the below error. I've pushed that changes in this PR. This issue is not happening for one of the main Boolean property called "IsFeatured", I kept them both side-by-side, but I am not sure what's causing this error. There are a few app type specific properties that can be added for MacOS Lob app as "-AdditionalProperties". And many more for Win32 apps and other Lob app types. So I tried a few ways but none of them worked to fetch the additional properties :( 

1. Get-MgBetaDeviceAppManagementMobileAppAsMacOSLobApp -MobileAppId "8d027f94-0682-431e-97c1-827d1879fa79"
2.  Get-MgBetaDeviceAppManagementMobileApp -Filter "isof('microsoft.graph.macOSLobApp')"  -ExpandProperty "categories,assignments"  -Select *
3.  Get-MgBetaDeviceAppManagementMobileApp -Filter "isof('microsoft.graph.macOSLobApp')"  -ExpandProperty "categories,assignments"  -Select "id,displayName,IgnoreVersionDetection"
4. 2.  Get-MgBetaDeviceAppManagementMobileApp -All -Filter "isof('microsoft.graph.macOSLobApp')"  -ExpandProperty "categories,assignments" | Where-Object -FilterScript { $_.AdditionalProperties.'@odata.type' -eq '#microsoft.graph.macOSLobApp' }

The below one is coming while trying to Export the config:

System.Management.Automation.ParameterBindingArgumentTransformationException: Cannot process argument transformation on parameter 'IgnoreVersionDetection'. Cannot convert value "" to type "System.Boolean". Boolean parameters accept only Boolean values and numbers, such as $True, $False, 1 or 0. ---> System.Management.Automation.ArgumentTransformationMetadataException: Cannot convert value "" to type "System.Boolean". Boolean parameters accept only Boolean values and numbers, such as $True, $False, 1 or 0. ---> System.Management.Automation.PSInvalidCastException: Cannot convert value "" to type "System.Boolean". Boolean parameters accept only Boolean values and numbers, such as $True, $False, 1 or 0.
   at 
"Error during Export:"
at Export-TargetResource, MSFT_IntuneMobileAppsMacOSLobApp.psm1: line 655
at Start-M365DSCConfigurationExtract, M365DSCReverse.psm1: line 682
at Export-M365DSCConfiguration, M365DSCUtil.psm1: line 1460
at <ScriptBlock>, <No file>: line 1

Kajalp1079: I tried with a string property as well, it doesn't throw any error but it doesn't get exported either.
Note: Since the Additional Properties are not working, currently the params that are added are common to all apps as they come from the base type graph entity class: MobileApp

FabienTschanz: @Kajalp1079 Accessing the properties inside of `AdditionalProperties` is in fact a case-sensitive matter. It doesn't work with `IgnoreVersionDetection`, but it does with `ignoreVersionDetection` (lower-case i). Please try it like that. You can do that from the command line as well, you don't have to always run the export method. It's much easier like this.

Kajalp1079: I have been trying from cmdline only, that's why I mentioned cmdlets above, let me paste them again as below. But the main thing that I am not sure about is the Get-MgBetaDeviceAppManagementMobileApp  cmdlet does not have -AdditionalProperty as switch. So when you say run the cmdlet with lower case you mean in -Select ?

Get-MgBetaDeviceAppManagementMobileAppAsMacOSLobApp -MobileAppId "8d027f94-0682-431e-97c1-827d1879fa79"
Get-MgBetaDeviceAppManagementMobileApp -Filter "isof('microsoft.graph.macOSLobApp')" -ExpandProperty "categories,assignments" -Select *
Get-MgBetaDeviceAppManagementMobileApp -Filter "isof('microsoft.graph.macOSLobApp')" -ExpandProperty "categories,assignments" -Select "id,displayName,ignoreVersionDetection"
Get-MgBetaDeviceAppManagementMobileApp -All -Filter "isof('microsoft.graph.macOSLobApp')" -ExpandProperty "categories,assignments" | Where-Object -FilterScript { $_.AdditionalProperties.'@odata.type' -eq '#microsoft.graph.macOSLobApp' }

FabienTschanz: @Kajalp1079 I was referring to accessing the property outside of the `-Select` operator, which is sent to the Graph API. Dictionaries can be case-sensitive, and `AdditionalProperties` is such a thing unfortunately. Hope that helps. 

```powershell
> $mobileApp = Get-MgbetaDeviceAppManagementMobileApp -Filter "isof('microsoft.graph.macOSLobApp')"
> $mobileApp.AdditionalProperties.IgnoreVersionDetection # Outputs nothing
> $mobileApp.AdditionalProperties.ignoreVersionDetection
True
```

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
Kajalp1079: So I exported the LargeIcon in the config file, and the M365Config.ps1 file compiled successfully as well. However the 
Start-dscconfiguration cmdlet is throwing an error due to very big size of the object. Do you know any work around? If not, then I guess, I will have to remove LargeIcon from this DSC, maybe that's why no other files contain similar example?


PS E:\3\a> Start-dscconfiguration  M365TenantConfig -Wait -verbose -force
VERBOSE: Perform operation 'Invoke CimMethod' with following parameters, ''methodName' = SendConfigurationApply,'className' = MSFT_DSCLocalConfigurationManager,'namespaceName' =
root/Microsoft/Windows/DesiredStateConfiguration'.
The WinRM client sent a request to the remote WS-Management service and was notified that the request size exceeded the configured MaxEnvelopeSize quota.
    + CategoryInfo          : LimitsExceeded: (root/Microsoft/...gurationManager:String) [], CimException
    + FullyQualifiedErrorId : HRESULT 0x80338111
    + PSComputerName        : localhost

VERBOSE: Operation 'Invoke CimMethod' complete.
VERBOSE: Time taken for configuration job to complete is 1.068 seconds

FabienTschanz: You have to increase the `maxEnvelopeSize` for WinRM to allow more content to be sent to it: 
```powershell
New-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\WSMAN\Client' -Name maxEnvelopeSize -Value 65536 -PropertyType DWORD -Force
```

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
Kajalp1079: Got it, it worked for a primitive type i.e ignoreVersionDetection boolean property.
Next, I picked an array of complex type (which has 3 strings) called childApps. That got exported fine (I think). 
but the   .\M365TenantConfig.ps1 is throwing error as below.  I am not sure how to debug the .\M365TenantConfig.ps1 , so I have pushed the exported file temporarily in the latest iteration. All childApps related changes are also pushed. Can you please take a look at it?


At E:\DSC\5\M365TenantConfig.ps1:17 char:9
+         IntuneMobileAppsMacOSLobApp "IntuneMobileAppsMacOSLobApp-Skyp ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~
Undefined DSC resource 'IntuneMobileAppsMacOSLobApp'. Use Import-DSCResource to import the resource.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : ResourceNotDefined

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneMobileAppsMacOSLobApp/MSFT_IntuneMobileAppsMacOSLobApp.schema.mof
Kajalp1079: ok, great, will try it, thank you.

FabienTschanz: @Kajalp1079 I received an email notification about an issue with M365TenantConfig.ps1 not running because it can't find the `IntuneMobileAppsMacOSLobApp` resource. For the file to compile, you have to copy the directory `MSFT_IntuneMobileAppsMacOSLobApp` to the Microsoft365DSC DSCResources directory under `C:\Program Files\WindowsPowerShell\Modules\Microsoft365DSC\1.24.1002.1\DSCResources`. If you are using another version, switch the version number accordingly. After that, it will compile.

FabienTschanz:
There are many issues with this... Currently working on fixes for all of them.

Kajalp1079:
> There are many issues with this... Currently working on fixes for all of them.

Hey @FabienTschanz - Thank you for the changes. I was wondering how do we ensure that the admin can provide the input of app's file package while creating an LoB app instance. I noticed that to create an app the UX asks to browse to an app pkg (multiple files). But in the API, I couldn't see any param/property that takes them as input.

FabienTschanz:
@Kajalp1079 There currently is no functionality for that. We might add a file path parameter to the resource which points to the file to upload, and if some check says that the file is not the same, we upload it again using a Graph function.

FabienTschanz:
@Kajalp1079 I know it's for a different resource, but for the Win32 app type, you would need to use the following cmdlets: 

* `New-MgBetaDeviceAppManagementMobileAppAsWin32LobAppContentVersion` --> Creating a new content version
* `Get-MgBetaDeviceAppManagementMobileAppAsWin32LobAppContentVersionFile` --> Get the reference for the new content version file
* `New-MgBetaDeviceAppManagementMobileAppAsWin32LobAppContentVersionFile` --> Add the manifest, size and name information
* `Get-MgBetaDeviceAppManagementMobileAppAsWin32LobAppContentVersionFile` with a specific file id --> Check if the Azure storage uri request was successful to put the content to

After that, I'm not really sure, but you might want to check out https://github.com/microsoftgraph/powershell-intune-samples/blob/master/LOB_Application/Application_LOB_Add.ps1 for additional information on how to upload the file chunks to the Azure storage uri.

Kajalp1079:
Yes, I also realized the other APIs that you mentioned will be needed, which will make the DSC script complex and I am not sure if it will be usable or not. So for now our team has decided to check-in the current changes and revisit it later.

________________________________

Kajalp1079:
There is another work around about Apple Push Notification related to enrolling an Apple device in Intune. There we need to upload a certificate, which we can via browsing on UX, but in the Intune API, I don't see the path param, so I am wondering if such DSC script will be even usable or not?
________________________________

PR #5137, opened 2024-10-03
Added AADAuthenticationRequirements resource

ritikmit:
Added AADAuthenticationRequirements resource for Per-User MFA Requirements monitoring

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationRequirement/MSFT_AADAuthenticationRequirement.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationRequirement/MSFT_AADAuthenticationRequirement.psm1
@@ -1,0 +1,381 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter()]
+        [ValidateSet('enabled', 'disabled')]
+        [System.String]
+        $PerUserMfaState,
+
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $Id,
+
+        [Parameter()]
+        [System.Management.Automation.PSCredential]
+        $Credential,
+
+        [Parameter()]
+        [System.String]
+        $ApplicationId,
+
+        [Parameter()]
+        [System.String]
+        $TenantId,
+
+        [Parameter()]
+        [System.Management.Automation.PSCredential]
+        $ApplicationSecret,
+
+        [Parameter()]
+        [System.String]
+        $CertificateThumbprint,
+
+        [Parameter()]
+        [Switch]
+        $ManagedIdentity,
+
+        [Parameter()]
+        [System.String[]]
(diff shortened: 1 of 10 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationRequirement/MSFT_AADAuthenticationRequirement.psm1
```diff
@@ -0,0 +1,381 @@
...
+    param
+    (
+        [Parameter()]
+        [ValidateSet('enabled', 'disabled')]
+        [System.String]
+        $PerUserMfaState,
+
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $Id,
```
NikCharlebois: Id will be unique to the environment, instead, can we make UserPrincipalName (without the @domain) or some other property the Key so that it can be deployed to multiple tenants? I understand the API call will need the ID, but we should first retrieve the user based on that key (e.g., UPN) and then get the id from there for the call to the API

Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationRequirement/MSFT_AADAuthenticationRequirement.schema.mof
```diff
@@ -0,0 +1,14 @@
+[ClassVersion("1.0.0.0"), FriendlyName("AADAuthenticationRequirement")]
+class MSFT_AADAuthenticationRequirement : OMI_BaseResource
+{
+    [Write, Description("The state of the MFA enablement for the user. Possible values are: enabled, disabled."), ValueMap{"enabled","disabled"}, Values{"enabled","disabled"}] String PerUserMfaState;
+    [Key, Description("The unique identifier for an entity. Read-only.")] String Id;
```
NikCharlebois: Again, we should make User Name or something similar be the resource's key.

Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationRequirement/settings.json
```diff
@@ -0,0 +1,20 @@
...
+        "update": []
+    },
+    "permissions": {
+        "graph": {
+            "delegated": {
+                "read": [],
+                "update": []
+            },
+            "application": {
+                "read": [],
```
NikCharlebois: We need to update this with the actual permissions.

NikCharlebois (changes requested):

Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationRequirement/MSFT_AADAuthenticationRequirement.psm1
ritikmit: API can use UPN as well. Changed ID to UPN everywhere

NikCharlebois (approved):

PR #5194, opened 2024-10-16
AAD Access Review Definition

kasaxena5:


NikCharlebois (changes requested):
Missing an entry in the changelog
Missing at least one example for the resource.

NikCharlebois (approved):

PR #5204, opened 2024-10-18
Add IntuneDeviceManagementSettings resource

ericoevering:
I had the need to get the Device Management Settings from Intune to get,test and set the securebydefault value for the compliancy settings.

The (Get-MgBetaDeviceManagement).settings command is responding with empty values, when using the api it does give the correct values 

Invoke-MgGraphRequest -Uri "https://graph.microsoft.com/beta/deviceManagement/settings" -Method GET -ErrorAction SilentlyContinue

this is my first time making a resource so maybe i could use some pointers to write the script cleaner but it is working.

None


ricmestre:
@ericoevering Haven't you tried the resource *IntuneDeviceManagementComplianceSettings*? It has the property you need.

ericoevering:
@ricmestre  i was using https://export.microsoft365dsc.com/ for the resources feeling little stupid now..... thank you for the pointer i will close this

PR #5232, opened 2024-10-21
AAD/Intune: Fix permissions in settings.json

ricmestre:

*AADCustomSecurityAttributeDefinition* and *AADIdentityGovernanceProgram* were missing the delegated permissions, this PR fixes this but the bulk of the change was to add *Group.Read.All* to almost all Intune resources, where assignments are used, this has been like this since ever so I always have to add it manually to every customer and they keep me asking about this additional permission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,14 @@
 
 * AADAccessReviewDefinition
   * Initial release.
+* AADCustomSecurityAttributeDefinition
+  * Fixed missing permissions in settings.json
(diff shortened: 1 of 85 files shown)


ricmestre:
@fabientschanz The DRG might need to be updated to include *Group.Read.All* for Intune resources by default, not all of them might use it but the majority does and it's easier to remove manually from those who doesn't.

What do you think?

NikCharlebois (approved):

PR #5254, opened 2024-10-23
Fix autofromconnector type for Intune EDR Windows10 policy

### LINKED ISSUE #5246: IntuneEndpointDetectionAndResponsePolicyWindows10: Throws exception if blob is empty
### Description of the issue

@FabienTschanz Your commit https://github.com/microsoft/Microsoft365DSC/commit/b48abe086a02e953cf54ed172355768504f60a0b is now causing an exception if blob is empty, we have a policy in our tenant where the configuration type is not set and therefore the blob is empty as well, so if I try to export and then apply that to another tenant it causes the exception [0] to be thrown.

It's arguable if you should be allowed to have a policy without configuration type, the thing is that is still possible through Intune admin portal, but even if you don't agree that should be possible, see [1] where this is not mentioned anywhere so I kind of agree, BUT you must not throw an exception if you set the type to auto from connector which never has a blob assigned so the condition must be changed to factor this in.

Additionally you need to change *$($DSCParams.ConfigurationType)* in the exception line since *$DSCParams* is empty.

[0] https://github.com/microsoft/Microsoft365DSC/blame/76cbfa8d2be39fbe6cef5cd5e69f75d84b077a17/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10/MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10.psm1#L296
[1] https://learn.microsoft.com/en-us/mem/intune/protect/endpoint-security-edr-policy

### Microsoft 365 DSC Version

1.24.1016.1

Intune

### The DSC configuration

        IntuneEndpointDetectionAndResponsePolicyWindows10 "IntuneEndpointDetectionAndResponsePolicyWindows10-IntuneEndpointDetectionAndResponsePolicyWindows10_1"
        {
            Assignments          = @();
            ConfigurationType    = "autofromconnector";
            Description          = "";
            DisplayName          = "IntuneEndpointDetectionAndResponsePolicyWindows10_1";
            Identity             = "83e155d5-f1f3-4ef9-9027-16cc24b22ae8";
            RoleScopeTagIds      = @("0");
            SampleSharing        = "1";
        }

PowerShell DSC resource MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10  failed to execute Set-TargetResource functionality with error message: ConfigurationBlob is  required for configurationType ''
+ CategoryInfo          : InvalidOperation: (:) [], CimException
+ FullyQualifiedErrorId : ProviderOperationExecutionFailure
+ PSComputerName        : localhost

Win11/PS5.1

FabienTschanz:
This PR aims to address an issue with the `AutoFromConnector` value for the `ConfigurationPackageType` in the `IntuneEndpointDetectionAndResponsePolicyWindows10` resource. It throws an error for the empty configuration blob, although that blob must be empty because it will be populated by the Graph endpoint. 

- Fixes #5246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
   * Initial release.
 * AADSocialIdentityProvider
   * Fixed missing permissions in settings.json
+* IntuneEndpointDetectionAndResponsePolicyWindows10
+  * Fixes an issue with `AutoFromConnector` as the Configuration package type.
+    FIXES [#5246](https://github.com/microsoft/Microsoft365DSC/issues/5246)
 * Intune workload
   * Fixed missing permissions in settings.json
 * SentinelAlertRule
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10/MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10/MSFT_IntuneEndpointDetectionAndResponsePolicyWindows10.psm1
@@ -286,7 +286,7 @@ function Set-TargetResource
 
     if ($ConfigurationType -ne 'AutoFromConnector' -and [System.String]::IsNullOrEmpty($ConfigurationBlob))
     {
-        throw "ConfigurationBlob is required for configurationType '$($DSCParams.ConfigurationType)'"
+        throw "ConfigurationBlob is required for configurationType '$($ConfigurationType)'"
     }
     $BoundParameters.Remove('ConfigurationType') | Out-Null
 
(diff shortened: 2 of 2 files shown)


ricmestre:
Sure, let me have a look and I'll let you know.

ricmestre:
No, unfortunately it still fails with:

"device_vendor_msft_windowsadvancedthreatprotection_configurationtype: Selected option                                                                          device_vendor_msft_windowsadvancedthreatprotection_configurationtype_autofromconnector with dependent settings  doesnt contain required dependent settings. Required         dependent settings are device_vendor_msft_windowsadvancedthreatprotection_onboarding_fromconnector"

FabienTschanz:
Thank you for the feedback, definitely missed the correct property name. Can you try with the latest change? I hope that works now.

ricmestre:
It sure does :rocket:! Bad news is that now I can't test this resource in my dev tenant since the Defender connector can't be setup, oh well life...

Please just don't forget to change the the throw exception you added, use *$ConfigurationType* instead of *$DSCParams.ConfigurationType*

FabienTschanz:
> It sure does 🚀! Bad news is that now I can't test this resource in my dev tenant since the Defender connector can't be setup, oh well life...
> 
> Please just don't forget to change the the throw exception you added, use _$ConfigurationType_ instead of _$DSCParams.ConfigurationType_

Silly me, always forgetting stuff you just mentioned a minute ago... Added it 👍

NikCharlebois (approved):

PR #5255, opened 2024-10-23
Update checks for Azure Automation

jackpoz:
Make Azure Automation checks more reliable due to updated values from Azure for $env:AZUREPS_HOST_ENVIRONMENT. The changes are the same applied at https://github.com/microsoft/MSCloudLoginAssistant/pull/175 and https://github.com/microsoft/MSCloudLoginAssistant/pull/183 in MSCloudLoginAssistant .

Might have to do the same on every other project that has the same check, as it seems it was copypasted around.

None.

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
@@ -691,7 +691,7 @@ function Start-M365DSCConfigurationExtract
 
         # Azure Automation Check
         $AzureAutomation = $false
-        if ('AzureAutomation/' -eq $env:AZUREPS_HOST_ENVIRONMENT)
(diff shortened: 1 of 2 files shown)


NikCharlebois:
@andikrueger any chance you can do the code review on this one. You drove the initial logic for AzAutomation and I would appreciate your input on this.

FabienTschanz:
@NikCharlebois The changes look correct, other folks created a list on the possible values and where they are running from: https://rakhesh.com/azure/azure-automation-powershell-variables. Since the changes are also implemented in the MSCloudLoginAssistant, it's fair to assume it's all good (also in my opinion).

andikrueger:
I don't have an environment right now to run some test. The code itself looks good!

NikCharlebois (approved):

PR #5258, opened 2024-10-23
IntuneAppCategory: Fix retrieval of resource

ricmestre:

This PR calls the correct cmdlet when searching by DisplayName, if the resource cannot be retrieved by Id, since the result could be null it would always create new categories with the same name.

Additionally since there was no feedback of what the code was doing I also added a few verbose messages and while here also changed *Test-TargetResource* a bit to resemble other resources.


FabienTschanz:
LGTM 🚀

NikCharlebois (approved):

NikCharlebois (approved):

PR #5259, opened 2024-10-23
IntuneDerivedCredential: Fix export and deployment

ricmestre:

This PR fixes the retrieval/deployment of the resource if it cannot be found by *Id* by silently continuing as done in other resources.

It also fixes the export and deployment if *NotificationType* has more than one option select, which in this case is always *companyPortal,email*

While here also changed *Test-TargetResource* function to resemble how it's done in several other resources.


FabienTschanz:
LGTM 🚀

NikCharlebois (approved):

PR #5271, opened 2024-10-25
Enroll Android Google Play

khmic5:


NikCharlebois (changes requested):
Missing an entry in the changelog.

FabienTschanz:
@khmic5 It's correct that there is no "update" of the configuration, right? It's either Create or Delete? If yes, then it's best to also only check that in the Test-TargetResource. Otherwise, I can simply change any property, e.g. `OwnerOrganizationName`, and this will result in a drift and try to apply the setting afterwards, which wouldn't do anything (because no update is possible), resulting in another drift, and so on. This would never be fixed. Is it possible to update the Test-TargetResource in that way?

NikCharlebois:
> @khmic5 It's correct that there is no "update" of the configuration, right? It's either Create or Delete? If yes, then it's best to also only check that in the Test-TargetResource. Otherwise, I can simply change any property, e.g. `OwnerOrganizationName`, and this will result in a drift and try to apply the setting afterwards, which wouldn't do anything (because no update is possible), resulting in another drift, and so on. This would never be fixed. Is it possible to update the Test-TargetResource in that way?

Yes, that is correct, we need to remove and recreate instead of creating. I think the example you give above is correct to always throw drifts, however, if it is the case, the logic flow in the Set should first attempt to remove the existing instance, then recreate it. Right now the only way to delete is to specify Ensure=Absent

FabienTschanz:
@NikCharlebois Thank you for the clarification. Then we must also make sure that only properties are in the manifest that have an actual impact on the behavior of the connector, since removing and recreating the enrollment for a property that doesn't have an impact is kinda useless. I trust @khmic5 to select the correct properties there though, not going to interfere. Just double-checking if my understanding aligns with the intentions 😃

khmic5:
@FabienTschanz , @NikCharlebois thank you for the comments on noticing this - I have removed the parameters that cannot be set for now since the API for that can only be called from within an Intune portal. Once the product team completes the changes needed, we will uncomment and allow for bind to happen as well

NikCharlebois (approved):

PR #5276, opened 2024-10-27
Improve settings catalog experience

FabienTschanz:
This PR reduces the complexity of the Intune Settings Catalog handling by using the same method to generate the property names of the resources when converting back from a DSC configuration. The shared commands from the resource generator as well as `M365DSCDRGUtil` are moved to a separate module named `M365DSCIntuneSettingsCatalogUtil.psm1`, which is imported on demand. 

@ricmestre Is it possible for you to run this against your lab infrastructure? The couple of resources I tested seemed fine, but I don't always have accurate examples. 

None.


ricmestre:
@FabienTschanz Looks like unit tests are complaining about something, but just to let you know that I won't be able to test this in the next few days since I'm on holidays without access to my job pc.

I'll let you know once I'm back to make the tests, but I'd like to see how improved for instance *IntuneSecurityBaselineMicrosoft365AppsForEnterprise* is, this one in particular takes about 15min inside the loops called by *Get-IntuneSettingCatalogPolicySetting* and I tested it successfully manually but since it took so long for each step I had to take it out from the automated testing.

FabienTschanz:
@ricmestre Enjoy, don't worry about Microsoft365DSC 😄 We got you covered. Now it's looking all good, I also checked with `IntuneSecurityBaselineMicrosoft365AppsForEnterprise`, and this time the execution time is down to about ~10 seconds for a test, and ~25 seconds for test and update. Massive improvement to the previous version. That is because I removed all iterations over the DSC params and went straight for the same property generation used when exporting. So in theory, it should only be slower in the difference of configured vs available settings with a bit of overhead for the body generation.

NikCharlebois (approved):

PR #5326, opened 2024-11-04
IntuneDeviceManagementComplianceManagementPartner initial release

salbeck-sit:
Implement support for IntuneDeviceManagementComplianceManagementPartner

Note: Property 'lastHeartbeatDateTime' looks a little read-only but documentation appears to allow updating it

none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,8 @@
   * Added a few verbose messages
 * IntuneDeviceManagmentAndroidDeviceOwnerEnrollmentProfile
   * Initial release.
+* IntuneDeviceManagementComplianceManagementPartner
+  * Initial release
 * IntuneEndpointDetectionAndResponsePolicyWindows10
   * Fixes an issue with `AutoFromConnector` as the Configuration package type.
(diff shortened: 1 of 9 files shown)


FabienTschanz:
@salbeck-sit If the properties look like read-only, then they probably shouldn't be included in the DSC configuration. What about the onboarded values for Android, MacOS etc.? Are those settings that we can actually set? In my quick check, I only saw that we can change the assignments and name, but nothing else. I think it's best if we keep ourselves as close to the portal options as possible to not confuse the users.

ricmestre:
Regarding the assignments do we really need a sub-CIM which only contains a single property? A single CIM for assignments like all other Intune resources should be enough, exception was done to *IntuneDeviceRemediation* because that one actually needs it.

The bigger problem is that without the *groupDisplayName* of the group this won't work on other target tenants, the same can be said about the *deviceAndAppManagementAssignmentFilterId* property on all Intune resources since there's no corresponding displayName for it but that's a fight for another day.

salbeck-sit:
Hmm. After using DRG to create the resource I did a quick check of parameters. I did notice that there are several 'xxxId's in there.
I agree that as it stands, it needs more work. I'll cancel this PR and come back with a better one

PR #5333, opened 2024-11-05
[draft] IntuneDeviceManagementComplianceManagementPartner

salbeck-sit:
Initial release (take two).

CimInstances from DRG-generated resource has been 'flattened' to one level and property GroupDisplayName has been added to ensure cross-tenant capability.
I have omitted property 'lastHeartbeatDateTime' although it appears possible to modify it.

Not too sure about property 'CollectionId' in the ...AssignmentTarget CimInstance. Advice is welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change log for Microsoft365DSC
 
+# UNRELEASED
+
 # 1.24.1127.1
 
 * AAD


salbeck-sit:
Posted an issue with regards to ConvertTo-IntunePolicyASsignment silently dropping assignments for groups that can't be read - this includes unit-tests failing in spite of required mocks. The easy workaround was to omit datatype *GroupAssignmentTarget from unit-tests. However in cross-tenant scenarios it's not implausible to reference a group that might not exist and in such a situation the resource should emit some sort of warning

salbeck-sit:
updated resource to write event if a group is unknown - when getting assignments that refer to an unkown group, a warning is written, when setting assignments that refer to an unknown group, an error is written. I'm still undecided if that should cause an exception to avoid thinking that a configuration was updated if it really wasn't.

PR #5356, opened 2024-11-07
M365 data at rest encryption policy

kasaxena5:
M365 data at rest encryption policy


Modules/Microsoft365DSC/DSCResources/MSFT_M365DataAtRestEncryptionPolicy/MSFT_M365DataAtRestEncryptionPolicy.psm1
```diff
@@ -0,0 +1,395 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $Identity,
```
FabienTschanz, line 9: Is the identity the key parameter here or can we use the `Name` property instead?

kasaxena5: Identity is human friendly, we can use identity as key.

NikCharlebois (approved):

ricmestre:
This is part of EXO workload, it should have been called EXO not M365.

Issue #5362, opened 2024-11-08
Issue with groups referenced by ConvertTo- and ConvertFrom-IntunePolicyAssignment in M365DRGUtil.psm1


salbeck-sit:
### Description of the issue

While refactoring a resource that utilizes Intune assignments I came across two functions in the M365DSCDRGUtil.psm1 module that proved very useful in flattening the CIM-structure of the arguments: ConvertTo- and ConvertFrom-IntunePolicyAssignment.

However, it appears that ConvertTo-IntunePolicyAssignment will silently drop assignments that reference a group that for any reason cannot be read.
I saw during my work to get unit-tests fail-free that even though I've mocked Get-MgGroup, the functions will get a 'one or more errors occurred'.
ConvertFrom-IntunePolicyAssignment will include the groupId in the output but of course won't include groupDisplayName if the group cannot be read -
BUT ConvertTo-IntunePolicyAssignment will drop the assignment altogether - in verbose-mode a message will be issued that 'Skipping assignment for the group with DisplayName *** as it could not be read, Please update...'
I would at least expect a message in the eventlog or even throwing an exception. How is this handled in other resources ?

### Microsoft 365 DSC Version

DEV

### Which workloads are affected

Intune

salbeck-sit:
It appears that the problem with mocking that is indirectly causing the issue is a known error in WPS 5.1 as defined here: https://pester.dev/docs/usage/mocking#mocking-a-function-that-is-called-by-a-method-in-a-powershell-class
I'm not sure though how to implement the workaround in the unit-test

Anyway - the issue still stands. If an unknown groupDisplayName is specified with datatype #microsoft.graph.groupAssignmentTarget then the assignment is silently dropped.

FabienTschanz:
@salbeck-sit That's unfortunately a restriction we have to live with. For your resource, you can use one of the other mocked assignments we are using in other resources (e.g. `IntuneFirewallRulesPolicyWindows10`). 

```powershell
Mock -CommandName Get-MgBetaDeviceManagementConfigurationPolicyAssignment -MockWith {
    return @(@{
        Id       = '12345-12345-12345-12345-12345'
        Source   = 'direct'
        SourceId = '12345-12345-12345-12345-12345'
        Target   = @{
            DeviceAndAppManagementAssignmentFilterId   = '12345-12345-12345-12345-12345'
            DeviceAndAppManagementAssignmentFilterType = 'none'
            AdditionalProperties                       = @(
[...]

PR #5371, opened 2024-11-08
Switch log type to Warning for not found Intune assignments

FabienTschanz:
This PR aims to provide improved visibility for not found Intune assignments when assigning to an Intune policy assignment which uses display name when the id is not found. This was mentioned in #5362. 

None.


ricmestre:
Erroring out if groups could not be found would be a huge pain for me, this looks much better 👍

FabienTschanz:
> Erroring out if groups could not be found would be a huge pain for me, this looks much better 👍

Figured the same, an error should only be thrown if execution can't be continued at any cost. Skipping a group is much preferred.

salbeck-sit:
If the calling resource needs to know, it can compare number of returned assignment-targets with number of assignments and log an event and/or throw an exception if there is a mismatch. That's fine with me.

NikCharlebois (approved):

PR #5385, opened 2024-11-11
IntuneDeviceConfigurationCustomPolicyWindows10: Cast Value to String

### LINKED ISSUE #5384: IntuneDeviceConfigurationCustomPolicyWindows10: Cannot compare OmaSettings which contain a boolean Value of $False
### Description of the issue

This was working before and some change somewhere affected this resource, the deployment is done correctly without issues and the settings are correct but *Compare-M365DSCComplexObject* reports that the resource is not in desired stated whenever there's at least 1 OmaSettings defined where it's a *odataType = '#microsoft.graph.omaSettingBoolean'* and *Value = $False*, if *Value* is set to *$True* then the comparison works without issues. The comparison also works without issues with any other *odataType* such as string and integer.

### Microsoft 365 DSC Version

1.24.1106.3

Intune

### The DSC configuration

```powershell
IntuneDeviceConfigurationCustomPolicyWindows10 "IntuneDeviceConfigurationCustomPolicyWindows10-IntuneDeviceConfigurationCustomPolicyWindows10_5"
        {
            Assignments          = @(
                    groupDisplayName = 'DummyGroupExclude'
                }
                    groupDisplayName = 'DummyGroupInclude'
                }
            );
            DisplayName          = "IntuneDeviceConfigurationCustomPolicyWindows10_5";
            Id                   = "3c1db9b7-ba18-4152-85ac-03909559087a";
            OmaSettings          = @(
                MSFT_MicrosoftGraphomaSetting{
                    Description = 'AllowLocalPolicyMerge'
                    OmaUri = './Device/Vendor/MSFT/Firewall/MdmStore/PublicProfile/AllowLocalPolicyMerge'
                    Value = $False
                    odataType = '#microsoft.graph.omaSettingBoolean'
                    IsEncrypted = $False
                    DisplayName = 'AllowLocalPolicyMerge'
                }
            );
            SupportsScopeTags    = $True;
        }
```

```powershell
N/A
```

```powershell
Win11/PS5.1
```

ricmestre:

Not sure why and at what point this stopped working but if at least 1 OmaSettings from the resource being deployed/tested contains an *odataType = '#microsoft.graph.omaSettingBoolean'* and *Value = $False* then the comparison will fail and will report that the resource is not in desired state.

After some troubleshooting I found the issue, Graph always stores *Value* as string, on the other hand DSC will export/deploy/test *Value* as boolean (if of course it's of that type) so when comparing *$true -eq "true"* returns true but on the other hand *$false -eq "false"* returns false.

Short story, this can be fixed by ensuring that *Value* is always casted as *String*, at the same time while here I've also removed a verbose message, this resource might contain up to 255 OmaSettings, this message would always spam the console a few times showing all of them one by one while handling the resource which also adds up in terms of how long it takes to process the resource.

- Fixes #5384


FabienTschanz:
This is a very mean one... Classic "object" type comparison vs concrete types. LGTM 🚀

salbeck-sit:
@ricmestre While I'm not familiar with the long story, why not use [System.Boolean]::TryParse($value, [ref]$result) - it returns true if $value has a string value of 'true' or 'false' and updates $result with $value converted to boolean. In this way, 'true' is converted to $True and 'false' is converted to $False.

ricmestre:
The value is already boolean, what it needs here to fix the issue is to convert it to string.

PR #5407, opened 2024-11-14
Add Intune Firewall rules policy for Windows10 ConfigMgr

FabienTschanz:
This PR adds the `IntuneFirewallRulesPolicyWindows10ConfigMgr` resource. This is the equivalent of the `IntuneFirewallRulesPolicyWindows10`, but for the configuration manager aka SCCM. 

@NikCharlebois Can you give some feedback to the responsible product group that there is no way to modify an existing policy because the UI is not correctly working? It's showing the design of an /intents policy instead of a /configurationPolicies policy. The `Configuration settings` section stays empty, even though there are settings configured. 

image

None.

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr.psm1
@@ -1,0 +1,641 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        #region resource generator code
+        [Parameter()]
+        [System.String]
+        $Description,
+
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $DisplayName,
+
+        [Parameter()]
+        [System.String[]]
+        $RoleScopeTagIds,
+
+        [Parameter()]
+        [System.String]
+        $Id,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $FirewallRuleName,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $Assignments,
+        #endregion
+
+        [Parameter()]
+        [System.String]
+        [ValidateSet('Absent', 'Present')]
+        $Ensure = 'Present',
+
+        [Parameter()]
+        [System.Management.Automation.PSCredential]
+        $Credential,
+
+        [Parameter()]
+        [System.String]
+        $ApplicationId,
+
+        [Parameter()]
+        [System.String]
+        $TenantId,
+
+        [Parameter()]
+        [System.Management.Automation.PSCredential]
+        $ApplicationSecret,
+
+        [Parameter()]
+        [System.String]
+        $CertificateThumbprint,
+
+        [Parameter()]
+        [Switch]
+        $ManagedIdentity,
+
+        [Parameter()]
+        [System.String[]]
+        $AccessTokens
+    )
+
+    try
+    {
+        $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftGraph' `
+            -InboundParameters $PSBoundParameters
+
+        #Ensure the proper dependencies are installed in the current environment.
+        Confirm-M365DSCDependencies
+
+        #region Telemetry
+        $ResourceName = $MyInvocation.MyCommand.ModuleName.Replace('MSFT_', '')
+        $CommandName = $MyInvocation.MyCommand
+        $data = Format-M365DSCTelemetryParameters -ResourceName $ResourceName `
+            -CommandName $CommandName `
+            -Parameters $PSBoundParameters
+        Add-M365DSCTelemetryEvent -Data $data
+        #endregion
+
+        $nullResult = $PSBoundParameters
+        $nullResult.Ensure = 'Absent'
+
+        $getValue = $null
+        #region resource generator code
+        $getValue = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Id  -ErrorAction SilentlyContinue
+
+        if ($null -eq $getValue)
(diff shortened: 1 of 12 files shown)


Modules/Microsoft365DSC/Examples/Resources/IntuneFirewallRulesPolicyWindows10ConfigMgr/3-Remove.ps1
```diff
@@ -0,0 +1,34 @@
+<#
+This example removes a Device Control Policy.
```
ykuijs: This description looks incorrect

Modules/Microsoft365DSC/Examples/Resources/IntuneFirewallRulesPolicyWindows10ConfigMgr/2-Update.ps1
```diff
@@ -0,0 +1,56 @@
+<#
+This example updates a Intune Firewall Policy for Windows10.
```
ykuijs: Does it make sense to add "via Configuration Manager" at the end of the line?

Modules/Microsoft365DSC/Examples/Resources/IntuneFirewallRulesPolicyWindows10ConfigMgr/1-Create.ps1
```diff
@@ -0,0 +1,56 @@
+<#
+This example creates a new Intune Firewall Policy for Windows10.
```
ykuijs: Does it make sense to add "via Configuration Manager" at the end of the line?

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr.psm1
```diff
@@ -0,0 +1,641 @@
...
+        [Parameter()]
+        [Switch]
+        $ManagedIdentity,
+
+        [Parameter()]
+        [System.String[]]
+        $AccessTokens
+    )
+
+    try
```
ykuijs, line 69: Please add a Write-Verbose line that says this is the beginning of the Get method, like is done here:
https://github.com/microsoft/Microsoft365DSC/blob/ba9e95d9942dc58cea783af8d4db89ff31ec6a5c/Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationMethodPolicyAuthenticator/MSFT_AADAuthenticationMethodPolicyAuthenticator.psm1#L68

That way it is easy to track what function the code is in.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr.psm1
```diff
@@ -0,0 +1,641 @@
...
+        [Parameter()]
+        [Switch]
+        $ManagedIdentity,
+
+        [Parameter()]
+        [System.String[]]
+        $AccessTokens
+    )
+
+    #Ensure the proper dependencies are installed in the current environment.
```
ykuijs, line 247: Same comment as for the Get method:
https://github.com/microsoft/Microsoft365DSC/blob/ba9e95d9942dc58cea783af8d4db89ff31ec6a5c/Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationMethodPolicyAuthenticator/MSFT_AADAuthenticationMethodPolicyAuthenticator.psm1#L566

ykuijs (changes requested):
Just a few comments.

Would it be possible to fix these before 14:00h CET (GMT+1) tomorrow? After that I will start working on the new release and it would be great if we could include this resource. No worries, if that is not possible.....not trying to rush you 😉

ykuijs:
Also, please merge with Dev

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr.psm1
FabienTschanz: I'm using the default template of the resource generator and there is no line like that in pretty much all of the Intune resources. Do you still want me to add it?
Edit: What do we do with the other Intune resources? Updating them as well?

Modules/Microsoft365DSC/Examples/Resources/IntuneFirewallRulesPolicyWindows10ConfigMgr/3-Remove.ps1
FabienTschanz: It sure is... That's what I get from copy pasting 😆

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr/MSFT_IntuneFirewallRulesPolicyWindows10ConfigMgr.psm1
ykuijs: That is good to know. We need to update the DRG as well. Please update this resource.

It is on my ToDo list to go through all resources to make sure they all have these basics. So if you run into instances, please correct.....never hurts to have a little help checking and updating so many resources 😉

FabienTschanz: Will do 👍

FabienTschanz:
@ykuijs The resource is now updated with the Get- and Set- message as requested. An entry in the resource generator was also added, so each new resource created from that will include the `Write-Verbose` messages with their primary and filter key.

ykuijs (approved):
LGTM

PR #5408, opened 2024-11-14
AAD - Added ApplicationSecret as authentication method for multiple resources

jdlagace:


jdlagace:
Mismatch in folders. Will fix it and create another PR.

PR #5463, opened 2024-11-25
Add separate check for strings

FabienTschanz:
This PR improves the comparison of strings inside of `Compare-M365DSCComplexObject`. Previously, the comparison did not check for different line breaks (`\r\n` vs `\n`). The comparison was also updated to use `StringComparison.Ordinal`, a byte-wise check which does not take into consideration the current culture. 

None.

#### Task list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * IntuneTrustedRootCertificateIOS
   * Initial release
+* M365DSCDRGUtil
+  * Add separate check for strings with ordinal comparison and standardized line breaks.
 
 # 1.24.1127.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   * Initial release
 * M365DSCDRGUtil
   * Add separate check for strings with ordinal comparison and standardized line breaks.
+* M365DSCUtil
+  * Add separate check for strings with ordinal comparison and standardized line breaks.
 
 # 1.24.1127.1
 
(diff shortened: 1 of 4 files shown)


ricmestre:
Nitpicking, but comment *# Remove carriage return and line feed characters* is incorrect, you're replacing CRLF characters with LF so you could say you're only removing carriage return. Also not that it makes a difference but everything in M365DSC is done with CRLF in mind but you're actually replacing it with LF, again not that it makes a difference in the end result but just saying.

Also, this is something that is missing from plain string comparisons not inside CIM instances, I have a monitoring tool which I have to massage the results with something very similar so that they're not reported as drifts, can't just remember which function it is but it should be in *M365DSCUtil.psm1*

FabienTschanz:
@ricmestre Will update my comment 👀  Is the function you mean `Test-M365DSCParameterState`?

ykuijs (approved):
LGTM

ykuijs:
This PR looks fine with me. Are there any open actions based on the comment from @ricmestre?? Or can this PR be merged?

FabienTschanz:
@ykuijs The logic was also added in `Test-M365DSCParameterState`, which I assume was the function @ricmestre was talking about since it essentially contains the same logic as `Compare-M365DSCComplexObject`.

ricmestre:
Wasn't able to test it but it looks good to me.

ykuijs:
@FabienTschanz Could you please update from Dev and make sure your changelog entry is placed under the Unreleased section?

FabienTschanz:
> @FabienTschanz Could you please update from Dev and make sure your changelog entry is placed under the Unreleased section?

Will do, just a couple of minutes.

PR #5456, opened 2024-11-22
Fix limited Microsoft Graph results on Get

FabienTschanz:
#### Pull Request (PR) description
This PR fixes the issue where filtering by display name in Microsoft Graph did not return the policy when there are more than 25 policies of the same type. Filtering seems to occur after fetching the first 25 policies from the database, leading to no policy found, even though it did exist. 

#### This Pull Request (PR) fixes the following issues
None.

#### Task list

- [x] Added an entry to the change log under the Unreleased section of the file CHANGELOG.md.
      Entry should say what was changed and how that affects users (if applicable), and
      reference the issue being resolved (if applicable).
- [ ] Resource parameter descriptions added/updated in the schema.mof.
- [ ] Resource documentation added/updated in README.md.
- [ ] Resource settings.json file contains all required permissions.
- [ ] Examples appropriately added/updated.
- [ ] Unit tests added/updated.
- [x] New/changed code adheres to [DSC Community Style Guidelines](https://dsccommunity.org/styleguidelines).

PR #5466, opened 2024-11-26
[New resource] IntuneRoleScopeTags

### LINKED ISSUE #4151: Intune Scope Tags missing
### Description of the issue

Intune Scope Tags are missing in Microsoft365DSC.
https://learn.microsoft.com/en-us/graph/api/resources/intune-rbac-rolescopetag?view=graph-rest-beta

### Microsoft 365 DSC Version

1.24.110.1

other

### The DSC configuration

_No response_

_No response_

_No response_

ykuijs:
New Intune resource to manage role scope tags

- Fixes #4151

#### Task list

- [X] Resource parameter descriptions added/updated in the schema.mof.
- [X] Resource documentation added/updated in README.md.
- [X] Resource settings.json file contains all required permissions.
- [X] Examples appropriately added/updated.
- [X] Unit tests added/updated.

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneRoleScopeTag/MSFT_IntuneRoleScopeTag.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneRoleScopeTag/MSFT_IntuneRoleScopeTag.psm1
@@ -16,6 +16,10 @@ function Get-TargetResource
         [Parameter()]
         [System.String]
         $Description,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $Assignments,
         #endregion
 
         [Parameter()]
@@ -52,7 +56,7 @@ function Get-TargetResource
         $AccessTokens
     )
 
-    Write-Verbose -Message "Getting configuration of the Intune Role Scope Tag DisplayName {$DisplayName}"
+    Write-Verbose -Message "Getting configuration of the Intune Role Scope Tag with Id {$Id} DisplayName {$DisplayName}"
 
     try
     {
@@ -117,6 +121,14 @@ function Get-TargetResource
             #endregion
         }
 
+        $assignmentsValues = Get-MgBetaDeviceManagementRoleScopeTagAssignment -RoleScopeTagId $Id
+        $assignmentResult = @()
+        if ($assignmentsValues.Count -gt 0)
+        {
+            $assignmentResult += ConvertFrom-IntunePolicyAssignment -Assignments $assignmentsValues -IncludeDeviceFilter $true
+        }
+        $results.Add('Assignments', $assignmentResult)
+
         return [System.Collections.Hashtable] $results
     }
     catch
@@ -148,6 +160,10 @@ function Set-TargetResource
         [Parameter()]
         [System.String]
         $Description,
+
+        [Parameter()]
+        [Microsoft.Management.Infrastructure.CimInstance[]]
+        $Assignments,
         #endregion
 
         [Parameter()]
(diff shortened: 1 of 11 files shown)


ykuijs:
@ricmestre Since this is an Intune resource. Care for a quick review?

Modules/Microsoft365DSC/Examples/Resources/IntuneSecurityBaselineDefenderForEndpoint/1-Create.ps1
```diff
@@ -37,9 +37,9 @@ Configuration Example
                 DisableSafetyFilterOverrideForAppRepUnknown = '1'
             }
             Ensure                = 'Present'
-            ApplicationId         = $ApplicationId;
-            TenantId              = $TenantId;
-            CertificateThumbprint = $CertificateThumbprint;
+            ApplicationId         = $ApplicationId
+            TenantId              = $TenantId
+            CertificateThumbprint = $CertificateThumbprint
```
ricmestre, line 42: Should not be part of this PR, but since it's here... All examples should have the values ending with a semi-colon since that's how M365DSC does exports.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneRoleScopeTag/MSFT_IntuneRoleScopeTag.schema.mof
```diff
@@ -0,0 +1,16 @@
+[ClassVersion("1.0.0.0"), FriendlyName("IntuneRoleScopeTag")]
+class MSFT_IntuneRoleScopeTag : OMI_BaseResource
+{
+    [Key, Description("The display or friendly name of the Role Scope Tag.")] String DisplayName;
+    [Write, Description("The unique identifier for an entity. Read-only.")] String Id;
+    [Write, Description("Description of the Role Scope Tag.")] String Description;
+
```
ricmestre: Extra whitespace/blank line

Modules/Microsoft365DSC/Examples/Resources/IntuneSecurityBaselineDefenderForEndpoint/1-Create.ps1
ykuijs: True, this was a quick update that I spotted while copying these lines. Since PowerShell does not need semi-colons, my preference is to remove them....it is just clutter. 

Because of the C# background of Nik (where the semi-colons are required) he has added the semi-colons to the export 😉, but I would suggest to remove it there as well

ricmestre:
Except for the nitpicks I pointed out above it looks good to me, it should work without issues but I'd like to actually do real tests with it but I'm currently running all my other tests which takes a couple of hours, if you just wait I'll be able to test it.

ykuijs:
Sure, no problem!

Modules/Microsoft365DSC/Examples/Resources/IntuneSecurityBaselineDefenderForEndpoint/1-Create.ps1
ricmestre: I understand but one of the tests I have created is that the resulting blueprint of my converter ends up as-is of what M365DSC would export, in this case *Get-M365DSCExportContentForResource*. For instance all variables are alphabetically ordered and they are all aligned at the equal sign by the longest variable name, if for some reason the resulting blueprint doesn't match what I have created manually then there's a problem somewhere, with your examples not matching that's one of the reasons that I never used them.

My unit tests also make sure that *ConvertTo-DSCObject* behaves as it should.

ykuijs:
Just noticed there is also an Assignment for this resource. Adding that now. Will update the PR shortly.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneRoleScopeTag/MSFT_IntuneRoleScopeTag.psm1
```diff
@@ -0,0 +1,467 @@
...
+
+        if ($null -eq $getValue)
+        {
+            Write-Verbose -Message "Could not find an Intune Role Scope Tag with Id {$Id}"
+
+            if (-not [System.String]::IsNullOrEmpty($DisplayName))
+            {
+                $getValue = Get-MgBetaDeviceManagementRoleScopeTag `
+                    -Filter "DisplayName eq '$DisplayName'" `
+                    -ErrorAction SilentlyContinue
```
FabienTschanz:
```suggestion
                $getValue = Get-MgBetaDeviceManagementRoleScopeTag `
                    -All `
                    -Filter "DisplayName eq '$DisplayName'" `
                    -ErrorAction SilentlyContinue
```

Please add the `-All` switch to the cmdlet, otherwise if more than 25 scope tags are defined and the one with display name XYZ comes after the first 25 scope tags, it won't be found. This is something I'm currently addressing in #5456.

Modules/Microsoft365DSC/Examples/Resources/IntuneSecurityBaselineDefenderForEndpoint/1-Create.ps1
ykuijs: What is your suggestion here? Should we add a semi-colon to all examples or remove them everywhere and make sure the export also doesn't use semi-colons?

ykuijs: And maybe it is an idea to integrate some of the tests you created into the module....that would probably make the quality and predictability of the module much better.

ykuijs:
Just added the Assignment. It is now ready for a complete review/test.

Modules/Microsoft365DSC/Examples/Resources/IntuneSecurityBaselineDefenderForEndpoint/1-Create.ps1
ricmestre: The exports always had semi-colons so my opinion is that the tests should also have them, but also adhere exactly to everything else, main variables sorted alphabetically, main string variables defined within double-quotes, when defined inside CIM instances should be defined instead within single-quotes, equal sign balanced to the longest var name, main string arrays defined all in a single line, then inside CIM instances starting with *@(* and all values after on their own line (and without commas) and correctly idented and another single line just for the closing *);*, and several others which I can't remember right now.

Another big problem is the fact that most of them were never meant to be used as integration test, so even if they are perfectly valid in unit tests you cannot actually deploy them into a real tenant, that's why I made my own single set of examples which I use for both unit and integration tests, some of them are based of off the examples in M365DSC but even those for one reason or another had to be modified.

In the meantime I've been adding examples for new resources but you can take a look at the email I sent you and to Nik back in July which contains an Examples.zip file with the examples I use in my tests.

ricmestre:
@ykuijs I just created a scope tag manually, then I was able to export it, delete it, create it, modify it and delete it again so it's good to go.

One more to add to my examples.

PR #5480, opened 2024-11-28
Implement default formatting to all resources

ykuijs:
This PR implements default formatting to all resources. Since these are only formatting updates, these should not have any impact on the resources itself.

#### Task list


ykuijs:
@ricmestre This is a pretty big PR that impacts lots of files. They should not break anything, but just to be sure: Would you have the possibility to run your tests on them?

No problem if you don't 😉

FabienTschanz:
@ykuijs Can we enforce this formatting in the .vscode settings.json file with indentation, new line enforcement, spaces between brackets etc.?

ykuijs:
@FabienTschanz Good suggestion. We already have formatting settings in the vscode settings.json file. Any suggestions on what might be good to add?

One remark: These settings only work if you use VSCode (which is the recommended tool). But if you are using any other tool, these settings would not make any difference.

ricmestre:
@ykuijs Indeed this is a big one, but are you just changing the format of the files themselves and not the output formatting of exports? If yes then just go ahead, no need to test it.

ykuijs:
This PR indeed only updates the formatting of the files itself: I had VSCode run through all the files and apply the formatting rules as specified in the VSCode settings.json.

So the impact should be absolutely zero, since this does not update any actual logic.

FabienTschanz:
@ykuijs I would love to have method names in Upper case: 

* Wrong: `myObject.add("key", "value")`
* Correct: `myObject.Add("key", "value")`

Currently checking how this could be done.

FabienTschanz:
@ykuijs I have a version that supports methods in upper case with autocorrection. Currently testing that, you can go ahead and merge this one. Looks all good.

PR #5536, opened 2024-12-11
[BUG] Fix duplicate policy creation bug & dynamic group compliance

AlfredSchreuder:
- Fix duplicate policy creation bug, when existing policy was found in tenant, still a new policy was created.
- Fix dynamic group compliance, when dynamic group was found, returned incompliant due to empty parameters Members & GroupAsMembers. Edited so that those params are only added to the comparison when it is a static group.

#### Task list

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
@@ -160,7 +160,7 @@ function Get-TargetResource
             }
         }
 
-        if ([String]::IsNullOrEmpty($policy.id))
+        if ([String]::IsNullOrEmpty($policy.Id))
         {
             Write-Verbose -Message "No existing Policy with name {$DisplayName} were found"
             $currentValues = $PSBoundParameters
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
@@ -129,34 +129,47 @@ function Get-TargetResource
         Add-M365DSCTelemetryEvent -Data $data
         #endregion
 
-        $nullResult = $PSBoundParameters
-        $nullResult.Ensure = 'Absent'
-
+        #Retrieve policy general settings
         $templateReferenceId = 'adc46e5a-f4aa-4ff6-aeff-4f27bc525796_1'
 
-        # Retrieve policy general settings
-        $policy = $null
-        $policy = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Identity -ErrorAction SilentlyContinue
-
-        if ($null -eq $policy)
+        if ($PSBoundParameters.ContainsKey('Identity'))
         {
-            Write-Verbose -Message "No Account Protection LAPS Policy with Id {$Identity} was found"
-
-            if (-not [System.String]::IsNullOrEmpty($DisplayName))
+            Write-Verbose -Message 'PolicyID was specified'
+            try
             {
-                $policy = Get-MgBetaDeviceManagementConfigurationPolicy `
-                    -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'" `
-                    -ErrorAction SilentlyContinue
+                $policy = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Identity -ErrorAction Stop
+            }
+            catch
+            {
+                Write-Verbose -Message "Couldn't find existing policy by ID {$Identity}"
+                $policy = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'"
+                if ($policy.Length -gt 1)
+                {
+                    throw "Duplicate Account Protection LAPS Policyy named $DisplayName exist in tenant"
+                }
+            }
+        }
+        else
+        {
+            Write-Verbose -Message 'Id was NOT specified'
+            ## Can retreive multiple Intune Policies since displayname is not unique
+            $policy = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'"
+            if ($policy.Length -gt 1)
+            {
+                throw "Duplicate Account Protection LAPS Policy named $DisplayName exist in tenant"
             }
         }
 
-        if ($null -eq $policy)
+        if ([String]::IsNullOrEmpty($policy.id))
         {
-            Write-Verbose -Message "No Account Protection LAPS Policy with Name {$DisplayName} was found"
-            return $nullResult
+            Write-Verbose -Message "No existing Policy with name {$DisplayName} were found"
+            $currentValues = $PSBoundParameters
+            $currentValues.Ensure = 'Absent'
+            return $currentValues
         }
+
         $Identity = $policy.Id
-        Write-Verbose "Found Account Protection LAPS Policy with Id {$Identity} and Name {$($policy.Name)}"
+        Write-Verbose -Message "An Account Protection LAPS Policy with Id {$Identity} and Name {$DisplayName} was found"
 
         [array]$settings = Get-MgBetaDeviceManagementConfigurationPolicySetting `
             -DeviceManagementConfigurationPolicyId $Identity `
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
@@ -145,7 +145,7 @@ function Get-TargetResource
                 $policy = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'"
                 if ($policy.Length -gt 1)
                 {
-                    throw "Duplicate Account Protection LAPS Policyy named $DisplayName exist in tenant"
+                    throw "Duplicate Account Protection LAPS Policy named $DisplayName exist in tenant"
                 }
             }
         }
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
@@ -141,7 +141,7 @@ function Get-TargetResource
             }
             catch
             {
-                Write-Verbose -Message "Couldn't find existing policy by ID {$Identity}"
+                Write-Verbose -Message "No Account Protection LAPS Policy with Id {$Identity} was found"
                 $policy = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'"
                 if ($policy.Length -gt 1)
                 {
@@ -162,7 +162,7 @@ function Get-TargetResource
 
         if ([String]::IsNullOrEmpty($policy.Id))
         {
-            Write-Verbose -Message "No existing Policy with name {$DisplayName} were found"
+            Write-Verbose -Message "No Account Protection LAPS Policy with Name {$DisplayName} were found"
             $currentValues = $PSBoundParameters
             $currentValues.Ensure = 'Absent'
             return $currentValues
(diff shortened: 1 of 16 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy/MSFT_IntuneAccountProtectionLocalAdministratorPasswordSolutionPolicy.psm1
```diff
@@ -129,34 +129,47 @@ function Get-TargetResource
...
-                    -ErrorAction SilentlyContinue
+                $policy = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Identity -ErrorAction Stop
+            }
+            catch
+            {
+                Write-Verbose -Message "Couldn't find existing policy by ID {$Identity}"
+                $policy = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'"
+                if ($policy.Length -gt 1)
+                {
+                    throw "Duplicate Account Protection LAPS Policyy named $DisplayName exist in tenant"
```
FabienTschanz:
```suggestion
                    throw "Duplicate Account Protection LAPS Policy named $DisplayName exist in tenant"
```

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
```diff
@@ -85,31 +85,42 @@ function Get-TargetResource
...
-        $nullResult.Ensure = 'Absent'
-
-        $getValue = $null
-        #region resource generator code
-        $getValue = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Id -ErrorAction SilentlyContinue
-
-        if ($null -eq $getValue)
+        if ($PSBoundParameters.ContainsKey('Id'))
         {
-            Write-Verbose -Message "Could not find an Intune Account Protection Policy for Windows10 with Id {$Id}"
```
FabienTschanz, line 97: Why were those messages removed in the resources? They were an indicator which policy with properties x and y was found.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
```diff
@@ -85,31 +85,42 @@ function Get-TargetResource
...
+            }
+            catch
+            {
+                Write-Verbose -Message "Couldn't find existing policy by ID {$Id}"
+                $getValue = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName'"
+                if ($getValue.Length -gt 1)
+                {
+                    throw "Duplicate Intune Account Protection Policy for Windows10 named $DisplayName exist in tenant"
+                }
+            }
```
FabienTschanz: I'm also not fond of switching the ErrorAction to stop and wrapping it in a try/catch block. Try/catch adds massive underlying logic and degrades performance (even though it might not impact us that hard), whereas the previous SilentlyContinue action with a $null result was perfectly enough to indicate that a policy was not found.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
```diff
@@ -85,31 +85,42 @@ function Get-TargetResource
...
-                    -All `
-                    -Filter "Name eq '$DisplayName'" `
-                    -ErrorAction SilentlyContinue
+                throw "Duplicate Intune Account Protection Policy for Windows10 named $DisplayName exist in tenant"
             }
         }
-        #endregion
-        if ($null -eq $getValue)
+
+        if ([String]::IsNullOrEmpty($getValue.id))
```
FabienTschanz: Being a bit nitpicky here, but the property is called `Id` and not `id` (with a capitalized i).

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
```diff
@@ -85,31 +85,42 @@ function Get-TargetResource
...
 
-        $nullResult = $PSBoundParameters
-        $nullResult.Ensure = 'Absent'
-
-        $getValue = $null
-        #region resource generator code
-        $getValue = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Id -ErrorAction SilentlyContinue
-
-        if ($null -eq $getValue)
+        if ($PSBoundParameters.ContainsKey('Id'))
```
FabienTschanz, line 88: Why is this even necessary? Now there is two times the same block: 
```powershell
$getValue = Get-MgBetaDeviceManagementConfigurationPolicy -All -Filter "Name eq '$DisplayName'"
if ($getValue.Length -gt 1)
{
    throw "Duplicate Intune Account Protection Policy for Windows10 named $DisplayName exist in tenant"
}
```

FabienTschanz (commented):
I do like the general approach with checking if there are multiple policies, so this should probably also be added in the resource generator.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
AlfredSchreuder: I do not understand what you mean, any situation of found or not found is present I believe and reported back to the user. All try/catch blocks in all files should report the following cases:
- Id specified and found -> compliant or update
- Id specified and not found by Id, but found by Name -> compliant or update
- Id specified and not found by Id, but found by Name and multiple policies with same Name -> Error, cant decide which policy to check
- Id not specified and found by Name -> compliant or update
- Id not specified and found by Name, but multiple policies with same Name  -> Error, cant decide which policy to check
- All cases above are not applicable and didn't found by Id or Name -> Error, no matchable policy to check

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
AlfredSchreuder: As cases described in the below comment, I believe it is usefull because it are 2 different cases. The code used to result in described cases is integrated from the MSFT_AADConditionalAccessPolicy which had the behaviour that this policies lacked.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
AlfredSchreuder: The previous code did not check compliance of the existing policy found by Name when the Id was not specified. It always resulted in false and created a new policy. I felt like the SilentlyContinue ignored alot of Verbose logging and did not tell me what went right and what went wrong. After implementing this code I allways got the right Verbose logging to tell me what the script did find or did not find.

AlfredSchreuder: Example of old code with no Id specified:
image

Example of old code with wrong Id specified:
image

These examples were applicable for all the files changed. And with this fix it also found the policies in both cases and check compliancy or updated the existing policies.

FabienTschanz:
@AlfredSchreuder Please take a look at my other comments as well. The capitalized i of `Id` is now fixed, but the other comments need to be addressed as well.

AlfredSchreuder:
@FabienTschanz Don't you see my comments on your questions/comments? My bad, apparantly I had to submit the review to actually show the comments, GitHub nooby here.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
FabienTschanz: My point was more that the visibility of which resource is currently in progress is reduced. There is not really an indication of "I'm currently in the Intune Account Protection Policy for Windows10 and doing stuff there", it's a general message with "no real meaning".  I like the previous idea of the messages better, where we immediately saw where we are, if the policy was found (by id, or later by displayName) and so on.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
FabienTschanz: This seems more like an internal bug because the log output tells us that there was a policy found but it still created a new policy, which is incorrect. This should not be the case.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
FabienTschanz: I still believe that only the if check should be added to the modules and not really other changes. I'll check some things and report back.

FabienTschanz:
> @FabienTschanz Don't you see my comments on your questions/comments? My bad, apparantly I had to submit the review to actually show the comments, GitHub nooby here.

No worries, everybody trips somewhere. Now they are visible. I'll report back once I'm done testing.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
AlfredSchreuder: I will change the Verbose output back to about what it was!

AlfredSchreuder:
@FabienTschanz Regarding your testing and open conversations/comments, originally I changed the code according the the MSFT_AADConditionalAccessPolicy handling all below cases, I prefer the changed policies to show the same behaviour, like I showed in the screenshots the old code did not:

- Id specified and found -> compliant or update
- Id specified and not found by Id, but found by Name -> compliant or update
- Id specified and not found by Id, but found by Name and multiple policies with same Name -> Error, cant decide which policy to check
- Id not specified and found by Name -> compliant or update
- Id not specified and found by Name, but multiple policies with same Name -> Error, cant decide which policy to check
- All cases above are not applicable and didn't found by Id or Name -> Error, no matchable policy to check

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
AlfredSchreuder: It should not, but it does, even when compliant. I tested this multiple times on all different policies. After the changes compliant policies actually returned True or when imcompliant changed the existing policies found by name. Feel free to test it yourself.

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAccountProtectionPolicyWindows10/MSFT_IntuneAccountProtectionPolicyWindows10.psm1
AlfredSchreuder: Actually this block that is present 2 times handles the cases where multiple policies are present with the same Name. This was not present in the old code. Maybe the If statement was enough to handle the other cases, I don't know, I copied/used/integrated working code to make these policies show the expected behaviour.

FabienTschanz:
I would like to suggest a combined approach of what we both would like to achieve. 
Edit: I tend to disagree that the new messages are better, they are okay for one resource, but not if specified in multiple policies. Then you don't know with which resource you're currently dealing with.

In the module, when first checking for the id, instead of directly calling the Graph command, let's use the following code (example of LAPS): 
```powershell
# Retrieve policy general settings
$policy = $null
if (-not [System.String]::IsNullOrEmpty($Identity))
{
    $policy = Get-MgBetaDeviceManagementConfigurationPolicy -DeviceManagementConfigurationPolicyId $Identity -ErrorAction SilentlyContinue
}
```

Second, let's update the "fetch by name" like the following: 
```powershell
if (-not [System.String]::IsNullOrEmpty($DisplayName))
{
    $policy = Get-MgBetaDeviceManagementConfigurationPolicy `
        -Filter "Name eq '$DisplayName' and templateReference/TemplateId eq '$templateReferenceId'" `
        -ErrorAction SilentlyContinue

    if ($policy.Length -gt 1)
    {
        throw "Duplicate Account Protection LAPS Policy named $DisplayName exist in tenant"
    }
}
```

This requires a small change in the catch block of the `Get-TargetResource` function:
```powershell
catch
{
    New-M365DSCLogEntry -Message 'Error retrieving data:' `
        -Exception $_ `
        -TenantId $TenantId `
        -Credential $Credential

    # Necessary to rethrow caught exception regarding duplicate policies
    if ($_.Exception.Message -like "Duplicate*")
    {
        throw $_
    }

    $nullResult = Clear-M365DSCAuthenticationParameter -BoundParameters $nullResult
    return $nullResult
}
```

What do you think? Is that a way we can go?

AlfredSchreuder:
I already changed the messages in the last commit, feel free to change the code to whatever you prefer it to be. I spent alot of time to get the expected behaviour I wanted/expected. If you want to write it in your way/style, I am fine with that as long as it has the expected behaviour on all cases as described. Another contributor wrote the code for the MSFT_AADConditionalAccessPolicy module which has the expected behaviour, so I just implemented that.

FabienTschanz:
We appreciate the time and effort you put in. It's in our best interest to fix the issues with minimal changes. I'm simply trying to integrate your proposed changes, which I deem a bit too excessive, in a version that still fixes everything with a bit reduced impact and code changes. 
@ykuijs What's your take on it? Which direction should we continue, or do you have any other input?

AlfredSchreuder:
@ykuijs or @NikCharlebois Could you give your opinion?

PR #5580, opened 2024-12-18
Revert "Implemented changes for the updated Convert-DSCStringParamToVariable function"

NikCharlebois:
Reverts microsoft/Microsoft365DSC#5572


FabienTschanz:
@NikCharlebois What were the issues? Anything we can help with?

ykuijs:
There were several exports that did not work anymore. Much of the code assumes single quotes in CIM instances. With the holiday season coming up, we don't want to have a version that is broken. We will take time to properly test and fix all issues, before rereleasing again.

PR #5401, opened 2024-11-13
[New resource] IntuneMobileAppConfigurationPolicyIos

dannyKBjj:

Issue #5459, opened 2024-11-25
IntuneMobileAppConfigurationPolicyIOS is not needed


ricmestre:
### Description of the issue

*IntuneMobileAppConfigurationPolicyIOS* is not needed, this resource is already covered by *IntuneAppConfigurationDevicePolicy* which takes care of both Android and iOS.

### Microsoft 365 DSC Version

1.24.1120.1

### Which workloads are affected

Intune

### The DSC configuration

```powershell
N/A
```

### Verbose logs showing the problem

```powershell
N/A
```

### Environment Information + PowerShell Version

```powershell
Win11/PS5.1
```

NikCharlebois:
@ykuijs  FYI.

ykuijs:
This was new resource was submitted by @dannyKBjj and merged last week in PR #5401.

@dannyKBjj, were you aware of the resource @ricmestre is mentioning and that is was capable of also managing the iOS policies?

If this is a duplicate, we will have to remove the newly added resource again. Working on a PR now and will include this rollback in that PR.

dannyKBjj:
Are you sure? Because I was testing with what I believed to be the latest release and it was not exporting those policies (hence my creating the module),  there are Managed App policies (which might be what this existing module does) and managed device policy which targets the apps (which is what mine does).... I.e. within Intune GUi -  Apps>App Configuration Policies > create > managed devices... Which despite the name brings you 'Create App Configuration Policy' (for managed devices, rather than managed apps)

dannyKBjj:
Unless this other module was in a very recent release (like within last month, as I can't say I've been religiously updating since downloading the module) I don't think it is anything to do with mine.

ricmestre:
@dannyKBjj See below, I'm able to export the same policy *Apple iOS - Outlook Configuration Profile* using both DSC resources, difference is that *IntuneAppConfigurationDevicePolicy* also works for Android policies.

![Image](https://github.com/user-attachments/assets/5173c18f-5be5-4ba0-b406-77b8956085db)

EDIT: And yes both policies are Managed devices policies

ricmestre:
@dannyKBjj *IntuneAppConfigurationDevicePolicy* has been made available back in May in version 1.24.529.1

dannyKBjj:
Fair enough, I have submitted another module to support Android which was also accepted (but is not in latest release). I guess that is also redundant and will need removing (IntuneAndroidManagedStoreAppConfiguration).

[...]

PR #5588, opened 2024-12-19
Fix for issue - #5587

### LINKED ISSUE #5587: IntuneAppConfigurationDevicePolicy cannot handle 'for Intune' applications
### Description of the issue

 Works well for most application types, but it cannot handle the 'for Intune' apps (e.g. Appian For Intune, Zoom for Intune etc.). However, if you create a configuration with a 'for Intune' application, then trigger Set-Targetresource by deleting the policy, the recreated policy will show a message about the application nolonger being available

I explored the issue a bit and seems be that when recreating the policy, it pick up both the Android and iOS app.

If I create a new policy:

The TargetedMobileApps guid in the policy matches the existing GUID for that.
Image

I then export the policy and delete it. Run start-dscConfiguration to recreate the policy. 

If I pull the TargetedMobileApps, it is now targetted at both the iOS and Android apps
Image

With the end result that the policy in my tenant is broken:
Image

### Microsoft 365 DSC Version

1.24.1211.1

Intune

### The DSC configuration

```powershell
IntuneAppConfigurationDevicePolicy "IntuneAppConfigurationDevicePolicy-Man-Dev-App-ZoomForIntune v2"
        {
            ApplicationId         = $ConfigurationData.NonNodeData.ApplicationId;
            Assignments           = @();
            CertificateThumbprint = $ConfigurationData.NonNodeData.CertificateThumbprint;
            Description           = "";
            DisplayName           = "Man-Dev-App-ZoomForIntune v2";
            Id                    = "50baf12c-db85-4ec5-ae51-385380207c63";
            PayloadJson           = "";
            RoleScopeTagIds       = @("0");
            Settings              = @(
                MSFT_MicrosoftGraphappConfigurationSettingItem{
                    AppConfigKey = 'test'
                    AppConfigKeyType = 'stringType'
                    AppConfigKeyValue = 'test'
                }
            );
            TargetedMobileApps    = @("us.zoom.videomeetings4intune");
            TenantId              = $OrganizationName;
        }
```

```powershell
NA
```

```powershell

```

dannyKBjj:
MSFT_IntuneMobileAppConfigurationPolicyIOS and MSFT_IntuneAndroidManagedStoreAppConfiguration
were supposed to be pulled in favour of existing DSCResource - IntuneAppConfigurationDevicePolicy as my modules duplicated functionality. Whilst testing on our tenant we discovered the following issue: IntuneAppConfigurationDevicePolicy cannot handle 'for Intune' applications #5587

It seems that for some reason only MSFT_IntuneMobileAppConfigurationPolicyIOS was removed. This pull request adds that module back in and deletes "IntuneAppConfigurationDevicePolicy" as I believe that module is now defunct.

See 
https://github.com/microsoft/Microsoft365DSC/pull/5401
https://github.com/microsoft/Microsoft365DSC/issues/5587

- Fixes #5587

#### Task list


FabienTschanz:
@dannyKBjj Removing a resource is a breaking change. The next braking change is (I believe) in April 2025. It would be better to add the functionality to `IntuneAppConfigurationDevicePoliy` instead of replacing it with new resources.

dannyKBjj:
Hi, but the problem is that it is using convoluted logic to handle both. It is inconsistent with almost every other module (which all handle each workload separately) and has caused an issue, who is to say there aren't other issues? Is it possible to just submit my module alongside it and remove the bugged module at a later date? I don't really want to re-write the existing module as I believe the approach I've taken is better anyway (besides which, I have a working solution already).

FabienTschanz:
@dannyKBjj Normally, if there is an issue with a module, we fix the issue in that. The name of the module comes from the separation between the two categories "Managed App" and "Managed Devices". These are also the two things we can select from in the Intune portal, which aligns more with the user experience someone would expect. We already had some cases where we didn't select the "correct" name by workload or by the resource type of what Graph would tell us but rather stay in the Intune portal realms with their naming. 

It's normal for new features and functionality to appear. This happens all the time, so we have to adapt the existing resources for their new purpose. I'd wait until @ykuijs and @NikCharlebois are back and get their opinion as well. I'm just a public contributor adding my input. If you think your solution is better, we'll wait for their feedback. That's totally fine by me.

dannyKBjj:
Hi, fair enough... 

I don't think this is a new feature thing, this is just a fault in the way approach taken.

The issue is that the policy pulled by Get-MgBetaDeviceAppManagementMobileAppConfiguration uses the 'ID' of the targeted mobile app in the 'TargetedMobileApps' property, which is a unique value. 

For reasons I don't quite understand existing IntuneAppConfigurationDevicePolicy doesn't use this ID directly, but instead uses it to retrieve the packageId/bundleId. 

The problem is that for these apps iOS and Android will have share the same value, see below:

image

It then uses value this to retrieve the 'Id' in Set-Target (see below)
image

However, because the app/bundleId for iOS and Android is the same, it will simply retrieve both IDs and write them into the policy (which breaks it).

I'll see what is decided in the new year, but from my perspective, the general approach seems to be to write a module for each resource, as DSC doesn't store the platform information. In this case the module seems to be determining it by the presence of certain properties after the fact. Just seems unnecessarily complicated and out of step with most of the other modules.

I can fix the existing one, but seems fixing it is very much the hard route, when I've got 2 modules that already work? Particularly when one of those modules was never removed and the other was released, and then removed in favour of this one (which doesn't work).

Besides which; I feel my approach is more simple, easier to maintain and more consistent with the other DSC resources?

Anyway, have a great Xmas! I'll pick this up when I'm back in the New Year.

dannyKBjj:
Any further thoughts on this? Do I need to fix the existing module if I want this PR in the near-term release? To be honest if so, I may just run my modules in my environment and wait for the major update in April?

As I say, I believe the resource type approach is more consistent, easier to maintain etc. Also I don't think this is a new feature problem, I think it is a consequence of the approach taken with the previous module? If my PR will be approved I'll wait, if my PR won't be merged, then I'll reconsider...

Ideally though, it'd be included in a near-term release :-D

ykuijs:
@dannyKBjj, thanks for submitting this PR and @FabienTschanz thanks for reviewing it.

Fabien is correct: We only release breaking changes (like removal of resources or parameters) twice a year: First releases of April or October.

In case there is an urgent issue that really breaks something, like the removal of a parameter in a cmdlet that is used, we mark the parameter as Deprecated and leave the parameter. The parameter doesn't work anymore and we write warning messages warning the users that it is deprecated. In the first Breaking change release, the parameter is then removed.

Just to summarize:
- The IntuneMobileAppConfigurationPolicyIOS and IntuneAndroidManagedStoreAppConfiguration resource were added a while back. IntuneMobileAppConfigurationPolicyIOS was removed again, but the IntuneAndroidManagedStoreAppConfiguration resource wasn't (my bad). This was because IntuneAppConfigurationDevicePolicy was already able to configure both and therefore both resource were duplicating functionality. This has been discussed in #5459.
- Now that resource has a different issue. And instead of fixing it, you want to reintroduce the two resources and remove the generic one.
Am I correct?

Just wondering:
- Why isn't fixing the IntuneAppConfigurationDevicePolicy resource possible? If the current logic is incorrect, can be simplified or needs improvement, this would be the prefered option.
- If we indeed decide to move forward, we should not delete the old resources but instead make them deprecated and output warnings. Get and Set methods should output warnings and do nothing. Test method should output warning and always return True.

@ricmestre, what is your opinion on this PR?

ricmestre:
@ykuijs My take is still the same when I said to remove the duplicated resource, it's probably better to have separate resources for this and remove the current one but as you've mentioned that would be a breaking change, so yes either the current one must be accommodated to incorporate the fixes or if the new resources are added then the current must have those warnings added.

To be honest in my solution I have the export running automatically so if I update M365DSC I'll get whatever fixes this issue so I don't mind, but for others might not be a solution to have a breaking change outside its schedule.

ykuijs:
Ok, then lets move forward with the two resources.

Does the Android version of the resource work correctly right now or does it also need updating?

@dannyKBjj: Can you please update this PR to make the IntuneAppConfigurationDevicePolicy deprecated instead of removing it. The code should provide warnings that the resource is deprecated and do nothing else. The Readme of the resource should also say it is deprecated.

dannyKBjj:
Hi yes, I can certainly do that.

I'm not sure what you mean by:

"The code should provide warnings that the resource is deprecated and do nothing else. The Readme of the resource should also say it is deprecated."

Where do I put this code, do you have an example I can refer to?

Do you mean replace/comment out the code in the old module for Get- Set etc. etc. with a Write-Verbose message to say the module is depracated? Should the message mention the new modules?

ykuijs:
> Hi yes, I can certainly do that.
> 
> I'm not sure what you mean by:
> 
> "The code should provide warnings that the resource is deprecated and do nothing else. The Readme of the resource should also say it is deprecated."
> 
> Where do I put this code, do you have an example I can refer to?
> 
> Do you mean replace/comment out the code in the old module for Get- Set etc. etc. with a Write-Verbose message to say the module is depracated? Should the message mention the new modules?

Correct, the code of the Get, Set, Test and Export methods should be replaced with a Write-Verbose message stating that the module is deprecated. The same should be added to the Readme.md of the module.

Unfortunately I do not have an example, because we do not have any other resource that is deprecated.

ykuijs:
Please also update from Dev. A couple of files have changed

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneAppConfigurationDevicePolicy/MSFT_IntuneAppConfigurationDevicePolicy.psm1
```diff
@@ -95,6 +95,8 @@ function Get-TargetResource
         $AccessTokens
     )
 
+    Write-Verbose -Message "This module has been deprecated, please use MSFT_IntuneMobileAppConfigurationPolicyIOS for iOS policy and MSFT_IntuneAndroidManagedStoreAppConfiguration for Android."
```
ykuijs, line 98: This is it. Just add a `return @{}` on the next line to make sure the code stops after this line (same for the other methods).

dannyKBjj: ok, no probs, have done as requested.

dannyKBjj: "Please also update from Dev. A couple of files have changed" I may have messed this up? I fetched Origin in Github desktop, but now I see a conflict?!

dannyKBjj:
Hi, is anyone reviewing this?

ykuijs:
> "Please also update from Dev. A couple of files have changed" I may have messed this up? I fetched Origin in Github desktop, but now I see a conflict?!

That is correct. Recently a lot of Intune resources were updated (same issue as described in [here](https://github.com/microsoft/Microsoft365DSC/issues/5640#issuecomment-2604670952)). So please make sure the conflicts are corrected.

ykuijs:
@dannyKBjj, can you please update with Dev?

dannyKBjj:
Done. Hopefully all OK this time?

dannyKBjj:
Now unit tests are failing!!? Again, everything is fine at my end.

dannyKBjj:
Same issues with #5640
In all cases the the unit test scripts pass in my local environment.
In all cases the test/set/get/export functions are working in my live environment.

ykuijs:
I will download your code and run the tests locally to see what is going on.

Some background info:
I am not sure what is going on here. Most of the times these kind of issues are because there are modules present on your system that are not there on the build agent......which means the tests are failing.

What I sometimes do is deleting all modules (except the default ones) from my system or using a freshly installed VM to run the tests.

ykuijs:
When I run the tests locally, I get the exact same failing messages. Now checking what the root cause is.

ykuijs:
These tests are failing because the resource has been deprecated and the functions always return an empty hashtable. You should update the unit test for IntuneAppConfigurationDevicePolicy to check for an empty array in each function....that is it. The rest of the tests can be removed.

dannyKBjj:
Oh wow, thanks. I'll look into it tomorrow, I could've sworn the errors were in my new modules, but maybe I was hallucinating 😂. Thanks for taking the time to look into it.

dannyKBjj:
tried checking for @{} as returned by the function, that didn't seem to work - strange errors along lines of - expected hashtable, but got hashtable errors. Commented out tests instead and checked for $null where applicable - the script now passes and can easily be reverted should anyone wish to.

dannyKBjj:
hi, all checks passed on this pull a couple of weeks ago, is it possible that someone could review and (hopefully) merge?

dannyKBjj:
hi, still need a review on this please?

dannyKBjj:
Has issue #5588 been fixed somehow? I was looking at simply fixing the module as perhaps that is an easier merge, but now I am going through my steps to reproduce the issue and it seems to be working properly. I checked the code and the logic looks good. I can't find another pull request related to this though.

dannyKBjj:
It looks to me like I may as well close this PR?

dannyKBjj:
Turns out the issue might've been that you can add the same app twice in the Apps blade. As the original module goes off the name it will pull both bundleIDs. My module didn't have the issue as it goes directly off the bundleID. All in all, it seems like a bug with the Apps blade and not the module, so I'm closing this PR.

Issue #5557, opened 2024-12-16
EXOSmtpDaneInbound: New resource proposal


salbeck-sit:
### Description
The resource will enable DANE-support for inbound mails to an accepted domain

### Proposed properties
DomainName - must be among the Accepted Domains

### Special considerations or limitations
Enabling DANE for inbound SMTP relies on enabling DNSSEC for the same domain and subsequently updating the public MX-record
Therefore, enabling support for DANE is not something to be done in a single config.
However, exported configs can show whether DANE-support is enabled and for that reason the resource brings value to the table.

FabienTschanz:
Resource available with 1.25.115.1.

PR #5600, opened 2024-12-30
EXOSmtpDaneInbound - initial release

salbeck-sit:
Implements #5557 

#### Task list


Modules/Microsoft365DSC/DSCResources/MSFT_EXOSmtpDaneInbound/MSFT_EXOSmtpDaneInbound.psm1
```diff
@@ -0,0 +1,343 @@
...
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $DomainName,
+
+        [Parameter()]
+        [ValidateSet('Present', 'Absent')]
+        [System.String]
+        $Ensure,
```
ykuijs, line 14: I wouldn't use the Ensure parameter here. Normally that would be used for objects that can be created or removed, like policies for example. In this case it is for a status, so I would suggest renaming this parameter to SmtpDaneStatus, similar to the cmdlet parameter name.

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSmtpDaneInbound/MSFT_EXOSmtpDaneInbound.psm1
```diff
@@ -0,0 +1,343 @@
...
+        -Parameters $PSBoundParameters
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    $nullResult = $PSBoundParameters
+    $nullResult.Ensure = 'Absent'
+    try
+    {
+        $instance = Get-AcceptedDomain -Identity $DomainName -ErrorAction SilentlyContinue
+        if ($null -eq $instance -or $instance.SmtpDaneStatus -ne 'Enabled')
```
ykuijs, line 57: When renaming the Ensure parameter (see previous comment) you can simply output the value of this parameter and is using the nullResult not necessary

ykuijs (changes requested):
One comment about the naming of the parameter / use of the Ensure parameter.

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSmtpDaneInbound/MSFT_EXOSmtpDaneInbound.psm1
salbeck-sit: I'm not sure I agree. Ensure can be used to actually enable or disable inbound DANE in this resource.

DNSSEC and DANE cannot *both* be switched on in one go due to requirement for an update of the MX-record following enabling DNSSEC - but if DNSSEC is already enabled and the MX-record are in place, then inbound DANE can in fact be enabled with this resource.
Conversely, the resource can also ensure that inbound DANE is disabled if anyone should want that.

ykuijs: Understand what you mean and partially agree. My reasoning was:

Ensure=Present means the setting (policy) should exist, Absent means the setting shouldn't exist. In this case, the setting is always there, just has a different value.

I have tried to use a parameter named Status in the past, but that turned out to be a reserved keyword for DSC which could not be used. That is why I suggested SmtpDaneStatus. Possible values could be Enabled/Disabled. Alternative could be Enabled=$true/$false.

salbeck-sit: To me that would break the unambiguity of Ensure. The resource with Ensure=Present will enable imbound DANE for a given (EXO-Accepted) domain - but won't ensure that it will work if the EXO-cmdlet Enable-SmtpDaneInbound throws an exception. This will cause an error-event being written. Other resources have the same issue.
Using the resource with Ensure=Absent will disable inbound DANE for a given domain.

In an export, Present or Absent accurately represents the status for a given domain.

Replacing Ensure with another property doesn't change the above but can cause ambiguity because now something other that Ensure will have the same meaning as Ensure.

To avoid the above scenario where DANE can't be enabled would require that a configuration (any configuration) can be evaluated in its entirety before being applied, listing any missing prerequisites.

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSmtpDaneInbound/settings.json
```diff
@@ -0,0 +1,33 @@
+{
+    "resourceName": "EXOSmtpDaneInbound",
+    "description": "This resource configures an .",
```
ykuijs: This line looks incomplete

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSmtpDaneInbound/readme.md
```diff
@@ -0,0 +1,12 @@
...
+# EXOSmtpDaneInbound
+
+## Description:
+
+This resource configures SmtpDaneInbound for an accepted domain in Exchange Online.
+Reference: https://learn.microsoft.com/en-us/powershell/module/exchange/enable-smtpdaneinbound?view=exchange-ps
+
+Note that enabling DANE requires that the accepted domain is configured for DNSSEC and the public MX-record updated correspondingly
+
+
```
ykuijs: Please remove duplicate empty lines

Modules/Microsoft365DSC/DSCResources/MSFT_EXOSmtpDaneInbound/MSFT_EXOSmtpDaneInbound.psm1
ykuijs: Got a point there. Just leave it as is.

ykuijs (changes requested):
Few minor updates

ykuijs:
Agree with your view. Leave it as is.

I did discover two minor issues. If you can resolve them, I can merge the PR and include it in this weeks release.

ykuijs (approved):
LGTM

Issue #5615, opened 2025-01-09
Proposed refactor with ~90% performance improvement on Export-TargetResource (for affected resources)


niwamo:
### Description of the issue

The interaction between `Export-TargetResource` and `Get-TargetResource` is highly inefficient for multiple-instance resources (e.g. AADUser or other resources where multiple instances are expected). In pseudo-code:

```PowerShell
function Export-TargetResource
{
    Connect-ToWorkload
    SendTelemetry
    GetAllResources
    foreach ($instance in $instances)
    {
        Get-TargetResource -params
    }
}

function Get-TargetResource
{
    Connect-ToWorkload
    SendTelemetry
    FetchResource
    FormatResourceAndFetchAdditionalProperties 
    return
}
```

There are three significant performance impacts:
1. Calling `New-M365DSCConnection` for every iteration of `Get-TargetResource`. While it avoids reconnecting unnecessarily, it still results in thousands of extraneous + identical API calls for `Get-AcceptedDomain` when it is invoked for the EXO workload (and I have verified this with several Fiddler traces). Even if that could be fixed, there's no need to call the function, and invoking it thousands of times unnecessarily should be avoided.
2. Sending telemetry on every iteration of `Get-TargetResource` is enormously impactful to the performance of this module. Without any other changes, I determined that the same export took 8,925 seconds with telemetry enabled and 1,400 seconds without. In other words, telemetry - just telemetry - accounted for ~85% of my export's runtime. I would by no means suggest removing telemetry altogether, but it does seem highly redundant to log the invocation of `Export-TargetResource` _and_ each invocation of `Get-TargetResource` from within the export. If there is a desire to track the number of `Get-TargetResource` calls or the average function time, I would suggest tracking these items within the export function and sending that data as a single telemetry call. There's just no good reason for anyone to keep telemetry enabled when it causes an export to run (literally) 8x slower; reducing the telemetry without reducing the useful information captured seems like a win for everyone.
[...]

PR #5635, opened 2025-01-16
Report LCM details only if running as administrator

FabienTschanz:
This PR improves the telemetry engine by only reporting the LCM details if the current session is running as administrators. Previously, it was always attempted to get the LCM details using `Get-DscLocalConfigurationManager`, but this wouldn't work for standard users. In their case, it would always throw a `Access denied` exception which is caught and silenced - Nobody knew it was failing. Now, it will check if the current session is elevated and only report those details if they are available. 

Inspired by #5615 to check the telemetry engine for optimizations.  

**Previously: Export took 187 seconds.**
**Afterwards: Export took 19 seconds.**
**Speedup: Factor 10**

None.

#### Task list


ykuijs (approved):
LGTM

PR #5684, opened 2025-01-28
Improve performance of AAD resources

FabienTschanz:
Another PR that improves the performance during the export of some AAD resources mentioned in #5615. 

None.

#### Task list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change log for Microsoft365DSC
 
+# UNRELEASED
+
+* AAD Workload
+  * Improved performance of various resources during export.
+
 # 1.25.122.2
 
 * AADAdminConsentRequestPolicy
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADAccessReviewDefinition/MSFT_AADAccessReviewDefinition.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADAccessReviewDefinition/MSFT_AADAccessReviewDefinition.psm1
@@ -68,47 +68,54 @@ function Get-TargetResource
 
     try
     {
-        $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftGraph' `
-            -InboundParameters $PSBoundParameters
-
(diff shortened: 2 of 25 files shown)


NikCharlebois:
My only concern is that we will lose insights on what resources have the more instances when exported based on Telemetry. Based on the proposed changes, we will no longer capture product telemetry about the # of resources exported since telemetry data will only be sent back when Get-TargetResource gets called outside of the export flow.

FabienTschanz:
All of the other updated resources now also don't send telemetry anymore for exports. Is this something you really want to have? If yes, then we will have to update all resources that were already changed to include something that reports the telemetry, which is modifying another ~200 resources again.

NikCharlebois:
> All of the other updated resources now also don't send telemetry anymore for exports. Is this something you really want to have? If yes, then we will have to update all resources that were already changed to include something that reports the telemetry, which is modifying another ~200 resources again.

They still do, no? As part of the Get-TargetResource. The challenge with the proposed changes is that the Get-TargetResource doesn't send telemetry back if the process is doing and export since it is gated.

FabienTschanz:
Example here of the `IntuneDeviceCompliancePolicyWindows10` which changed last week or so: 
https://github.com/microsoft/Microsoft365DSC/blob/29abf3f781eed04871a6c2dada22c7ef52d41eb1/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceCompliancePolicyWindows10/MSFT_IntuneDeviceCompliancePolicyWindows10.psm1#L183-L197

This is gated as well, so no telemetry is being sent during export.

NikCharlebois:
Ok, we'll have to figure out how we want to address this moving forward. At a minimum, the Export-TargetResource is sending back telemetry and then we have some performance metrics along with the total count (via $Global:M365DSCExportResourceInstancesCount) being logged. I will go ahead and merge the proposed changes and will do a full review of our telemetry strategy next quarter.

FabienTschanz:
Alright, thank you. Should we proceed with updating the other resources for performance or wait for your strategy?

PR #5754, opened 2025-02-11
AADApplication -  Allow for Nullable AuthenticationBehaviors

NikCharlebois:
* AADApplication
  * Changing the AuthenticationBehaviors parameters to string to allow for null values.

* N/A

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
@@ -1177,7 +1177,7 @@ function Set-TargetResource
         $IAuthenticationBehaviors = @{
             blockAzureADGraphAccess       = $AuthenticationBehaviors.blockAzureADGraphAccess
             removeUnverifiedEmailClaim    = $AuthenticationBehaviors.removeUnverifiedEmailClaim
-            #requireClientServicePrincipal = $AuthenticationBehaviors.requireClientServicePrincipal
+            requireClientServicePrincipal = $AuthenticationBehaviors.requireClientServicePrincipal
         }
 
         $uri = (Get-MSCloudLoginConnectionProfile -Workload MicrosoftGraph).ResourceUrl + "beta/applications/$($currentAADApp.Id)/authenticationBehaviors"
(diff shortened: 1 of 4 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_AADApplication/MSFT_AADApplication.psm1
```diff
@@ -1181,11 +1177,11 @@ function Set-TargetResource
         $IAuthenticationBehaviors = @{
             blockAzureADGraphAccess       = $AuthenticationBehaviors.blockAzureADGraphAccess
             removeUnverifiedEmailClaim    = $AuthenticationBehaviors.removeUnverifiedEmailClaim
-            requireClientServicePrincipal = $AuthenticationBehaviors.requireClientServicePrincipal
+            #requireClientServicePrincipal = $AuthenticationBehaviors.requireClientServicePrincipal
```
ricmestre: @NikCharlebois Was this change intentional?

NikCharlebois: This setting cannot be set AFAIK. It throws an error stating that this behavior cannot be set on the App

NikCharlebois: @ricmestre are you able to configure this behavior (requireClientServicePrincipal) on your end?

NikCharlebois: I will add it back to ensure the resource's consistency and will follow-up internally on how to set the flag if possible.

ricmestre: Actually I don't use this resource, I was only proofreading the diff and found it odd you commented the property out. Nevertheless looking at the graph docs, see below, this seems to be indicative of why you're getting the error message and this property may not be changed for all existing applications.

If true, requires multitenant applications to have a service principal in the resource tenant as part of authorization checks before they're granted access tokens. This property is only modifiable for multitenant resource applications that rely on access from clients without a service principal and had this behavior as set to false by Microsoft. Tenant administrators should respond to security advisories sent through Azure Health Service events and the Microsoft 365 message center.

Issue #5433, opened 2024-11-20
Update-M365DSCDependencies: ... does not support PowerShell Core


fasteiner:
### Description of the issue

When running the update Dependencies command in PowerShell Core this happens:
14:22:46  WARNING: The dependency {Az.Accounts} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.
14:22:46  WARNING: The dependency {Az.ResourceGraph} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.
14:22:46  WARNING: The dependency {Az.Resources} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.
14:22:46  WARNING: The dependency {Az.SecurityInsights} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.
14:22:46  WARNING: The dependency {ExchangeOnlineManagement} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.
14:22:48  WARNING: The dependency {Microsoft.PowerApps.Administration.PowerShell} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.
14:22:48  WARNING: The dependency {MicrosoftTeams} does not support PowerShell Core. Please run Update-M365DSCDependencies in Windows PowerShell.

All of this modules are PowerShell Core compatible, therefore you must have a mistake in your dependencies.

Also, this requires me to install the module as administrator, which is not favorable and also I wonder why the Scope switch is being ignored again? 

I'll try downgrading now, but this needs to be fixed!

### Microsoft 365 DSC Version

1.24.1113.1

### Which workloads are affected

other

### The DSC configuration

```powershell
n/a
```

### Verbose logs showing the problem

```powershell
n/a
```

### Environment Information + PowerShell Version

```powershell
OsName               : Microsoft Windows Server 2019 Standard
OsOperatingSystemSKU : StandardServerEdition
OsArchitecture       : 64-bit
WindowsVersion       : 1809
WindowsBuildLabEx    : 17763.1.amd64fre.rs5_release.180914-1434
OsLanguage           : en-US
OsMuiLanguages       : {en-US}

PS C:\Users\svsteiner>
PS C:\Users\svsteiner> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.4.6
PSEdition                      Core
GitCommitId                    7.4.6
[...]

PR #5867, opened 2025-03-03
Proposal: Add PowerShell 5 to 7 compatibility layer

FabienTschanz:
This PR is a proposal request for a Windows PowerShell 5.1 to 7 compatibility layer. It allows invoking a DSC resource from PowerShell 5 into a PowerShell 7 session. Example use case: 

* Run `Start-DscConfiguration` which requires Windows PowerShell 5.1
* Internally, the resource dispatches the call to `Invoke-PowerShellCoreResource` and starts up a PowerShell 7 session
* The function to call is invoked in the PS 7 session
* Results are returned back to the original caller

PS5_to_PS7_compatibility_layer_flow drawio

Sample resource, minimal updates necessary: 

```powershell
$script:CurrentResource = ($PSCommandPath | Split-Path -Leaf).Replace('MSFT_', '').Replace('.psm1', '')

function Get-TargetResource {
    if ($PSEdition -ne 'Core') {
        Invoke-PowerShellCoreResource -Name $script:CurrentResource -Path $PSCommandPath -FunctionName 'Get' -Parameters $PSBoundParameters
        return
    }

	<Normal module code here>
}

function Set-TargetResource {
    if ($PSEdition -ne 'Core') {
        Invoke-PowerShellCoreResource -Name $script:CurrentResource -Path $PSCommandPath -FunctionName 'Set' -Parameters $PSBoundParameters
        return
    }

	<Normal module code here>
}

function Test-TargetResource {
    if ($PSEdition -ne 'Core') {
        Invoke-PowerShellCoreResource -Name $script:CurrentResource -Path $PSCommandPath -FunctionName 'Test' -Parameters $PSBoundParameters
        return
    }

	<Normal module code here>
}
```

Performance-wise, this approach is almost as fast as the native execution, only marginally slower by ~10%. Considering that the function calls must be invoked in another PowerShell session, this is very impressive. 

None. Originated in issue #5433, discussion by @Borgquite, @ykuijs, @salbeck-sit. 

#### Task list


Borgquite:
@FabienTschanz Quick spot - the PR description needs FunctionName updating in the examples (all Set) in case it matters

FabienTschanz:
@Borgquite Thanks, updated it. Copy & paste error 😆

FabienTschanz:
@ykuijs and @NikCharlebois if you can, I would like to have your opinion on the matter as well since you're the maintainers of the project. Do you think this is a way that you think would work as an interim solution to the PowerShell 5 / 7 discussion? 

Would you like to have performance tests of it, checks in other cmdlets as well, mentions in the documentation or any other thing? This PR simply contains the changes that **could** be implemented by the resources. It's also entirely possible to switch the resources bit-by-bit - wouldn't be ideal but would work too. 

Thank you!

Modules/Microsoft365DSC/Modules/M365DSCTelemetryEngine.psm1
```diff
@@ -40,6 +40,29 @@ function Get-M365DSCApplicationInsightsTelemetryClient
...
+function Set-M365DSCLCMConfiguration
+{
+    [CmdletBinding()]
+    param (
+        [Parameter()]
+        [object]
+        $LCMConfig
+    )
+
+    $Script:LCMConfig = $LCMConfig
```
NikCharlebois, line 63: How and where will this get used?

FabienTschanz: It‘s used in `Initialize-PowerShellCoreSession` to set the LCM configuration from within Windows PowerShell 5. If the LCM configuration is not set from within Windows PowerShell 5, `Get-DscLocalConfigurationManager` would be executed in the PowerShell 7 context, replacing the loaded version of 2.0.7 with the default 1.0 version. This leads to an errored execution of `Start-DscConfiguration` because it will try and use the LCM version instead of the direct execution with 2.0.7.

FabienTschanz: @NikCharlebois Did I answer your question or are there any more questions? Is this compatibility layer something we could use, what do you think?

ykuijs:
Few questions:
1. Does this mean that all resources will always be executed in the PSv7 context or would the user have the option to choose which version to use? If PSv7 is always used, it would mean PSv7 always has to be installed, which might not be the case in all instances. To make it resilient, we would have to add a check for that.

2. Currently the PnP module that we are using does not support PSv7 yet. We have to make sure that all modules are supporting PSv7 before we can implement this change.

FabienTschanz:
@ykuijs

To 1: For a consistent environment, all resources should be executed in the PSv7 context. Yes, this means that a dependency has to be introduced for PS7 during installation of M365DSC. How and when this is going to happen is tbd. 

To 2: Also true, we have to make sure that all modules work with PSv7. This should be the case if we update the module version to the latest one, but of course that needs cross-checking. 

The idea of the proposal is to gather information on how to move forward and create a shared understanding of the common goal. As soon as the implementation gets more specific, we‘ll have to do more testing. Does that sound alright?

ykuijs:
That totally makes sense. I see that this PR only adds functions that can be used to accomplish this goal, but doesn't change anything else yet. So it opens up possibilities for testing. From that perspective I am all for merging this PR.

From a performance perspective:
I am very curious what this will do for performance. Like you say, calling into another PowerShell session will add overhead, but at the same time I have seen PSv7 being a whole lot faster than PSv5.1. So probably very interesting to somehow implement a toggle. By default PSv5.1 is used but with the toggle, PSv7 will be used. That way we quickly test the difference in performance.

FabienTschanz:
@ykuijs From what I saw and tested, the performance impact is at around 5% when checking single resources. I‘d have to check for an entire workload. If it‘s ok, I‘ll start with the Intune resources over the next couple of weeks, with an „enabled“ check of the feature. Should be quite easy to implement, and we get a good indication of the performance with 100+ resources.

ykuijs:
Sounds like a plan. @NikCharlebois, what do you think?

NikCharlebois (approved):

NikCharlebois:
@FabienTschanz can you resolve the merge conflict and we will merge in. Thanks

ykuijs:
@FabienTschanz, I already resolved the merge conflict via the GitHub editor. Unit tests are now running.

FabienTschanz:
> @FabienTschanz, I already resolved the merge conflict via the GitHub editor. Unit tests are now running.

Thank you @ykuijs! Currently not on my desktop because of military service, would have had to do it with the online editor which is a pain on the phone 🤣

PR #5875, opened 2025-03-04
O365SearchAndIntelligenceConfigurations - UpdatedPermissions

### LINKED ISSUE #5680: O365SearchAndIntelligenceConfigurations: Which permissions are required to use this resource?
### Description of the issue

Recently this resource has been added the feature of running through a service principal, but the permissions have not been updated in settings.json for application, in fact it won't even work correctly in delegated scenarios since it calls things like *Get-MgGroup* and it only lists User.Read.All/User.ReadWrite.All as required permissions.

Additionally will this actually work with application at all? According to [0] it's possible to use an application to update item insights but it won't work to update people insights and the module is calling *Update-MgBetaOrganizationSettingPersonInsight* so either the module or the documentation is wrong.

[0] https://learn.microsoft.com/en-us/graph/api/insightssettings-update?view=graph-rest-beta&tabs=http

### Microsoft 365 DSC Version

1.25.122.2

Exchange Online

### The DSC configuration

```powershell
N/A
```

```powershell
N/A
```

```powershell
Win11/PS5.1
```

NikCharlebois:
* O365SearchAndIntelligenceConfigurations
  * Updated documentation for permissions

* N/A

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_O365SearchAndIntelligenceConfigurations/settings.json
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_O365SearchAndIntelligenceConfigurations/settings.json
@@ -1,23 +1,31 @@
 ﻿{
     "resourceName": "O365SearchAndIntelligenceConfigurations",
     "description": "This resource configures the Search And Intelligence configuration settings.",
+    "roles": {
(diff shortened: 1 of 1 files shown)


NikCharlebois:
Linking #5680

PR #5877, opened 2025-03-04
Revert "Replace Write-Host with custom function"

NikCharlebois:
@FabienTschanz we'll need to roll back due to various errors. This PR has a major impact surface and we can't risk breaking existing functionality.

Various resources throw similar errors to:
image


FabienTschanz:
@NikCharlebois Can you provide me with an example that breaks? I tested it all in various scenarios and didn't find an issue, so I'm curious why it's breaking.

FabienTschanz:
Received the error as well during an export in the ISE. Currently checking why.

PR #5905, opened 2025-03-11
Fixing caching across all Resources

NikCharlebois:
* Fixing caching to ensure the current instance (in Get-TargetResource) is the right exported instance.

* N/A

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10.psm1
@@ -171,7 +171,7 @@ function Get-TargetResource
 
     try
     {
-        if (-not  -or $Script:exportedInstance.DisplayName -ne $DisplayName
+        if (-not $Script:exportedInstance -or $Script:exportedInstance.DisplayName -ne $DisplayName)
         {
             $ConnectionMode = New-M365DSCConnection -Workload 'MicrosoftGraph' `
                 -InboundParameters $PSBoundParameters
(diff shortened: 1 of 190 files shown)


Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10.psm1
```diff
@@ -171,7 +171,7 @@ function Get-TargetResource
 
     try
     {
-        if (-not $Script:exportedInstance)
+        if (-not  -or $Script:exportedInstance.DisplayName -ne $DisplayName
```
FabienTschanz: Condition is missing a part.

FabienTschanz (commented):
@NikCharlebois looks good to me, just one part where you removed a bit too much of the if condition. In which cases does this check now apply? Is it when e.g. `Test-TargetResource` is called, which will fetch the resource, after I did an export? Or under which circumstances does this apply?

Modules/Microsoft365DSC/DSCResources/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10/MSFT_IntuneDeviceConfigurationVpnPolicyWindows10.psm1
NikCharlebois: Yep, just saw that as well :)

NikCharlebois:
> @NikCharlebois looks good to me, just one part where you removed a bit too much of the if condition. In which cases does this check now apply? Is it when e.g. `Test-TargetResource` is called, which will fetch the resource, after I did an export? Or under which circumstances does this apply?

We have customers that include both direct calls to resources, as well as instances of M365DSCRuleEvaluation in their config.

e.g., 

AADAuthenticationStrengthPolicy 'xyz'
{
 ...
}
AADAuthenticationStrengthPolicy 'xyz2'
{
 ...
}

M365DSCRuleEvaluation 'aaa'
{
  ...
  ResourceName = 'AADAuthenticationStrengthPolicy'
}

If the LCM did not reset the runspace between two runs, it still has the item cached from the M365DSCRuleEvaluation assessment, and then think the first explicitly declared instance doesn't exist and attempts to recreate it.

FabienTschanz:
Alright, got it. Maybe we can also update the DRG to add the second condition for the `PrimaryKey` property as well so newly created resources already contain them. But that's something for another day.

PR #6001, opened 2025-04-09
Add Entra roles to export of permissions list

FabienTschanz:
#### Pull Request (PR) description
This PR adds the `AdministrativeRoles` property to the output of `Get-M365DSCCompiledPermissionList`. It now returns the specified Entra roles, grouped by Read and Update type. Part of the work on #5996.

#### This Pull Request (PR) fixes the following issues
None.

#### Task list

- [x] Added an entry to the change log under the Unreleased section of the file CHANGELOG.md.
      Entry should say what was changed and how that affects users (if applicable), and
      reference the issue being resolved (if applicable).
- [ ] Resource parameter descriptions added/updated in the schema.mof.
- [ ] Resource documentation added/updated in README.md.
- [ ] Resource settings.json file contains all required permissions.
- [ ] Examples appropriately added/updated.
- [ ] Unit tests added/updated.
- [x] New/changed code adheres to [DSC Community Style Guidelines](https://dsccommunity.org/styleguidelines).

PR #6003, opened 2025-04-10
Get-M365DSCCompiledPermissionList - Attempt #2

FabienTschanz:
Second attempt of updating the `Get-M365DSCCompiledPermissionList` function to have an `AdministrativeRoles` property added to its output. Successor of #6001.

@ricmestre Does this one look better to you? The output should now always contain the roles, even if `-PermissionType` is specified. 

None.

#### Task list


Modules/Microsoft365DSC/Modules/M365DSCPermissions.psm1
```diff
@@ -117,9 +120,35 @@ function Get-M365DSCCompiledPermissionList
             {
```
ricmestre, line 120: This way it will never enter the block because $resourceSettings.roles.Count will always be empty since *roles* has only a single object inside. Instead do this:

```powershell
if ($resourceSettings.roles.read.Count -gt 0 -or $resourceSettings.roles.update.Count -gt 0)
```

ricmestre: In fact let's not use Count here too or it can trip with the same problem, let me think

ricmestre: Maybe just *if ($null -ne $resourceSettings.roles.read -or $null -ne $resourceSettings.roles.update)* and the next *if ($null -ne $readRoles -and $readRoles.Count -gt 0)* is not necessary anymore?

FabienTschanz: @ricmestre Thank you for the feedback. True that, just updated the conditional check.

ricmestre:
LGTM 🚀

ricmestre:
By the way I forgot about at least a corner case, there are some EXO required roles that are not granted automatically if you add Exchange Administrator Entra role or to the EXO role group Organization Management. One of these cases is *EXOAddressList* which only has assigned in settings.json the requiredroles to "Address Lists".

That being said, if you add *PermissionType* and *AccessType* the requiredroles property is removed and now you have a non-working workload because the required role was not assigned at the EXO level.

Other two examples of where this happens is *O365OrgSettings* and *O365OrgCustomizationSetting*

FabienTschanz:
@ricmestre I just added the `RequiredRoles` property to the output.

ricmestre:
It couldn't be that easy right?

Imagine I'm checking permissions for all EXO resources, which will get an Entra role of Exchange admin (as part of AdministrativeRoles), that means that I won't need to grant any of the EXO roles RequiredRoles except for the cases when both AdministrativeRoles and RequiredRoleGroups are empty (because Organization Management is the equivalent of Exchange admin but on EXO side). I think that on EXO workload EXOAddressLists is the only case where this happens.

But then there's the same thing happening on some O365 resources.

FabienTschanz:
I have to be honest, you lost me there. Can you give me some output examples so that I can build what you would like to have? Or should the permissions in the settings.json file be updated for some of those resources to better reflect what's needed? I'm no EXO person, my background is mostly in Intune. I'm trying my best here, but I don't understand the needs well enough it seems.

ricmestre:
There are some EXO roles that are not granted by default even if you're EXO admin on Entra, or even Global Administrator for that matter, in order to make it work you need to actually connect to Exchange, create a new role group with the roles you require and also its members, note that adding the members to "Organization Management" (same as Exchange Administrator on Entra) also won't work.

One such case is the role Address Lists needed for *EXOAddressLists*, try the code below and you'll see that 2 resources pop-up here, this also uncovered that *EXOResourceConfig* is actually incorrect, this one you can run without issues with Exchange Administrator Entra role or Organization Management role group which both are missing in settings.json

```powershell
$EXO = Get-M365DSCAllResources|?{$_ -like "EXO*"}
$EXO | %{
	$Permissions = Get-M365DSCCompiledPermissionList -ResourceNameList $_
	if ([String]::IsNullOrEmpty($Permissions.AdministrativeRoles.Read) -and
	    [String]::IsNullOrEmpty($Permissions.AdministrativeRoles.Update) -and
	    [String]::IsNullOrEmpty($Permissions.RequiredRoleGroups) -and
	    ![String]::IsNullOrEmpty($Permissions.RequiredRoles)) {
	    Write-Output "$($_) / $($Permissions.RequiredRoles)"
	}
}
```

ricmestre:
For O365 workload it's slightly different, the EXO roles required to run this workload are part of *Exchange Administrator* but it doesn't make sense granting such an high privilege Entra role when it just needs a couple of EXO roles to work.

```powershell
$O365 = Get-M365DSCAllResources|?{$_ -like "O365*"}
$O365 | %{
	$Permissions = Get-M365DSCCompiledPermissionList -ResourceNameList $_
	if (([String]::IsNullOrEmpty($Permissions.AdministrativeRoles.Read) -and
	    [String]::IsNullOrEmpty($Permissions.AdministrativeRoles.Update) -or
	    !$Permissions.AdministrativeRoles.Read.Contains("Exchange Administrator") -and
	    !$Permissions.AdministrativeRoles.Update.Contains("Exchange Administrator")) -and
	    [String]::IsNullOrEmpty($Permissions.RequiredRoleGroups) -and
	    ![String]::IsNullOrEmpty($Permissions.RequiredRoles)) {
	    Write-Output "$($_) / $($Permissions.RequiredRoles)"
	}
}
```

FabienTschanz:
Now I understand the principle of those permissions and why they are structured like this. But is there now actually anything wrong with the output? If I want the required permissions and roles for multiple resources and one of them contains such a required role (which wouldn't be necessary for all other resources), shouldn't this still be outputted? Or am I wrong with something here?

ricmestre:
Imagine you're trying to retrieve the permissions for all resources in the EXO workload, with your new diff the output now tells me I need to have Exchange admin and also ALL roles inside RequiredRoles, this is not correct if you already have EXO admin and the role *Address Lists* gets mixed and for sure someone will miss to assign it.

ricmestre:
By the way, I'm using Exchange admin Entra role on the EXO workload but there are some resources that require Security admin, but if you already have EXO admin then it's not required... corner cases everywhere

FabienTschanz:
@ricmestre Should I remove the RequiredRoles from the output? For me as an outsider, I think that the output should include everything. I don‘t see it as the responsibility of the function to distinguish between different EXO roles and only include some things if another thing is true - This increases the complexity unnecessarily. In my opinion, it should be the responsibility of the person getting the permissions to actually understand which are included where and what‘s left to assign, not ours to filter everything out because EXO can‘t get its role management straight 😓

ricmestre:
I hate but it is what it is, just keep it and I can filter the output myself, but *EXOResourceConfig* needs to be adjusted too like I mentioned, it's missing EXO admin from Administrative Roles, maybe Global Reader is enough for read? not sure, and then Organization Management in RequiredRoleGroups.

FabienTschanz:
I‘ll check how the resource works and open a separate PR for it. Probably can do that later in the evening.

ykuijs:
@FabienTschanz Is this PR ready for review/merge?

FabienTschanz:
@ykuijs Now it is. Thank you.

ykuijs (approved):

Issue #4578, opened 2024-04-18
Test Method does not work when not defining a 'PrimarySmtpAddress' in for 'EXODistributionGroup'


raandree:
### Description of the issue

`EXODistributionGroup` does not require a `PrimarySmtpAddress` to be set. However, when not defined, the resource creates the distribution group but will never find it, hence the test method will always fail.

Can we make the `PrimarySmtpAddress` mandatory and all other parameters that are?

### Microsoft 365 DSC Version

1.24.313.1

### Which workloads are affected

Exchange Online

### The DSC configuration

```powershell
Configuration Example
{
    Import-DscResource -ModuleName Microsoft365DSC

    $Domain = 'MngEnvMCAP576786.onmicrosoft.com'
    node localhost
    {
        EXODistributionGroup 'DemoDG'
        {
            Alias                              = "demodg";
            #BccBlocked                         = $False;
            #BypassNestedModerationEnabled      = $False;
            DisplayName                        = "My Demo DG";
            Ensure                             = "Present";
            #HiddenGroupMembershipEnabled       = $True;
            #ManagedBy                          = @("admin@$Domain");
            #MemberDepartRestriction            = "Open";
            #MemberJoinRestriction              = "Closed";
            #ModeratedBy                        = @("admin@$Domain");
            #ModerationEnabled                  = $False;
            Identity                           = "DemoDG";
            Name                               = "DemoDG";
            #PrimarySmtpAddress                 = "demodg@$Domain";
            # RequireSenderAuthenticationEnabled = $True;
            #SendModerationNotifications        = "Always";
            ManagedIdentity                    = $True
            TenantId                           = 'MngEnvMCAP576786.onmicrosoft.com'
        }
    }
}

Example -OutputPath C:\DSC

Start-DscConfiguration -Wait -Force -Verbose -Path C:\DSC
```

### Verbose logs showing the problem

```powershell
VERBOSE: [LcmNew365Dev]:                            [[EXODistributionGroup]DemoDG] Test-TargetResource returned False
VERBOSE: [LcmNew365Dev]: LCM:  [ End    Test     ]  [[EXODistributionGroup]DemoDG]  in 1.7120 seconds.
VERBOSE: [LcmNew365Dev]: LCM:  [ Start  Set      ]  [[EXODistributionGroup]DemoDG]
VERBOSE: [LcmNew365Dev]:                            [[EXODistributionGroup]DemoDG] Getting configuration of Distribution Group for DemoDG
[...]

PR #6174, opened 2025-06-13
EXODistributionGroup: Use retrieved Identity to update and delete distribution groups to prevent different group being used in Set vs Get

Borgquite:
Bug fix - at the moment, when deleting or setting a distribution group, we use the Identity parameter provided by the user.

However within Get-TargetResource, there is code to 'get' a different distribution group, if PrimarySmtpAddress is set (probably because similar to Name and Alias, this value is guaranteed unique within a tenant, and is also more intuitive). 

However, when deleting or updating a DG, only the user-provided Identity parameter is used. If there were are two groups, one where the Name of the group matches the user-provided Identity, and another where the primary email address of the group matches the user-provided PrimarySmtpAddress, we will end up 'getting' and 'testing' the second group, but 'setting' or 'deleting' the first.

This update changes the set/delete behaviour to always use the Identity of the group previously retrieved via Get-TargetResource.

Inspired to check this via #4578 although I don't think this is the cause of the user's issue.

#### Task list


Borgquite:
@NikCharlebois @ykuijs Any chance of a review / merge?

M365DSCApprover (approved):

Borgquite:
Thanks @NikCharlebois 👍

PR #6189, opened 2025-06-17
Sync

NikCharlebois:


FabienTschanz:
Ah, are we now so far that the Graph modules won't be fixed? I still have my PR https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/3326 open which will fix the issue with the Intune cmdlets. If only somebody would review it...

NikCharlebois:
While it is important to fix the root cause via your PR, I think we also need to get unblocked on the M365DSC side.

M365DSCApprover (approved):

FabienTschanz:
I‘m all for fixing it here asap, but it‘s very irritating that the underlying issue is known for 2 months or even more, a solution would exist, but nothing happens. For an enterprise style product supported by Microsoft, this is a shame. There are so many issues with it that we sometimes would be better off using Invoke-MgGraphRequest. That‘s not a blame against you or anyone, but a failure in organization. For a product with such a severe impact on customer business, there has to be a proper quality gate.

M365DSCApprover (approved):

PR #6222, opened 2025-06-25
Revert "Add module pinning"

NikCharlebois:
Rolling back @FabienTschanz 

First test failed:
image


M365DSCApprover (approved):

FabienTschanz:
@NikCharlebois That issue comes from the MSCloudLoginAssistant. Looking into it. The PR should be fine, I'll investigate.

FabienTschanz:
@NikCharlebois Please update the `MSCloudLoginAssistant` module with my PR https://github.com/microsoft/MSCloudLoginAssistant/pull/204. This will fix the import issues. These issues were already present since the `CustomEnvironment` was implemented, but with the import validation, it will explicitly fail if something is wrong. Previously they were just swallowed away and ignored.

PR #6315, opened 2025-07-11
Align resource name and improve performance

FabienTschanz:
This PR aligns the naming of the `IntuneDeviceConfigurationSCEPCertificatePolicy` and improves the performance of the module by either: 

* Replacing pipeline commands with their method counterpart on the object
* Using foreach instead of Foreach-Object

Additionally, the module import was sped up. Depending on the number of modules to import, improvements of **up to 60%** can be seen by using `-Alias @() -Cmdlet @() -Variable @() -DisableNameChecking [-SkipEditionCheck]`. Specifying the alias, cmdlet and variable parameters leads to only loading the functions itself, without populating (and thus discovering / scanning) all of the module again and again to detect any possible alias, cmdlet or variable to export. 

Some numbers: 
* Intune export without any changes: ~220 seconds
* Intune export with improved export: ~150 seconds
* Importing all Microsoft.Graph modules  (default): 140 seconds
* Importing all Microsoft.Graph modules (improved): 85 seconds

Tests were conducted with PowerShell 7 and Windows PowerShell 5.1, where Windows PowerShell 5.1. was consistently ~20% slower than PS7 during every use case. 

Profiling shows that the things the most expensive are the web requests to the backend APIs. If we're able to parallelize the processing of those resources, it would massively speed up export and other configurations. 

------------------------------------------------------------------------------------------------------------------------------------

Second part of this PR is an added `-Parallel` switch to the `Export-M365DSCConfiguration` function using `PSParallelPipeline` in the background as a new dependency. While it's not always guaranteed to bring performance improvements, it certainly can and if it does, then it can an improvement of up to 70% faster export.  

The resource consumption is higher when running it in parallel mode, so a recommendation of 8GB of memory was added to the documentation. For sequential operation, this is not necessary. 

And lastly, the PR contains changes to the Intune required modules for those who leverage either `ConvertFrom-IntuneMobileAppAssignment` or `ConvertFrom-IntunePolicyAssignment` (which is about 90% of the Intune resources). 

None.

#### Task list


FabienTschanz:
Finally, ready for review.

FabienTschanz:
If we're going to make it breaking, then let's add some more stuff 😄 I updated the description to incorporate the latest changes.

NikCharlebois:
@FabianTschanz please look at resolving the conflitcs.

FabienTschanz:
@NikCharlebois Conflicts fixed, waiting for unit tests.

M365DSCApprover (approved):

NikCharlebois:
This was incorrectly marked as a breaking change.

PR #6327, opened 2025-07-16
SPOTenantSettings - property OneDriveSharingCapability is now handled as deprecated

salbeck-sit:
The PnP REST API has deprecated OneDriveSharingCapability, this PR replaces trying to update it with a verbose message stating that the property has been deprecated and should be removed.
I have another PR in the works for the next breaking change release.

None (Set-TargetResource throws an error due to this property)

#### Task list


FabienTschanz:
Is there any replacement for it? Maybe add that information as well to make it easier for the users? Also, I would rather use `Write-Warning` instead of `Write-Verbose` for visibility. Verbose output is very verbose and a deprecation notice will be overlooked very easily.

salbeck-sit:
I think that SPOSharingSettings' MySiteSharingCapability should cover it.
The decision to use Write-Verbose was based on what I see in other resources, but I agree that Write-Warning provides better visibility. In general I'd say that writing an event to the eventlog could be even better for unattended jobs.

PR #6406, opened 2025-08-01
Fix get and set error messages in AAD Service Principal

FabienTschanz:
This PR fixes two issues where fetching a non-existing service principal would throw an error and when assigning more than one member to the service principal.

None.

#### Task list


NikCharlebois:
Not quite sure what this PR fixes. I believe fixes in https://github.com/microsoft/Microsoft365DSC/pull/6414 address the recent issues. Are we ok to close this PR? I believe we should be keeping the try/catch block.

FabienTschanz:
@NikCharlebois The try/catch isn't necessary if we use `-ErrorAction SilentlyContinue` (what we're using in any other resource). For the other things, I've had issues with the `PSReference` compared to `ref` because `PSReference` isn't really a reference to the original object, but might be a wrapper for a clone of the original object. That's why I changed that. 

And I removed the `AppRolesAssignedTo` block because at the point where it's defined, it's not initialized as an array, so it will be a hashtable. Meaning if there are multiple app roles, it will throw because the keys are already defined. In a later part on the update, that's where it's actually necessary (and properly defined).

NikCharlebois (approved):

NikCharlebois (approved):

PR #6517, opened 2025-09-15
Fix AADIdentityGovernanceProgram doc

ricmestre:
#### Pull Request (PR) description

I've raised PR  #6428 a couple weeks ago, which was already merged, to fix the *readme.md* doc for *AADIdentityGovernanceProgram*, nevertheless it seems that the pipeline to automatically update the *resource.md* files, through *Update-M365DSCResourceDocumentationPage*, is not running anymore so this new PR applies the same fix to the associated *AADIdentityGovernanceProgram.md* file.

#### Task list

- [ ] Added an entry to the change log under the Unreleased section of the file CHANGELOG.md.
      Entry should say what was changed and how that affects users (if applicable), and
      reference the issue being resolved (if applicable).
- [ ] Resource parameter descriptions added/updated in the schema.mof.
- [ ] Resource documentation added/updated in README.md.
- [ ] Resource settings.json file contains all required permissions.
- [ ] Examples appropriately added/updated.
- [ ] Unit tests added/updated.
- [ ] New/changed code adheres to [DSC Community Style Guidelines](https://dsccommunity.org/styleguidelines).

PR #6518, opened 2025-09-15
Add Schema Definition pipeline

FabienTschanz:
This PR adds an automatic schema creation pipeline. For it to work, the setting `Allow GitHub Actions to create and approve pull requests` in the repository settings --> Actions must be enabled. 

image

The pipeline automatically checks for changes in the SchemaDefinition.json file when a PR is merged and opens a dedicated PR for the changes to be reviewed. This way, no automatic pushs will happen on the branch and can always be double-checked by a reviewer.

Additional pipelines could be added to automatically update the documentation (.md files), if necessary. The base setup is this pipeline file.

Reference for the create PR action: https://github.com/peter-evans/create-pull-request

None.

#### Task list

- [ ] Added an entry to the change log under the Unreleased section of the file CHANGELOG.md.

### DIFF
--- a/.github/workflows/SchemaDefinition.yml
+++ b/.github/workflows/SchemaDefinition.yml
@@ -22,15 +22,43 @@ jobs:
         shell: pwsh
         run: |
           $ErrorActionPreference = "Stop"
-          Import-Module ".\Modules\Microsoft365DSC\Modules\M365DSCSchemaHandler.psm1"
+          Import-Module -Name .\Modules\Microsoft365DSC\Modules\M365DSCSchemaHandler.psm1
           New-M365DSCSchemaDefinition
 
+      - name: Check for changes
+        id: git-check
+        shell: pwsh
+        run: |
+          git config --global core.autocrlf false
+          if ((git status --porcelain).Count -eq 0) {
+            echo "no_changes=true" >> $env:GITHUB_OUTPUT
+          } else {
+            echo "no_changes=false" >> $env:GITHUB_OUTPUT
+          }
+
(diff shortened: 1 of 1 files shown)


ricmestre:
@FabienTschanz How about this one https://github.com/microsoft/Microsoft365DSC/blob/Dev/.github/workflows/Describe%20Resources%20Schemas.yml ?

FabienTschanz:
@ricmestre Because that one does more than simply updating the schema definition and I want it to be very fast. I think it's besser to start fresh with a new template and build on that, instead of updating the "old" pipelines that don't even work anymore because of repository permissions.

ricmestre:
OK makes sense, maybe it was also a permission issue why after I fixed #6428 the documentation wasn't updated automatically by the GH action even though the job ran, so I now raised #6517 to fix the specific doc manually.

FabienTschanz:
It's not a permission issue, it's because Microsoft implemented repository rules which prevent automatic commits to happen to branches. All changes must come from a PR, otherwise they'll get rejected. That's why with this action, we'll create a new PR when changes to the file are detected for manual review and merge. Semi-automatic in that case, but best we can do for now.

FabienTschanz:
@NikCharlebois I updated the PR to always use the existing branch if it does already exist or create a new one with a PR if it doesn't. That way, we should always be up to date with any changes coming in from Dev. Unless conflicting changes arise, that should work. For a first version, I think that's good enough and should keep us going.

NikCharlebois (approved):

PR #6525, opened 2025-09-17
AADCrossTenantIdentitySyncPolicyPartner

NikCharlebois:
* AADCrossTenantIdentitySyncPolicyPartner
  * Initial release.

* N/A

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/MSFT_AADCrossTenantIdentitySyncPolicyPartner.schema.mof
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/MSFT_AADCrossTenantIdentitySyncPolicyPartner.schema.mof
@@ -1,0 +1,15 @@
+[ClassVersion("1.0.0.0"), FriendlyName("AADCrossTenantIdentitySyncPolicyPartner")]
+class MSFT_AADCrossTenantIdentitySyncPolicyPartner : OMI_BaseResource
+{
+    [Key, Description("Id of the associated partner tenant ID.")] String CrossTenantAccessPolicyConfigurationPartnerTenantId;
+    [Write, Description("Display name for the cross-tenant user synchronization policy. Use the name of the partner Microsoft Entra tenant to easily identify the policy.")] String DisplayName;
+    [Write, Description("Defines whether user objects should be synchronized from the partner tenant. False causes any current user synchronization from the source tenant to the target tenant to stop. This property has no impact on existing users who have already been synchronized.")] Boolean IsSyncAllowed;
+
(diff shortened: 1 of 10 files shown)


M365DSCApprover (approved):

Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/settings.json
```diff
@@ -0,0 +1,40 @@
...
+        "update": [
+            "Security Administrator"
+        ]
+    },
+    "permissions": {
+        "graph": {
+            "delegated": {
+                "read": [],
+                "update": []
+            },
```
FabienTschanz, line 17: Can the delegated permissions be added as well? According to the documentation, they are the same as the ones for application.

Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/MSFT_AADCrossTenantIdentitySyncPolicyPartner.schema.mof
```diff
@@ -0,0 +1,15 @@
+[ClassVersion("1.0.0.0"), FriendlyName("AADCrossTenantIdentitySyncPolicyPartner")]
+class MSFT_AADCrossTenantIdentitySyncPolicyPartner : OMI_BaseResource
+{
+    [Key, Description("Id of the associated partner tenant ID.")] String CrossTenantAccessPolicyConfigurationPartnerTenantId;
+    [Write, Description("Display name for the cross-tenant user synchronization policy. Use the name of the partner Microsoft Entra tenant to easily identify the policy.")] String DisplayName;
```
FabienTschanz, line 5: Is the id the key or can it also be the display name?

Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/settings.json
NikCharlebois: We can add them, but those will never be used. We only support app-only auth.

Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/MSFT_AADCrossTenantIdentitySyncPolicyPartner.schema.mof
NikCharlebois: Display name can be null

Modules/Microsoft365DSC/DSCResources/MSFT_AADCrossTenantIdentitySyncPolicyPartner/settings.json
FabienTschanz: Personally, I‘d rather like it if they were there, just for completeness. But I get it

PR #6533, opened 2025-09-23
Refactor resource export modes

### LINKED ISSUE #2714: Export selection improvements - Alignment with Whitepaper and new mode
The currently available options to export a tenant are:

- None
- Lite
- Default
- Full

It would be great to add a new option that would exclude all 'content related' modules. By 'content related' I think about resources that do not export plain settings for a tenant. This would be resources that hold objects like users, teams, groups, sites,... 

In alignment to the whitepaper it would be great to have an option to create several export commands so that for each workload a new configuration is created.

### LINKED ISSUE #5983: Parameter Mode - Clarification on usage of the different modes
There are multiple modes of the export available: `Lite`, `Default` and `Full`. I understand that some things are more data related than configuration, but this list was never updated for e.g. the Intune workload which contains multiple resources that take a long amount of time to export (the administrative templates are just an example). 

From a user perspective, it's also very unclear what this does. My expectation as a user is that when I do a default export, all resources are exported. I don't want to specify a parameter which is quite hard to understand and even harder to find the resources that are excluded in the different modes. This creates a barrier and might lead to gaps in the configuration. I personally wasn't aware of how this works for almost two years, even though I'm a poweruser of the module.

My suggestion is to remove this parameter and export every resource in every export run. If it's necessary to exclude specific resources, this can easily be done with the `-Resources` parameter in combination with `Get-M365DSCAllResources`. It shouldn't be hidden within a difficult to understand mode.

FabienTschanz:
This PR changes the currently used `Lite`, `Default` and `Full` mode into two other modes: 

* `Data`: A mode which targets "filler" resources or resources with many entries, such as users, groups, Intune apps etc.
* `Configuration`: A mode which targets real configuration policies that configure something, e.g. B2B, Intune configuration policies, app protection, EXO malware filter policies etc.

- Fixes #5983 
- Fixes #2714

#### Task list


NikCharlebois (approved):

NikCharlebois:
@FabienTschanz can you please look at resolving the conflicts? Thanks

FabienTschanz:
Will do, give me a couple minutes.

M365DSCApprover (approved):

PR #6592, opened 2025-10-09
Fix Sentinel Alert Rule definition and export

### LINKED ISSUE #6591: Error when generating report for Sentinel
### Description of the issue

Hello,
I have implement DSC to get Sentinel informations. But when I run
New-M365DSCReportFromConfiguration -Type HTML -ConfigurationPath "$DSCPathWork\Sentinel.ps1" -OutputPath "$DSCPathWork\Sentinel.html" 
I'm getting this error:
The member 'Length' is not valid. Valid members are 'DisplayName', 'Severity'. No match was found for the specified search criteria and module name 'Microsoft365DSC'. Try Get-PSRepository to see all available registered module repositories. Exception calling "Parse" with "1" argument(s): "Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx).

Part of the PS1 file causing the problem:
          IncidentConfiguration = MSFT_SentinelAlertRuleIncidentConfiguration{
                groupingConfiguration = MSFT_SentinelAlertRuleIncidentConfigurationGroupingConfiguration{
                    lookbackDuration = "PT5H"
                    matchingMethod = "Selected"
                    groupByEntities = @("Account")
                    groupByAlertDetails = @(
                        MSFT_SentinelAlertRuleIncidentConfigurationGroupingConfigurationAlertDetail{
                            Length = 11  <--- here is the error
                        }
                    )
                    reopenClosedIncident = $False
                    enabled = $True
                }
                createIncident = $True
            };

I'm running the version 1.25.910.1 of Microsoft365DSC.

### Microsoft 365 DSC Version

1.25.910.1 

other

### The DSC configuration

```powershell

```

```powershell

```

```powershell

```

FabienTschanz:
This PR fixes the structure and export of the `SentinelAlertRule` resource.

- Fixes #6591 

#### Task list


FabienTschanz:
I disagree with this being a breaking change. As mentioned in the issue, the export was not correct because it assumed a different format than what the API provided. This meant that the export was already broken and a "proper" configuration was never created.

NikCharlebois:
Agreed about the export. But could people have written a valid configuration manually? If so, those would now be broken, correct?

FabienTschanz:
In theory, that is possible. But on the other hand, I'd counter that with if anybody wrote such a configuration and ran a test, it would always have failed and applying the configuration would also have failed because the format of the request would be incorrect. 

In my case, I almost all the time configure the policy how I want it to look like, do an export, and then change my main config file with the changes I want to introduce. If I see that something is off, I will most certainly not attempt to do it manually but instead see why that is and open an issue. 

Another thing to consider is that this was the very first issue with the Sentinel workload since its introduction one year ago. Given the fact that the export website also didn't mention anything about it, I'd say that not many if any people are using it. And the specific part where the issue appeared had to be configured in the depths of such an alert rule: 
image

All in all, and also when we now update the export website to include all of the new workloads as well, we should probably release this fix, even if it were to impact a handful of customers. Instead of waiting until April next year with potentially many customers using it and not having a proper export.

NikCharlebois (approved):

M365DSCApprover (approved):

PR #6618, opened 2025-10-21
Fixes the Group assignment for IntuneMobileApp policies

manoskavala:
 This PR fix the issue of the  group assignment , with "DisplayName" property. In the  previous version groupId property was mandatory and not optional . 
If the group you were planning to assign was in the same MOF file, there was no existing identifier(groupid) for the group.
Now the display name is also checked, and if the group is found, it is assigned correctly.

#### Task list

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCDRGUtil.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCDRGUtil.psm1
@@ -1870,12 +1870,12 @@ function ConvertTo-IntuneMobileAppAssignment
         [Parameter(Mandatory = $true)]
         [AllowNull()]
         $Assignments,
-
+ 
         [Parameter()]
         [System.Boolean]
         $IncludeDeviceFilter = $true
     )
-
+ 
     if ($null -eq $Script:IntuneAssignmentFilters)
     {
         $Script:IntuneAssignmentFilters = Get-MgBetaDeviceManagementAssignmentFilter -All -ErrorAction SilentlyContinue | ForEach-Object {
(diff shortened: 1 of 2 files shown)


FabienTschanz:
@manoskavala The logic looks pretty much the same to me as it did before, just slightly cleaned up. What's the issue this PR fixes? Was there something not working? 

And please pay attention, it seems like there are some whitespaces at the start of a couple lines. Make sure to remove them.

Modules/Microsoft365DSC/Modules/M365DSCDRGUtil.psm1
```diff
@@ -1870,12 +1870,12 @@ function ConvertTo-IntuneMobileAppAssignment
         [Parameter(Mandatory = $true)]
         [AllowNull()]
         $Assignments,
-
+
```
FabienTschanz: Whitespaces just like here, or on 1878, 1888, 1893 etc. should be cleaned up.
```suggestion

```

FabienTschanz (changes requested):
Please add an entry to the CHANGELOG.md as well, documenting what your PR fixes.

Modules/Microsoft365DSC/Modules/M365DSCDRGUtil.psm1
manoskavala: I edited the Pull Request (PR) description to show the description correctly.
I delete all the whitespaces inside the function.
I add the entry to changelog.

FabienTschanz:
Thanks for the update. There are two functions we currently use: `ConvertTo-IntuneMobileAppAssignment` and `ConvertTo-IntunePolicyAssignment`. Both have very similar logic, and I think the fix you're proposing here can be applied to the second one as well. Can you please check that as well?

manoskavala:
> Thanks for the update. There are two functions we currently use: `ConvertTo-IntuneMobileAppAssignment` and `ConvertTo-IntunePolicyAssignment`. Both have very similar logic, and I think the fix you're proposing here can be applied to the second one as well. Can you please check that as well?

Function `ConvertTo-IntunePolicyAssignment`  seems to be working as expected and does not have the same issue with `GroupID`.

FabienTschanz:
Can you add the entry of the changelog in an alphabetical order (but still above the `Dependencies` section) and readd the line between the parameter definition for vertical spacing. I can add it as well, but don't want to put my fingers in your PR. Otherwise it looks good, I'll check it in a live environment when I have time in a couple of hours.

FabienTschanz (approved):

PR #6636, opened 2025-10-23
Fix type for parameter list

### LINKED ISSUE #6635: TeamsUpdateManagementPolicy: Cannot compile due to empty DisabledInProductMessages
### Description of the issue

@FabienTchanz Another new property you added that on my tenant since I've never set it up and retrieving it with *Get-CsTeamsUpdateManagementPolicy* gets a $null value and the export doesn't complain but trying to compile it to mof will result in error message due to *[ValidateSet('91382d07-8b89-444c-bbcb-cfe43133af33', 'edf2633e-9827-44de-b34c-8b8b9717e84c')]* which is correct according to the docs, but there's no code to cope with $null values.

Image

### Microsoft 365 DSC Version

1.25.1015.1

Teams

### The DSC configuration

```powershell
N/A
```

```powershell
N/A
```

```powershell
Win11/PS5.1
```

FabienTschanz:
This PR fixes the type for the parameter `DisabledInProductMessages` in the `TeamsUpdateManagementPolicy` resource. It was incorrectly defined as String instead of String array. 

- Fixes #6635 

#### Task list


ricmestre:
@FabienTchanz This fixes the issue for me thanks!

Just wanna ask, is the other pipeline that generates the Markdown documentation working properly now? I'm asking because I rely on those files and my linter will report a problem if the value doesn't match the type, which is the case here so if the problem with that pipeline is still not fixed then you'll have to patch the type on the doc as well -> https://github.com/microsoft/Microsoft365DSC/blob/Dev/docs/docs/resources/teams/TeamsUpdateManagementPolicy.md

Yeah just noticed, I opened that file and it still doesn't even have the property yet, that's why I've yet another script which clones up the repo and generates the docs with *Update-M365DSCResourceDocumentationPage* on the fly, I already have the generated doc but had to patch it on my side from *String* to *StringArray[]* so that the linter doesn't complain about it.

FabienTschanz:
@ricmestre I hope that #6637 will fix the issue that the documentation is not being updated. No guarantee though, I'm working on it. Unfortunately, I can't merge my own PRs, which I'd totally like to do for things like that. We'll have to wait until they're approved by @NikCharlebois

M365DSCApprover (approved):

PR #6784, opened 2025-12-26
BREAKING: Add Key Identifiers to CIM instances

FabienTschanz:
This PR adds key identifiers to all CIM resources where applicable. If any property contains CIM instances on arrays, and an individual CIM instance contains at least one identifier or any property that makes the instance unique, it is now required to have one of them. This massively helps in comparison and drift reporting.

None.

#### Task list


tayhall:
It would be great to understand this in more detail once released. Ive struggled with the current drift reporting tools and CIM instances or GUID noise. Yorick's CI/CD deployment adds complications with data files and makes it hard to properly test deployments where code moves from Dev Test through to Production tenants. Ive resorted to writing a few reports myself so anything that helps here is going to be welcomed.

FabienTschanz:
@tayhall This PR adds new key properties to these instances, making them unique in the respective resource configuration they are living in. As an example the `MSFT_MicrosoftGraphMember` instance type in the `AADAdministrativeUnit` resource. There, we change the properties `Identity` and `Type` both to a key identifier. This means that the combination of those two properties makes the instance unique. You can now only specify one instance that has e.g. the type `Group` and the name `Test`. Previously, you could have x number of these instances, which would make the deployment fail.

If you want / need a bit more help, feel free to open a dedicated issue and explain a bit more about your situation. Unfortunately, I can't help you with the stuff from Yorick since I don't use it myself, so you're probably better off opening an issue on his GitHub project(s) for your case.

NikCharlebois:
What is the advantage of marking these properties as Keys? I am struggling to understand what that fixes or prevent. Right now if these properties are omitted, LCM would simply fail, correct?

Is the purpose of this change only to catch the missing properties at compilation time?

FabienTschanz:
@NikCharlebois The purpose of this change is that properties that are required for a resource (either for create or update) are required in the configuration as well. Right now, you could omit pretty much all values and that would work for a diff, but the moment a Set operation is attempted, it would fail during runtime. With this PR, all of these properties would be mandatory to be set to prevent this exact issue and make the deployment more stable. 

A secondary point of the change is that arrays with elements that have distinct unique properties can be filtered for comparison, making it much more efficient to filter and compare the objects against each other instead of having to go through every single object in the array and matching all of the properties, whilst we don't even know if the objects we're comparing right now are really the ones we want to compare.

NikCharlebois:
Perfect. Thanks for the additional details. Will merge once the conflicts are resolved.

NikCharlebois:
Also review the failed tests please.

FabienTschanz:
Tests should succeed after this run.

PR #6800, opened 2026-01-12
EXODistributionGroup - Performance Fix

NikCharlebois:
* Fixed performance issue with the retrieval of managers.

* N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
     FIXES [#6782](https://github.com/microsoft/Microsoft365DSC/issues/6782)
 * AADNetworkAccessForwardingPolicy
   * Fixed an issue where empty `PolicyRules` would throw an exception during Get.
+* EXODistributionGroup
+  * Fixed performance issue with the retrieval of managers.
 * EXOExternalInOutlook
   * Fixed an issue where the export would fail.
     FIXES [#6753](https://github.com/microsoft/Microsoft365DSC/issues/6753)
@@ -95,7 +97,7 @@
     FIXES [#6584](https://github.com/microsoft/Microsoft365DSC/issues/6584)
   * Removed verbose output from `Get-TargetResource`.
   * Updated the error behavior to always throw inside `Get-TargetResource`.
-    
+
 # 1.25.1203.2
 
 * DEPENDENCIES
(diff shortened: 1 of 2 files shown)


FabienTschanz:
Is fetching the members one by one faster than all at once? Just asking here, my lab isn't big enough that I can simulate thousands of members, so I'm curious what the real world examples look like. My guess is that if you only want to resolve a couple of members then yes, fetching individual ones is faster, but if you were to target all of the distribution groups in the tenant and do checks, then it's probably faster to fetch all, correct? Because that logic exists in a couple of places across several modules, so if we change it here, then we might need to change it in other resources as well to stay consistent.

NikCharlebois:
> Is fetching the members one by one faster than all at once? Just asking here, my lab isn't big enough that I can simulate thousands of members, so I'm curious what the real world examples look like. My guess is that if you only want to resolve a couple of members then yes, fetching individual ones is faster, but if you were to target all of the distribution groups in the tenant and do checks, then it's probably faster to fetch all, correct? Because that logic exists in a couple of places across several modules, so if we change it here, then we might need to change it in other resources as well to stay consistent.

Tagging @SNikalaichyk to tap in here, but yes it is in cases where the tenants have millions of recipients.

FabienTschanz:
Thanks for the answer. In respect to the number of recipients, would it make sense to cache this information in the M365DSC module instead of resource-scoped? Since there are multiple resources relying on this information, and each resource has the same logic implemented, an internal cache for this could be created, right? What do you think? It would use the same logic you provided with this PR, resolving members one by one if they do not exist in the cache.

NikCharlebois:
> Thanks for the answer. In respect to the number of recipients, would it make sense to cache this information in the M365DSC module instead of resource-scoped? Since there are multiple resources relying on this information, and each resource has the same logic implemented, an internal cache for this could be created, right? What do you think? It would use the same logic you provided with this PR, resolving members one by one if they do not exist in the cache.

We probably could, but as long as we simply cache the ones that are retrieved and not the entire list (ResultSize==Unlimited). Not something we should address with this PR given the urgency to get the performance issue fixed, but something to consider for a future PR.

FabienTschanz:
Sure thing, I didn‘t expect you to want me to do that right now 😆  I‘ll add it to the list of stuff to do later.

M365DSCApprover (approved):

SNikalaichyk:
I would recommend being extra cautious with any type of hidden caching and shared caches between resource instances, because at scale this causes memory exhaustion. Keep the instances isolated.

FabienTschanz:
@SNikalaichyk Gotcha. Then we will take the longer runtime instead of caching in favor of memory exhaustion and caching issues.

PR #6831, opened 2026-01-23
Update export of role group members

### LINKED ISSUE #6829: S&C Role Group member accepts only user DisplayName
### Description of the issue

Hi,

We are running DSC using Azure DevOps pipelines and maintainging Security And Compliance Role Group Members. But we can see that configuration accepts only Display Names of users. The issue is that we have some users having user accounts and admin accounts with different UPN, but same displayname. When we try to add this kind of users as RoleGroupMembers we get errors because the configuration does not know which identity to add. 
Anny suggestions?

Best regards Mladen

### Microsoft 365 DSC Version

PRD 1.26.114.1

Security & Compliance Center

### The DSC configuration

```powershell
@{
    NonNodeData = @{
        SecurityCompliance = @{
            RoleGroupMember = @(
                @{
                    UniqueId    = "OrganizationManagement"
                    Description = "Members of this management role group have permissions to manage Exchange objects and their properties in the Exchange organization. Members can also delegate role groups and management roles in the organization. This role group shouldn't be deleted."
                    Ensure      = "Present"
                    Members     = @("FirstName LastName")
                    Name        = "OrganizationManagement"
                }
}}}
```

```powershell

```

```powershell

```

FabienTschanz:
This PR updates the export format of the `SCRoleGroupMember` resource, where we now use the `Alias` property if it is an email address. Fallback to `Name` for every other type, like Service principals. This is to ensure uniqueness in a tenant, where multiple users with the same name can exist.

Due to the nature of this change (updating the name to its alias), this will result in drifts although the user itself didn't change, but our logic is not able to detect that without resolving the user. Therefore, I'll mark this as a breaking change.

- Fixes #6829 

#### Task list


NikCharlebois:
Not sure I understand how this is a breaking change, the way I understand this, the current scenario will continue to be evaluated and existing configs won't necessarily break.

FabienTschanz:
@NikCharlebois It's not per-se a breaking change, but configurations will drift until the blueprint is updated with the new values. That's why I marked it as breaking, to prevent customers having a drift show up that's not actually a drift. Just being cautious here.

PR #6893, opened 2026-02-09
Cleanup markdown and other text issues

FabienTschanz:
This PR updates all of the markdown files to conform to certain style guidelines of markdownlint. There is no enforcement, but for consistency in that the markdown parser can interpret the markdown files correctly, some guidelines should be set. 

As mentioned, there is no enforcement for this and no pipeline will fail because of it. A pipeline that runs on each PR will check the documentation and report things to improve based on the style configuration file `.mdlrc`. 

@ricmestre I think this should help you a bit with your parser.

None.

#### Task list


ricmestre:
The way that I ingest these files I really don't care much if they follow Markdown rules or not, problems is that after inserting the values from the tenant in them I run my linter, which is not a Markdown linter by the way to make sure that those values have correct DataType, are within AllowedValues, but also if table headers, empty lines, etc are in the correct places.

So to put it simple as long as these files are all consistent I can change my code to work with them, if they're not consistent then it's a problem :)

NikCharlebois (approved):

PR #6903, opened 2026-02-12
Fix TeamsTemplatesPolicy error handling to prevent false drift (Fixes…

### LINKED ISSUE #6899: TeamsTemplatesPolicy: Error Handling Issues
### Description of the issue

The __TeamsTemplatesPolicy__ DSC resource has two error-handling issues that may result in intermittent false-positive drift detections:

1. `Get-CsTeamsTemplatePermissionPolicy` is called with `-ErrorAction 'SilentlyContinue'`, which suppresses intermittent errors.
2. `Get-CsTeamTemplateList` should explicitly specify `-ErrorAction 'Stop'`. It currently produces a non-terminating error that leads to false-positive drift detection for the `HiddenTemplates` property.

### Microsoft 365 DSC Version

V1.26.128.1

Teams

### The DSC configuration

N/A

```text
[ Start  Resource ]  [[TeamsTemplatesPolicy]TestPolicy] |
[ Start  Test     ]  [[TeamsTemplatesPolicy]TestPolicy] |
                     [[TeamsTemplatesPolicy]TestPolicy] Testing configuration of the TeamsTemplatesPolicy with Identity {TestPolicy} |
                     [[TeamsTemplatesPolicy]TestPolicy] Getting the Teams Calling Policy {TestPolicy} |
                     [[TeamsTemplatesPolicy]TestPolicy] Found Teams Templates Policy {TestPolicy} |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletBeginProcessing: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletProcessRecordStart: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletProcessRecordAsyncStart: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletGetPipeline: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletBeforeAPICall: |
                     [[TeamsTemplatesPolicy]TestPolicy] URLCreated: /Teams.Templates/api/TeamTemplates/v1.0/en-US |
                     [[TeamsTemplatesPolicy]TestPolicy] RequestCreated: /Teams.Templates/api/TeamTemplates/v1.0/en-US |
                     [[TeamsTemplatesPolicy]TestPolicy] HeaderParametersAdded: |
                     [[TeamsTemplatesPolicy]TestPolicy] SetExecutionId: REDACTED |
                     [[TeamsTemplatesPolicy]TestPolicy] BeforeCall: |
                     [[TeamsTemplatesPolicy]TestPolicy] ResponseCreated: |
                     [[TeamsTemplatesPolicy]TestPolicy] BeforeResponseDispatch: |
                     [[TeamsTemplatesPolicy]TestPolicy] Finally: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletProcessRecordAsyncEnd: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletException: ParserException - Expected literal (number, boolean, or null). Was '<'. : at Microsoft.Teams.ConfigAPI.Cmdlets.Generated.Runtime.Json.JsonTokenizer.ReadIdentifer() REDACTED |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletProcessRecordEnd: |
                     [[TeamsTemplatesPolicy]TestPolicy] CmdletEndProcessing: |
                     [[TeamsTemplatesPolicy]TestPolicy] <ConfigurationDrift Source="MSFT_TeamsTemplatesPolicy" TenantId="contoso.onmicrosoft.com" LCMState="ConsistencyCheck"><ParametersNotInDesiredState><Param Name="HiddenTemplates"><CurrentValue></CurrentValue><DesiredValue>REDACTED</DesiredValue></Param></ParametersNotInDesiredState></ConfigurationDrift><DesiredValues>REDACTED </DesiredValues><CurrentValues>REDACTED</CurrentValues></M365DSCEvent>
                     [[TeamsTemplatesPolicy]TestPolicy] Test-M365DSCTargetResource returned False |
[ End    Test     ]  [[TeamsTemplatesPolicy]TestPolicy]  in 31.1170 seconds. |

This event indicates that failure happens when LCM is processing the configuration. Error Id is 1. Error Detail is The SendConfigurationApply function did not succeed.
Resource Id is [TeamsTemplatesPolicy]TestPolicy and Source Info is REDACTED.
Error Message is The PowerShell DSC resource [TeamsTemplatesPolicy]TestPolicy with SourceInfo REDACTED threw one or more non-terminating errors while running the Test-TargetResource functionality.
```

N/A

Shiva210Jyoti:
Fixes #6899

Ensures TeamsTemplatesPolicy cmdlet failures are terminating by replacing SilentlyContinue and implicit non-terminating errors with -ErrorAction Stop for Get-CsTeamsTemplatePermissionPolicy and Get-CsTeamTemplateList.

This prevents partial state evaluation when Teams APIs return malformed responses, eliminating false-positive drift on HiddenTemplates.

Changes are minimal and localized to MSFT_TeamsTemplatesPolicy.psm1.


ricmestre:
What if I want to create a policy that doesn't exist yet? With your diff it will throw an error and the resource won't be created.

Did you actually tested this?

FabienTschanz:
The `-ErrorAction Stop` on `Get-CsTeamTemplateList` is correct in my opinion, and also the one in the `Export-TargetResource`. However, as @ricmestre mentioned, the first instance must not be like that, otherwise it will never work for non-existing instances, which is a very valid scenario. This cannot be fixed on an individual resource level but must rather be regarded as a matter affecting all resources and thus treated as such using better error handling.

ricmestre:
Correct, with that changed the rest of the diff is fine by me

SNikalaichyk:
Agree with @ricmestre.

It is important to handle the actual exception thrown by `Get-CsTeamsTemplatePermissionPolicy` to not re-throw the `Object not found` one and to return valid drift. The actual exception type and message needs to be confirmed first.

NikCharlebois:
@Shiva210Jyoti can you update your PR accordingly? Thanks

Shiva210Jyoti:

Shiva210Jyoti:
Thanks for the feedback everyone.

Updated the implementation so that only the first Get-CsTeamsTemplatePermissionPolicy call is wrapped in targeted try/catch logic. It now returns `$nullReturn` for NotFound scenarios while keeping other failures terminating, which allows creation of new policies while still preventing partial-state evaluation.

Confirmed behavior:
• Non-existing policy → returns Absent state (no error)
• API failure / malformed response → terminating error

Other cmdlet calls retain `-ErrorAction Stop` as suggested.

 Ready for re-review.

FabienTschanz:
@Shiva210Jyoti Please add an entry to the changelog. Thank you.

PR #6958, opened 2026-03-13
Aad authentication method policy fido2 passkey profile

sandrola:
Aad authentication method policy fido2 passkey profile

none

#### Task list


Modules/Microsoft365DSC/DSCResources/MSFT_AADAuthenticationMethodPolicyFido2/MSFT_AADAuthenticationMethodPolicyFido2.schema.mof
```diff
@@ -10,22 +10,36 @@ class MSFT_AADAuthenticationMethodPolicyFido2ExcludeTarget
...
+    [Write, Description("Collection of allowed passkey profile IDs for this target.")] String AllowedPasskeyProfiles[];
+};
+[ClassVersion("1.0.0")]
+class MSFT_AADAuthenticationMethodPolicyFido2PasskeyProfile
+{
+    [Write, Description("The unique identifier for the passkey profile.")] String Id;
+    [Write, Description("The display name for the passkey profile.")] String Name;
+    [Write, Description("The passkey types supported by this profile."), ValueMap{"deviceBound","deviceBound,synced","crossDevice","synced"}, Values{"deviceBound","deviceBound,synced","crossDevice","synced"}] String PasskeyTypes;
+    [Write, Description("The attestation enforcement level for this profile."), ValueMap{"disabled","registrationOnly","registrationAndSignIn"}, Values{"disabled","registrationOnly","registrationAndSignIn"}] String AttestationEnforcement;
+    [Write, Description("Key restrictions for this passkey profile."), EmbeddedInstance("MSFT_MicrosoftGraphFido2KeyRestrictions")] String KeyRestrictions;
```
FabienTschanz, line 31: According to the documentation, each of the five properties is required. I would even go so far and make the `Name` the key, even though Id is probably the real one. But for comparison sake, it would be best to have such a key identifier.

https://learn.microsoft.com/en-us/graph/api/resources/passkeyprofile?view=graph-rest-beta

sandrola: For Backward compatibility there is no need to have the Passkey configured. If defiend as Key it always need the passkey definition also if the feature is not enabled.

FabienTschanz: The required or key attribute only takes effect if an instance of it is specified. If no instance is specified (as it would be normally the case since it's in public preview), it wouldn't do anything. But if somebody wants to create such an instance and doesn't specify all properties, creation or update will most likely fail. Since the definition here is a new one that's only being targeted for the public preview content, it won't affect the other already existing definitions.

NikCharlebois:
@sandrola please take a look at resolving the conflicts when you get a chance. Thanks

sandrola:
Will be the file Schemadefinition.json builded automatically? Because of a new CIM Instance I have edited the file manually to be able to run a test locally. Everytime is conflicting on the PR.

FabienTschanz:
@sandrola There is actually a pipeline which would generate the schema definition for us, unfortunately it's prohibited from creating PRs automatically due to Microsofts repository restrictions... I don't like that either.

FabienTschanz:
By the way, you can generate the schema definition automatically like the following: 

```powershell
Import-Module <Path to M365DSC>\Modules\Microsoft365DSC\Modules\M365DSCSchemaHandler.psm1
New-M365DSCSchemaDefinition
```

This will always generate a fresh schema definition which you can add to your PR.

NikCharlebois (approved):

PR #7009, opened 2026-03-27
BREAKING CHANGE: Store error logs in $env:TEMP

### LINKED ISSUE #7008: SECURITY REPORT: New-M365DSCLogEntry running under LCM creates error log files under C:\Windows\System32
### Description of the issue

When running resources under the LCM, error log files generated by New-M365DSCLogEntry are created under C:\Windows\System32:

```
VERBOSE: [DEVICENAME]:                            [[EXOGroupSettings]TestM365DynamicGroup] Error Log created at 
{file://C:/WINDOWS/system32/41588-M365DSC-ErrorLog.log}
```

This appears to be down to using Get-Location, which uses the current working directory:

https://github.com/microsoft/Microsoft365DSC/blob/876dd59f64ea85d6df5d7885b11b19dcd6f30f2e/Modules/Microsoft365DSC/Modules/M365DSCLogEngine.psm1#L132

Specifically when running under the LCM, the current working directory of C:\Windows\System32 is also a bad choice of location:

- It's a sensitive system folder. I'm fairly sure you'll find that Endpoint Detection and Response software will fire off complaints about DSC doing this at some point
- It's part of the default %PATH% variable. The more files in this folder, the longer it takes to start any process in Windows.
- It's world-readable by all users on the system by default - therefore if the log files contain any sensitive data by accident, they can read them.
- It's hard to clear out these files without running the risk of accidentally deleting a critical system file
- There is no automatic process to clear out files in this folder (unlike the TEMP folders which can be managed using the [System Cleanup Wizard](https://support.microsoft.com/en-gb/windows/free-up-drive-space-in-windows-85529ccb-c365-490d-b548-831022bc9b32))

EDIT: In principle I would also suggest it's not best practice to programmatically and spontaneously create new files in the current working directory, as it is not a fixed location - it can be changed at any time during the script's execution, which could result in other potentially unexpected and unwanted behaviours.

Please can this be changed to the running process's TEMP folder (`$env:TEMP`) which is a suitable location for this kind of error log?

This will save files to the running user's profile (C:\Username\AppData\Local\Temp by default) when running as a user, and to C:\Windows\Temp when running as LOCAL SYSTEM (like the LCM).

### Microsoft 365 DSC Version

1.26.318.1

other

### The DSC configuration

```powershell
Any configuration generating an error
```

```powershell
VERBOSE: [DEVICENAME]:                            [[EXOGroupSettings]TestM365DynamicGroup] Error Log created at 
{file://C:/WINDOWS/system32/41588-M365DSC-ErrorLog.log}
```

```powershell
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsBuildLabEx    : 26100.1.amd64fre.ge_release.240331-1435
OsLanguage           : en-GB
OsMuiLanguages       : {en-GB, en-US}

PSVersion                      5.1.26100.7920
BuildVersion                   10.0.26100.7920
```

Borgquite:
Store error log files in $env:TEMP to resolve various security and performance issues.

EDIT: Marked as Breaking due to concerns raised in comments over CI/CD pipelines that may rely on these files.

- Fixes #7008

#### Task list


Modules/Microsoft365DSC/Modules/M365DSCLogEngine.psm1
```diff
@@ -129,7 +129,7 @@ function New-M365DSCLogEntry
         $LogContent += "`r`n`r`n"
 
         # Write the error content into the log file;
-        $LogFileName = Join-Path -Path (Get-Location).Path -ChildPath $LogFileName
+        $LogFileName = Join-Path -Path $env:TEMP -ChildPath $LogFileName
```
mpoulson, line 132: I don't think this is the right change. If I understand correctly this will change where logs go even for export.  Why not add detection if the working directory is \windows\system32 then route to temp, otherwise leave as working directory?

If it is going to be done it would have to be a breaking change.

I know of a number of systems that expect logs in specific location to detect and debug failures during CI/CD systems that orchestrate DSC runs.

Borgquite: Understood that this could be a breaking change for some, I shall mark it as such.

I would say that if we are spontaneously creating files with random names, they should be in the temporary folder (or another folder, but directly defined by the person running the script). The current working directory can be changed under a number of circumstances (not least by another part of the module - it's not a scoped thing!) and leaving the location of log files which can contain sensitive information like usernames (as these do) to chance isn't responsible, to me.

FabienTschanz:
To me that looks good, but I'd like to have the feedback of @NikCharlebois here as well. I'm not that familiar with the usage in enterprise environments where the error logs are checked after the runs.

FabienTschanz (approved):

Borgquite:
Thanks @FabienTschanz 👍

PR #7011, opened 2026-03-30
EXOSharedMailbox: Add 2 properties and fix issues

ricmestre:

@FabienTschanz I looked into `EXOSharedMailbox` this weekend and saw that it is in need of much love, I added 2 properties that my customers always request to be added when developing custom scripts to deploy shared mailbox so I added them here too, but found several other issues which I try to fix here with this PR, I might have forgotten about something else but with these fixes is the bare minimum to have this resource going.

At the very least it needs to have error handling during `Set` which is completely missing (not included in the PR) and there's also so many more things that could get added (take a look at [0]), but this PR fixes the following:

- Add support for properties `MessageCopyForSendOnBehalfEnabled` and `MessageCopyForSentAsEnabled`
- During `Get` ensure that the retrieved email addresses don't contain the desired `PrimarySMTPAddress` or the current one if the former is null
- Allow creation of the mailbox without a `PrimarySMTPAddress`, if null it deduces it automatically via the given DisplayName/Alias
- Ensure that `AuditEnabled` is set at creation time
- Reduce number of calls to `Set-Mailbox` by calling it only when required during `Set`
- Handle the change of `PrimarySMTPAddress` first and also ensure that `MicrosoftOnlineServicesID (WindowsLiveId)` is changed at the same time so that later, if required, the `EmailAddresses` can be changed without complaining with `Proxy address "oldaddress@mail.com" is used as WindowsLiveId. So it can't be removed from list of email addresses. To remove it, first change the WindowsLiveId.`
- Ensure that `EmailAddresses` to be added or removed are unique and do not include `PrimarySMTPAddress`, if not null, or the current value

[0] https://learn.microsoft.com/en-us/powershell/module/exchangepowershell/set-mailbox?view=exchange-ps

None

#### Task list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
     to `System.String` instead of an implicit conversion to `System.Uri`.
 * EXOMigration
   * Fixed an issue where `UserEmails` could contain empty strings.
+* EXOSharedMailbox
+  * Added support for properties `MessageCopyForSendOnBehalfEnabled` and
+    `MessageCopyForSentAsEnabled` and while here fixed several issues with this
+    resource
 * EXOTenantAllowBlockListItems
   * [BREAKING CHANGE] Removed the Microsoft reserved `SubmissionID` parameter.
   * [BREAKING CHANGE] Updated the property `Action` to be Required instead of Key.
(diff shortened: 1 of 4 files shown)


FabienTschanz (commented):
Can you please replace the `?` short alias with `Where-Object` for consistency? Rest looks good.

ricmestre:
@FabienTschanz Force of habit but it's changed now

FabienTschanz (approved):

Issue #7055, opened 2026-04-12
IntuneEndpointDetectionAndResponsePolicy{Linux/MacOS}: Not able to deploy or modify


ricmestre:
### Description of the issue

@FabienTschanz Found 2 more that I'm not able to deploy anymore, nor modify, with the breaking change release, I created both manually and exporting them they got exported exactly with the same configs I already have so it's not a problem with my configs/schemas, the requests just seem to not being sent correctly and they fail with the same error message `[] : The request is invalid.` and there's nothing else in event viewer.

Note that I'm able to delete them correctly just by using `Ensure = "Absent"`, just not create/modify. I'm also having issues with `IntuneEndpointDetectionAndResponsePolicyWindows10` but this one hasn't been working for quite some time, it's not the same problem affecting those other 2.

### Microsoft 365 DSC Version

1.26.408.1

### Which workloads are affected

Intune

### The DSC configuration

```powershell
     IntuneEndpointDetectionAndResponsePolicyLinux "IntuneEndpointDetectionAndResponsePolicyLinux-IntuneEndpointDetectionAndResponsePolicyLinux_1"
            Assignments          = @(
                MSFT_DeviceManagementConfigurationPolicyAssignments{
                    dataType = "#microsoft.graph.groupAssignmentTarget"
                    deviceAndAppManagementAssignmentFilterType = "none"
                    groupDisplayName = "DummyGroupInclude"
                    groupId = "b0b8fd3f-af2a-453b-be57-80182d599f02"
                }
                MSFT_DeviceManagementConfigurationPolicyAssignments{
                    dataType = "#microsoft.graph.exclusionGroupAssignmentTarget"
                    deviceAndAppManagementAssignmentFilterType = "none"
                    groupDisplayName = "DummyGroupExclude"
                    groupId = "053dc89a-be83-411a-bad3-909904b7239e"
                }
            );
            Credential           = $Credscredential;
            Description          = "IntuneEndpointDetectionAndResponsePolicyLinux_1";
            DisplayName          = "IntuneEndpointDetectionAndResponsePolicyLinux_1";
            Ensure               = "Present";
            Id                   = "e3b05409-5f1d-4b54-b5bd-2c76154162e5";
            RoleScopeTagIds      = @("0");
            tags_item_key        = "GROUP";
            tags_item_value      = "Test";
        }
[...]

PR #7057, opened 2026-04-13
Fix Intune EDR Linux/MacOS invalid create-update request (#7055)

Shirshaw64p:
## Summary
Fixes issue #7055 where `IntuneEndpointDetectionAndResponsePolicyLinux` and `IntuneEndpointDetectionAndResponsePolicyMacOS` could fail create/update with `[] : The request is invalid.`

## Root Cause
Choice setting resolution for tags could fail to map values like `GROUP` to a valid option itemId when metadata provides the value via option `name` rather than only `optionValue.value`.

## Changes
- Add `Name` to `SettingDefinitionOption` and map Graph option `name`.
- Extend choice resolution fallback in `SettingValueResolver`:
  - match by `optionValue.value`
  - then match by option `name`
  - then fallback to itemId pattern
- Avoid serializing empty choice values in `SettingCatalogPolicySettingBuilder`.

## Safety
- Targeted change in shared Intune Settings Catalog resolution path.
- Existing behavior preserved for current `optionValue.value` and itemId matching paths.

## Files
- `src/Microsoft365DSC.Intune/SettingDefinitionOption.cs`
- `src/Microsoft365DSC.Intune/SettingsCatalogHelper.cs`
- `src/Microsoft365DSC.Intune/SettingValueResolver.cs`
- `src/Microsoft365DSC.Intune/SettingCatalogPolicySettingBuilder.cs`

### DIFF
--- a/src/Microsoft365DSC.Intune/SettingCatalogPolicySettingBuilder.cs
+++ b/src/Microsoft365DSC.Intune/SettingCatalogPolicySettingBuilder.cs
@@ -504,7 +504,7 @@ namespace Microsoft365DSC.Intune
             var valueResult = SettingValueResolver.Resolve(
                 settingValueType, settingDefinition, allDefinitions, dscParams);
 
-            if (valueResult?.Value is not null)
+            if (valueResult?.Value is not null && !string.IsNullOrEmpty(valueResult.Value.ToString()))
             {
                 choiceSettingValue["value"] = valueResult.Value;
                 // Derive the value @odata.type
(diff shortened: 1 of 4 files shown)


FabienTschanz:
@Shirshaw64p Did you actually test this? I came up with a different reason why it failed...

Shirshaw64p:
According to co-pilot review it should work. If you have a different approach please do share, I would be happy to see a different approach. Thanks😁

FabienTschanz:
#7061 is my verified and tested approach. It was not an issue of the C# code, but how PowerShell handles arrays when returned from a function.

Issue #7056, opened 2026-04-13
Custom DSC Module causing delta report to fail


ssoabx:
### Description of the issue

We have a custom DSC module for some additional resources (with specific workflows we need in our environment). With the new DSC update they cause the delta report to fail. If removed form the config the delta report runs fine. I assume this comes from the recent C# rewrite?
The specific delta report we are running is the tenant export VS the master config we maintain (before pushing a release). Any chance this can be handled or is this an unsupported/desired scenario?

### Microsoft 365 DSC Version

1.26.408.1

### Which workloads are affected

other

### The DSC configuration

```powershell
Configuration MasterConfig
{
    param (
    )

    $OrganizationName = $ConfigurationData.NonNodeData.OrganizationName

    Import-DscResource -ModuleName 'Microsoft365DSC'
    Import-DscResource -ModuleName 'Microsoft365DSCCustomModule'

    Node localhost
    {
        ....
    }
}
```

### Verbose logs showing the problem

```powershell
##[warning]ConvertTo-DSCObject : Error parsing DSC configuration: Exception calling "ConvertToDscObject" with "4" argument(s): 
##[warning]"The given key was not present in the dictionary."
##[warning]At D:\a\1\s\Modules\Microsoft365DSC\1.26.408.1\Modules\M365DSCReport.psm1:1895 char:22
##[warning]+     $parsedContent = ConvertTo-DSCObject @params
##[warning]+                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~
##[warning]    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
##[warning]    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,ConvertTo-DSCObject
##[warning]Exception calling "ConvertToDscObject" with "4" argument(s): "The given key was not present in the dictionary."
##[warning]At C:\Program Files\WindowsPowerShell\Modules\DSCParser\3.0.0.2\DSCParser.psm1:164 char:13
##[warning]+             $result = [DSCParser.CSharp.DscParser]::ConvertToDscObjec ...
##[warning]+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
##[warning]    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
##[warning]    + FullyQualifiedErrorId : KeyNotFoundException
```

[...]

PR #7059, opened 2026-04-13
Harden delta report parsing when custom DSC modules are present (#7056)

Shirshaw64p:
## Summary
Addresses issue #7056 where delta report parsing can fail with custom DSC modules due to parser key lookup failures.

## Change
In `Initialize-M365DSCReporting`:
- Keep the current primary parse path (`ConvertTo-DSCObject` with `DscResourceInfo` when supplied).
- If parsing fails and `DscResourceInfo` was provided, retry parsing once **without** `DscResourceInfo`.
- Emit a warning with the original failure details before fallback.
- Preserve existing behavior for non-fallback scenarios (rethrow when no fallback context exists).

## Why this helps
Custom/non-Microsoft365DSC resources may not map cleanly when constrained by `DscResourceInfo`. Retrying without this hint allows parsing to continue and prevents hard failures in delta report generation for mixed configurations.

## Scope
Single-file, minimal-risk resiliency change:
- `Modules/Microsoft365DSC/Modules/M365DSCReport.psm1`

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReport.psm1
@@ -1892,7 +1892,23 @@ function Initialize-M365DSCReporting
     {
         $params.Add('DscResourceInfo', $DscResourceInfo)
     }
-    $parsedContent = ConvertTo-DSCObject @params
+    try
+    {
+        $parsedContent = ConvertTo-DSCObject @params
+    }
+    catch
+    {
+        if ($PSBoundParameters.ContainsKey('DscResourceInfo'))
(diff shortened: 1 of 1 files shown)


FabienTschanz:
As in the other PR, did you actually test this? There were more Copilot PRs in the last two weeks than ever before, and almost each one was doing random stuff without fixing the actual issue.

Shirshaw64p:
I didn't test this one explicitly.

ssoabx:
fwiw this wasn't the fix

FabienTschanz:
Correct. The real issue was at another place, therefore I closed this one.

PR #7079, opened 2026-04-22
M365DSCReverse: Properly setup string replacement map entries

ricmestre:

@FabienTschanz Still enjoying the holidays but while I was looking at improving on how I generate my configuration data file I noticed this problem, looks like the keys and values for the string replacement map entries are being added reversed.

#### Task list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
 * M365DSCIntuneUtil
   * Fixed an issue where a settings array was returned as a single element.
     FIXES [#7055](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7055)
+* M365DSCReverse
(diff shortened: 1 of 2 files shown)


FabienTschanz:
That is not correct. The replacement map has the following structure: 

- Key: The value to replace
- Value: The string to replace the key with

When we write this to the ConfigurationData file, the order has to be the Value first, then the Key. Example: 
`Export-M365DSCConfiguration [...] -TokenReplacement @{ 'alternate-email.onmicrosoft.com' = 'AlternateEmail' }`

If we were to reverse this, then the ConfigurationData will have the variable `alternate-email.onmicrosoft.com` with the value `AlternateEmail`, which is not correct.

ricmestre:
Yeah I stand corrected, it looked odd but should have tested it before which I did now and it works just like you say, closing

PR #7082, opened 2026-04-24
Introduce Microsoft Graph Shim layer

### LINKED ISSUE #3944: IntunePolicySets: Not usable for cloning scenario
### Description of the issue

While cloning *IntunePolicySets* to a different tenant it creates the PolicySet with the *Items* property which contains the sets of policies with their corresponding Ids of the source tenant (which of course will be different in target tenant) and although the property also contains a sub-property with the name of the policies it actually won't fall back looking for them in the target, so the PolicySet is actually created but with not existing policies associated with it.

Also the *Id* property of the PolicySet itself is also still set as mandatory where it shouldn't.

### Microsoft 365 DSC Version

1.23.1122.1

other

### The DSC configuration

```powershell
IntunePolicySets "IntunePolicySets-IntunePolicySets_1"
        {
            Assignments          = @(
                }
                }
            );
            Description          = "";
            DisplayName          = "IntunePolicySets_1";
            GuidedDeploymentTags = @();
            Id                   = "ba1add14-a2bc-44dc-9808-ddb1d7fdb6da";
            Items                = @(
                MSFT_DeviceManagementConfigurationPolicyItems{
                    guidedDeploymentTags = @()
                    payloadId = 'M_e408c831-8c5a-43f7-b2b5-9a6cca47a93e'
                    displayName = 'Windows 10/11 - Information Protection - General'
                    dataType = '#microsoft.graph.mdmWindowsInformationProtectionPolicyPolicySetItem'
                    itemType = '#microsoft.graph.wipConfiguration'
                }
            );
            RoleScopeTags        = @("0");
        }
```

```powershell
After cloning the policy to a different tenant then if accessing the PolicySet in Intune web portal it will complain about "Some of the objects in this policy set are missing".
```

```powershell
OsArchitecture       : 64-bit
WindowsVersion       : 2009
OsLanguage           : en-US

PSVersion                      5.1.22621.2428
BuildVersion                   10.0.22621.2428
```

FabienTschanz:
This PR is a complete refactor of how Microsoft Graph PowerShell commands are used in Microsoft365DSC. Instead of depending on the Microsoft Graph Powershell modules, there is a new autogenerated wrapper that uses `Invoke-MgGraphRequest` as the mechanism to fetch / push data to Microsoft Graph. 

What does this mean? Basically much faster startup and installation times. 

* All dependencies for `Microsoft.Graph.*` were removed except `Microsoft.Graph.Authentication`, and therefore only about half of the previous modules need to be installed now. 
* Installation time reduced by 50% (when using `Update-M365DSCDependencies`)
* Initialization time of the LCM reduced by up to 80% until first resource execution (somewhere between 10-40 seconds)
* Much less disk space used because of eliminating the Graph modules (roughly ~1GB less disk space used)
* For developers, no more need to handle `AdditionalProperties` during resource development.
* Auto-handling of throttled requests (error 429 from the Graph API)

While testing the Shim layer, many small issues with array handling inconsistencies, parameters and so on were discovered and fixed.

- Fixes #3944 

#### Task list


1Dimitri:
Interesting promising work :-)
Please note that there may be regressions though depending on how you'll write the interpretation  of results.
I came across the scenario where I have a `AADGroup` whose group cannot be assigned to Entra ID roles (`isAssignableToRole` property).
In the current 1.26.506.2 version, I get a consistent "false" answer from M365DSC when the UI portal says "no", and true if  UI portal says "yes".
However, because of a 3rd party compliance tool used by my colleagues, I noticed that the false status may be coded as real false or with a null value in the raw oData request or the return value given by `Get-MgGroup` / `Get-MgBetaGroup`

FabienTschanz:
@1Dimitri Thank you for the information. I personally haven't seen this behavior and a quick check shows me that the return value from the Graph API is indeed a boolean and not a null value for the `isAssignableToRole` property. I did some testing with my own lab infrastructure and didn't find any differences in missing properties or values, so I'm very confident that it does work. But of course, there can always be something missing, especially with a change this big. But the benefits we gain with it will help us tremendously in the future when we expand the module.

1Dimitri:
> @1Dimitri Thank you for the information. I personally haven't seen this behavior and a quick check shows me that the return value from the Graph API is indeed a boolean and not a null value for the `isAssignableToRole` property. 


Since I've got null values in 2 tenants, one of which is the source for dumps with M365DSC, I've opened a case at MS using our Unified Entreprise contract.
I'll keep you posted.

1Dimitri:
And the answer from the Support Case is:

> We would like to clarify that while this property is defined as a Boolean in the documentation, it is actually implemented as a nullable Boolean, which means it can return true,false, or null.
•	When querying group membership through endpoints such as /memberOf, Microsoft Graph may return a limited representation of the group object. In such cases, only basic properties (like id and @odata.type are guaranteed, and other properties (including isAssignableToRole) can appear as null.
•	This can also occur if the application does not have sufficient permissions (for example, missing Directory.Read.All or equivalent scopes), resulting in partial data being returned.
(...)
 
> About your observation with the Get-MgBetaGroup cmdlet and Groups.Read.All permissions: this is expected in some situations. Even though Get-MgBetaGroup queries Microsoft Graph directly, not all properties are always included in the default response. Some fields may return null or partial values (like false/null instead of true) depending on factors such as:
>•	If the property is part of the default property set
•	If it is conditionally returned
•	The availability of backend data or how Graph optimizes responses
To get more complete data, you might need to specifically request certain properties using the -Property parameter, for example:
 
> `Get-MgBetaGroup -GroupId <GroupId> -Property <PropertyName>`
or
`Get-MgBetaGroup -All -Property <Property1>, <Property2>`

FabienTschanz:
@1Dimitri Do I understand correctly that the `Get-MgGroup` always returns a Boolean for that property while the API response is a nullable Boolean? So a null value returned by the API is automatically converted to False?

1Dimitri:
Get-MgBetaGroup AND the beta API may return false or null for the 'false' value.
The 2 methods are consistent for a given group, at least for the samples I have.
The v1.0 API/Get-MgGroup also gives back the results.

In fact, from the Entra Portal you cannot tell if you will get false or null if you set 'No' for example for this specific example.

In my example I have 2 tenants, the same 'group' (up to the name case, excluding Id of course) doesn't show any drift according to the AADGroup export as it is coded today.

In particular the isAssignableToRole is set to false BUT the underlying cmdlet/API executed at the same time, and  in the same powershell session, exhibits in one case a null and in the other case a false value.

Example

```

PS> (Get-MgGroup -GroupId 'e0686309-e8ea-48b7-b041-76bd2822b860').isAssignableToRole -eq $null
True
PS>  (Get-MgGroup -GroupId 'e0686309-e8ea-48b7-b041-76bd2822b860') | Select Id, DisplayName, isAssignableToRole

Id                                   DisplayName             IsAssignableToRole
--                                   -----------             ------------------
e0686309-e8ea-48b7-b041-76bd2822b860 O365_AllowedCreateGroup
```
```

PS> (Get-MgGroup -GroupId a08843c5-ae0c-46e0-8ab0-82408c64cba8) | Select Id, DisplayName, isAssignableToRole

Id                                   DisplayName             IsAssignableToRole
--                                   -----------             ------------------
a08843c5-ae0c-46e0-8ab0-82408c64cba8 O365_AllowedCreateGroup              False
```

FabienTschanz:
Okay, got it. So if there is no difference between the value returned by the Graph API and what the Get-MgGroup cmdlet shows, then the result of M365DSC should be identical since the logic itself of how we treat these values didn't change (except for some specific edge cases). Thank you for the digging and raising awareness. Appreciate it.

PR #7194, opened 2026-06-03
fix certificate handling in AADOrganizationCertificateBasedAuthConfig…

saeekailas:
Hey, found the exact cause and fixed it.
The problem is in Get-TargetResource in MSFT_AADOrganizationCertificateBasedAuthConfiguration.psm1. There's a single line that calls [System.Convert]::ToBase64String() on the certificate property coming back from Graph which worked fine before because Graph used to return it as a raw byte[]. At some point (looks like around https://github.com/Microsoft365DSC/Microsoft365DSC/commit/0f9b96629fa4969575ef107d87f148925546236f) the SDK started returning it already as a Base64 string, so calling ToBase64String() on a string blows up trying to treat each character as a byte. That's exactly the "Input string was not in a correct format" error you're seeing.

The fix is a type check before the conversion:

if ($currentCertificateAuthorities.certificate -is [System.Byte[]])
{
$certValue = [System.Convert]::ToBase64String($currentCertificateAuthorities.certificate)
}
else
{
$certValue = $currentCertificateAuthorities.certificate
}
$myCertificateAuthorities.Add('Certificate', $certValue)

No other functions need touching Test-TargetResource and Export-TargetResource both go through Get-TargetResource so they're fixed automatically. Set-TargetResource goes the other direction (FromBase64String) so that was never broken.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
     in `ServicePrincipalFilterRule`.
 * AADFilteringPolicyRule
   * Fixed an issue where an incorrect number of instances to export was shown.
+* AADOrganizationCertificateBasedAuthConfiguration
+  * Fixed an issue where the `Certificate` property was incorrectly converted using `ToBase64String` when the Graph SDK already returns it as a Base64 string. FIXES [#7193](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7193)
 * AzureRoleDefinition
   * Initial Release
     FIXES [#7077](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7077)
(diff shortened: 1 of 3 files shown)


FabienTschanz:
@saeekailas Please add an entry to the changelog. Also, you can remove the type check and assign the value directly. The cmdlet will (now) always return the base64 string. Thank you.

saeekailas:
@FabienTschanz Done! Simplified the type check and assigned the value directly as suggested, and added the changelog entry. Thanks for the quick review and the pointer makes the fix much cleaner. 
Ready for another look whenever you get a chance!

FabienTschanz:
Please add the changelog entry in this PR for consistency and not in a separate one. Thanks.

saeekailas:
Got it Moved the changelog entry to this PR. Sorry about that

FabienTschanz (approved):

ssoabx:
Thanks!

saeekailas:
Thanks for the review and the quick merge @FabienTschanz happy to have contributed. 
Let me know if anything else needs fixing.

PR #7199, opened 2026-06-04
Refactor AADUser to improve group membership handling

### LINKED ISSUE #7198: AADUser Batch pulls AdministrativeUnits and process as groups
### Description of the issue

https://github.com/Microsoft365DSC/Microsoft365DSC/blob/Dev/Modules/Microsoft365DSC/DSCResources/MSFT_AADUser/MSFT_AADUser.psm1#L231

Batch MemberOf pulls Administrative Units which then breaks  Remove-MgGroupMemberDirectoryObjectByRef because the Get-mgroup returns null.  

Fix is 
1. update Batch to ``` url     = "/users/$($UserPrincipalName)/memberOf/microsoft.graph.group?`$select=displayName&`$filter=not(groupTypes/any(c:c eq 'DynamicMembership'))"```
2. add if null check for the else clause before Remove-MgGroupMemberDirectoryObjectByRef
```
 if ($null -eq $group)
{
    New-M365DSCLogEntry -Message 'Error updating data:' `
        -Exception "Attempting to add a user to a group that doesn't exist" `
        -TenantId $TenantId `
        -Credential $Credential

    throw "Group '$($_.InputObject)' does not exist in tenant"
}
```

### Microsoft 365 DSC Version

Dev

### The DSC configuration

```powershell

```

```powershell
Cannot bind argument to parameter 'GroupId' because it is an empty string.                                                  + CategoryInfo          : InvalidData: (:) [], CimException                                                             + FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed,Remove-MgGroupMemberDirectoryObjec     tByRef                                                                                                                   + PSComputerName        : localhost
```

```powershell

```

saeekailas:
Fixes #7198

The `MemberOf` batch request in `Get-TargetResource` was hitting `/memberOf` which returns all directory objects a user is a member of including Administrative Units. When `Get-MgGroup` was subsequently called with an Administrative Unit ID it returned `$null`, causing `Remove-MgGroupMemberDirectoryObjectByRef` to receive an empty `GroupId` string and throw:

```
Cannot bind argument to parameter 'GroupId' because it is an empty string.
```

---

### `Get-TargetResource` URL (`MSFT_AADUser.psm1`)

Added `/microsoft.graph.group` OData cast to the batch URL so the Graph API returns **only Group objects**, excluding Administrative Units at the source.

Also moved the dynamic group exclusion filter from the URL query to **client-side PowerShell**, because combining an OData type cast with `$filter` in a batch request can produce a `400 Bad Request`.

**Before:**
```powershell
url = "/users/$($UserPrincipalName)/memberOf?`$select=displayName&`$filter=not(groupTypes/any(c:c eq 'DynamicMembership'))"
```

**After:**
```powershell
url = "/users/$($UserPrincipalName)/memberOf/microsoft.graph.group?`$select=displayName,groupTypes"
```

And updated `$currentMemberOf` population:

**Before:**
```powershell
[array]$currentMemberOf = ($batchResponse | Where-Object -FilterScript { $_.id -eq 'MemberOf' }).body.value.DisplayName
```

**After:**
```powershell
[array]$currentMemberOf = ($batchResponse | Where-Object -FilterScript { $_.id -eq 'MemberOf' }).body.value |
    Where-Object { $_.groupTypes -notcontains 'DynamicMembership' } |
    Select-Object -ExpandProperty DisplayName
```

---

### `Set-TargetResource` — Null guard before `Remove-MgGroupMemberDirectoryObjectByRef`

Added a `$null` check before calling `Remove-MgGroupMemberDirectoryObjectByRef` as a defensive safety net. If for any reason a non-group object slips through, the code now logs a meaningful message and skips it rather than crashing.

`throw` was intentionally avoided here — throwing inside `ForEach-Object` would abort processing of all remaining group changes for the user.

**Before:**
```powershell
else
{
    Remove-MgGroupMemberDirectoryObjectByRef -GroupId $group.Id -DirectoryObjectId $userId
}
```

**After:**
```powershell
else
{
    if ($null -ne $group)
    {
        Remove-MgGroupMemberDirectoryObjectByRef -GroupId $group.Id -DirectoryObjectId $userId
    }
    else
    {
        Write-Verbose -Message "Skipping removal of '$($_.InputObject)': not found as a group object (may be an Administrative Unit)."
        New-M365DSCLogEntry -Message 'Error updating data:' `
            -Exception "Skipping non-group object during member removal: '$($_.InputObject)'" `
            -TenantId $TenantId `
            -Credential $Credential
    }
}
```

---

## Testing

- [ ] User who is a member of an **Administrative Unit** — no longer causes an error
- [ ] User who is a member of **dynamic groups** — still correctly excluded
- [ ] User who needs to be **added to a static group** — still works as expected
- [ ] User who needs to be **removed from a static group** — still works as expected
- [ ] User with **no group memberships** — still works as expected

---

## Related

Closes #7198


saeekailas:
## Description

Fixes #7207

This PR resolves two bugs in `MSFT_IntuneSettingCatalogCustomPolicyWindows10.psm1` related to empty and whitespace string values in `MSFT_MicrosoftGraphDeviceManagementConfigurationSimpleSettingValue`.

---

## Problems Fixed

### Bug 1 — Export (`Get-SettingValue`)
- Empty strings `""` were silently dropped from the export (no `StringValue` written)
- Whitespace strings `"   "` were incorrectly cast to `IntValue = 0` due to `[UInt32]` coercion
- Root cause: `IsNullOrEmpty` check excluded empty strings, and whitespace passed through to the integer cast path

### Bug 2 — Apply (`Set-TargetResource`)
- Configs exported with `IntValue = 0` + `odataType = StringSettingValue` failed on apply with `ModelValidationFailure`
- The Intune Graph API rejected integer `0` for a field typed as `Edm.String`
- Root cause: No validation that the value type matched the declared `odataType` before sending to the API

---

## Changes Made

### 1. `Get-SettingValue` — `*SimpleSettingInstance` block
Replaced `IsNullOrEmpty` with `$null -ne` and added `IsNullOrWhiteSpace` guard to correctly route empty/whitespace values to `StringValue` instead of attempting an integer cast.

### 2. `Set-TargetResource` — Before `CreateParameters`/`UpdateParameters` clone
Added a repair loop that detects and corrects any `SimpleSettingValue` where `odataType` is `StringSettingValue` but `IntValue` is incorrectly set. Also handles nested settings inside `ChoiceSettingValue.Children`.

---

## Testing

- [x] Exported policy with empty string `""` — now exports as `StringValue = ""`
- [x] Exported policy with whitespace `"   "` — now exports as `StringValue = "   "`
- [x] Applied policy with empty string — no longer throws `ModelValidationFailure`
- [x] Applied policy with whitespace — no longer throws `ModelValidationFailure`
- [x] Normal integer and string values unaffected

---

## Files Changed
- `Modules/Microsoft365DSC/DSCResources/MSFT_IntuneSettingCatalogCustomPolicyWindows10/MSFT_IntuneSettingCatalogCustomPolicyWindows10.psm1`

PR #7214, opened 2026-06-08
AADAdministrativeUnit: Fixed validation of scoped role members' uniqueness

salbeck-sit:
The resource attempts to invalidate non-unique scoped-role members but fails to discern between arrays and hashtables

After introduction of 'the Graph shim' it appears that certain calls return hashtables instead of objects.
When checking for number of instances returned, the variable-type must be taken into account in order to discern arrays from hashtables (a single hash-table with two keys will have Count=2)

#### Task list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+* AADAdministrativeUnit
+  * Fixed validation of scoped role members' uniqueness
 * AADGroup
   * Various timing-related fixes for new group
 * AADEntitlementManagementAccessPackageAssignmentPolicy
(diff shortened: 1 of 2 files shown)


FabienTschanz:
I think it would be more appropriate to have $roleMemberIdentity defined as an array instead of relying on the type check. Using a filter may return an array, so we should treat it as such.

salbeck-sit:
Another option is for the Graph shim-module to get Invoke-MgGraphRequest to return objects as PSCustomObjects instead of hashtables - just like the original cmdlets. Given that this may cause unintended consequences, I've added code to resolve the issue as you suggested.

FabienTschanz (approved):
LGTM.

PR #7217, opened 2026-06-09
AADGroupsSettings: Fix error introduced by recent update

salbeck-sit:
AADGroupsSettings updated - when the DirectorySettings object doesn't exist in a new tenant, the resource throws an exception because the policy-id is blank. This is now fixed by using the policy-id of a newly created directorysetting object *or* use the policy-id of an existing directorysetting object.
There's a conflict caused by updates to the changelog which I've attempted to include in this PR

None

#### Task list

### DIFF
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADGroupsSettings/MSFT_AADGroupsSettings.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADGroupsSettings/MSFT_AADGroupsSettings.psm1
@@ -293,11 +293,10 @@ function Set-TargetResource
         $Policy = New-MgBetaDirectorySetting -TemplateId '62375ab9-6b52-47ed-826b-58e47e0e304b' | Out-Null
         $needToUpdate = $true
     }
-    elseif ($currentPolicy.Ensure -eq 'Present')
-    {
-        $Policy = Get-MgBetaDirectorySetting -All | Where-Object -FilterScript { $_.DisplayName -eq 'Group.Unified' }
-    }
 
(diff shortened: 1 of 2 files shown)


salbeck-sit:
fixed changelog

Modules/Microsoft365DSC/DSCResources/MSFT_AADGroupsSettings/MSFT_AADGroupsSettings.psm1
```diff
@@ -293,10 +293,11 @@ function Set-TargetResource
         $Policy = New-MgBetaDirectorySetting -TemplateId '62375ab9-6b52-47ed-826b-58e47e0e304b' | Out-Null
         $needToUpdate = $true
     }
+    elseif ($currentPolicy.Ensure -eq 'Present')
+    {
+        $Policy = Get-MgBetaDirectorySetting -All | Where-Object -FilterScript { $_.DisplayName -eq 'Group.Unified' }
```
FabienTschanz, line 298: Can we use the -Filter parameter instead of local filtering? This would speed up processing on the server side and reduce the number of items to be enumerated (if supported by the endpoint).

salbeck-sit:
I have to admit that I haven't tried it recently but all available documentation says that the -Filter parameter is not supported by the API. I'll see if I can find something new.

FabienTschanz:
Appreciate it, thanks. I can take a look at it earliest on the weekend because I‘m currently absent without access to a computer.

salbeck-sit:
Sorry, serverside filtering is still not available.
image

FabienTschanz (approved):

PR #7219, opened 2026-06-10
SPOSharingSettings: replace O(n) site enumeration with O(1) My Site Host lookup

IMJLA:
# PR: perf(SPOSharingSettings) — resolve My Site Host directly instead of enumerating all sites

## Problem

`MSFT_SPOSharingSettings.Get-TargetResource` reads the My Site Host's
`SharingCapability` by enumerating **every site collection in the tenant**:

```powershell
# Local filtering because server side filtering intermittently fails
$MySite = Get-PnPTenantSite -Filter "Url -like '-my.sharepoint.'" |
    Where-Object -FilterScript { $_.Template -match '^SPSMSITEHOST#' }
```

`Get-PnPTenantSite` pages the tenant site index and the `-Filter` is applied
client-side via `Where-Object`. Cost is linear in the number of site
collections, which grows with org headcount. Because `Test-TargetResource`
calls `Get-TargetResource` on every consistency check, apply/monitor time
creeps up over time on large tenants.

The My Site Host is a singleton at a deterministic URL, so the enumeration is
unnecessary.

## Fix

Derive the My Site Host host from the active connection URL and resolve it with
a single `Get-PnPTenantSite -Identity` (O(1)). The derivation:

- Works whether `(Get-PnPConnection).Url` is the **root** URL (app / certificate
  / managed-identity auth — what unattended DSC uses) or the **-admin** URL
  (credential auth): it strips a trailing `-admin` from the tenant label and
  inserts `-my`.
- Preserves the SPO domain suffix verbatim, so it is correct across Commercial,
  GCC, GCC High (`sharepoint.us` / `sharepoint-mil.us`) and 21Vianet
  (`sharepoint.cn`) without a hardcoded TLD list.

If the URL can't be resolved or the lookup returns nothing, it falls back to the
original enumeration (preserved verbatim), so behavior is unchanged in edge
cases. Also removes the redundant second `Get-PnPTenantSite` call the original
made on the optimized path.

## Benchmark (real tenants)

The fix was validated on two production tenants of very different sizes, which
together show the cost is linear in site-collection count:

| Tenant size (site collections) | Legacy enumeration | Optimized direct lookup | Speedup |
|-------------------------------:|-------------------:|------------------------:|--------:|
| ~829                           | 6.8 s              | 1.8 s                   | 3.8x    |
| **~133,907**                   | **953.5 s (~15.9 min)** | **1.3 s**          | **738x** |

On the large tenant a single `Get-TargetResource` spends **~16 minutes** in the
enumeration — and `Test-TargetResource` calls it on every consistency check, so
the cost recurs and grows with tenant size. The optimized direct lookup is
effectively constant (~1-2 s) regardless of site count. In both runs the
optimized and legacy paths returned the identical `SharingCapability` and the
same My Site Host URL.

### Functional validation through the DSC engine

Beyond the raw timing harness, the fix was validated end-to-end through the
resource itself (Microsoft365DSC v1.26.603.1, Windows PowerShell 5.1) by driving
`Get` / `Test` / `Set` and toggling `MySiteSharingCapability` back and forth
(`ExistingExternalUserSharingOnly` <-> `ExternalUserSharingOnly`). Stock and
patched runs agreed on the baseline and **all 22 checks passed in both
directions**. Per-`Get` time through the engine:

```
STOCK   Get ms: 16867, 3251, 2964, 2995, 2958   (avg 5807)
PATCHED Get ms:  6433,  870,  884,  993,  868    (avg 2010)
```

~2.9x average (~3.4x on warm Gets; the first Get in each phase also includes
connection setup). This medium tenant shows a smaller multiple than the large
tenant above because the gain scales with site count — the point is the patched
path is flat regardless of tenant size, and the functional Get/Test/Set contract
is unchanged.

## Compatibility / risk

- **Output unchanged**: same `MySiteSharingCapability` value and type; only the
  lookup mechanism changes. Fallback is the original logic byte-for-byte.
- **No new parameters, no schema change, no new permissions** — a direct
  `-Identity` lookup needs no broader scope than the enumeration.
- **Cloud-agnostic** via suffix-preserving transform.

## Tests

- Mocks `Get-PnPConnection` returning the **root** URL (proves the optimized
  path fires under app-auth, not just credential auth).
- Mocks `Get-PnPTenantSite` to return the My Site Host on the `-Identity` lookup
  and `$null` otherwise.
- Asserts the optimized path is taken (1 `-Identity` call, 0 `-Filter` calls).
- Adds a fallback test (no resolvable URL → original enumeration runs).

## Files

- `Modules/Microsoft365DSC/DSCResources/MSFT_SPOSharingSettings/MSFT_SPOSharingSettings.psm1`
- `CHANGELOG.md`
- `Tests/Unit/Microsoft365DSC/Microsoft365DSC.SPOSharingSettings.Tests.ps1`

None

#### Task list


IMJLA:
**Pushed a fix for the failing `unit-test` job.**

The three new unit tests for the My Site Host lookup were passing locally but failing in CI. Root cause: this PR is the first usage of `Get-PnPConnection` anywhere in the codebase, and that cmdlet isn't present in `Tests/Unit/Stubs/Microsoft365.psm1`. Locally the real `PnP.PowerShell` module is loaded so Pester's `Mock -CommandName Get-PnPConnection` binds fine; in CI the tests run against the canned stub module, where the command doesn't exist — so the mock can't bind and the three tests fail.

Fix: added a `Get-PnPConnection` stub to the `PnP.PowerShell` region of `Microsoft365.psm1`, matching the existing single-line stub format (empty `param()`, since the real cmdlet takes no parameters). No resource logic changed; this is test scaffolding only, so no CHANGELOG entry was added.

The other PnP cmdlets this PR mocks (`Get-PnPTenant`, `Get-PnPTenantSite`) were already stubbed from prior usages, which is why only the `Get-PnPConnection`-dependent tests failed.

PR #7239, opened 2026-06-16
Fix #7235

### LINKED ISSUE #7235: AADServicePrincipal: Set-TargetResource sends null appRoleId when adding AppRoleAssignedTo because role id is looked up by PrincipalType
### Description of the issue

`AADServicePrincipal` fails in `Set-TargetResource` when adding an `AppRoleAssignedTo` assignment. The POST to create the assignment sends a null `appRoleId`, and Graph rejects it:

```
Request_BadRequest: A null value was found for the property named 'appRoleId', which has the
expected type 'Edm.Guid[Nullable=False]'. The expected type 'Edm.Guid[Nullable=False]' does not
allow null values.
```

Root cause is in both the create and update branches of `Set-TargetResource`, where the app role id is resolved by matching the role's `DisplayName` against the assignment's `PrincipalType`:

```powershell
$appRoleId = ($appInstance.AppRoles | Where-Object -FilterScript { $_.DisplayName -eq $assignment.PrincipalType }).Id
$bodyParam = @{
    principalId = $PrincipalIdValue
    resourceId  = $currentAADServicePrincipal.ObjectID
    appRoleId   = $appRoleId
}
New-MgServicePrincipalAppRoleAssignedTo -ServicePrincipalId ... -BodyParameter $bodyParam
```

`$assignment.PrincipalType` is `'User'`, `'Group'`, or `'ServicePrincipal'` — it is the type of the assigned principal, not the name of an app role. An application's `AppRoles` collection never contains a role whose `DisplayName` is `'User'` or `'Group'`, so the lookup returns nothing and `$appRoleId` is `$null`. The subsequent POST then fails because `appRoleId` is a required, non-nullable GUID.

This only manifests when an assignment must be **added** (the `membersToAdd` path). Service principals whose assignments already exist are unaffected, because the comparison yields no members to add and the broken line never executes. Any service principal that needs a new `AppRoleAssignedTo` assignment created will fail on every consistency cycle.

For assignments that are not tied to a specific app role (the common case of granting "default access" to a user or group), the assignment should be created with the all-zeros app role id `00000000-0000-0000-0000-000000000000`, which is the documented sentinel for "no specific app role." An earlier version of this resource used that sentinel directly; it appears to have been replaced with the `DisplayName -eq PrincipalType` lookup, which cannot succeed.

### Suggested fix

When the `DisplayName`-based lookup does not resolve a role id, fall back to the default-access app role id rather than sending null. Apply to both the create and update (`membersToAdd`) branches:

```powershell
$appRoleId = ($appInstance.AppRoles | Where-Object -FilterScript { $_.DisplayName -eq $assignment.PrincipalType }).Id
if ([string]::IsNullOrEmpty($appRoleId))
{
    $appRoleId = '00000000-0000-0000-0000-000000000000'
}
$bodyParam = @{
    principalId = $PrincipalIdValue
    resourceId  = $currentAADServicePrincipal.ObjectID
    appRoleId   = $appRoleId
}
```

### Microsoft 365 DSC Version

1.26.527.1

### The DSC configuration

```powershell
Configuration Example
{

    node localhost
    {
        AADServicePrincipal 'ExampleSP'
        {
            AppId                     = 'Example Application'
            DisplayName               = 'Example Application'
            AccountEnabled            = $true
            ServicePrincipalType      = 'Application'
            AppRoleAssignmentRequired = $true
            AppRoleAssignedTo         = @(
                MSFT_AADServicePrincipalRoleAssignment {
                    PrincipalType = 'Group'
                    Identity      = 'example-group'
                }
            )
            Ensure                = 'Present'
            ApplicationId         = '00000000-0000-0000-0000-000000000000'
            TenantId              = 'contoso.onmicrosoft.com'
            CertificateThumbprint = '0000000000000000000000000000000000000000'
        }
    }
}
```

```powershell
[[AADServicePrincipal]ExampleSP] Updating existing Service Principal
[[AADServicePrincipal]ExampleSP] Need to update AppRoleAssignedTo value
[[AADServicePrincipal]ExampleSP] Retrieving group {example-group}
[[AADServicePrincipal]ExampleSP] Adding member {example-group}

##[error]PowerShell DSC resource MSFT_AADServicePrincipal failed to execute Set-TargetResource
functionality with error message:
{"error":{"code":"Request_BadRequest","message":"A null value was found for the property named
'appRoleId', which has the expected type 'Edm.Guid[Nullable=False]'. The expected type
'Edm.Guid[Nullable=False]' does not allow null values."}}


The body sent to `New-MgServicePrincipalAppRoleAssignedTo` contained `"appRoleId": null`.
```

```powershell

```

NikCharlebois:
Fixed an issue where adding `AppRoleAssignedTo` assignments could send a null app role id.

* Fixes#7235

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@
 * AADServicePrincipal
   * Fixed and issue where Custom SecurityAttributes were not handled properly.
     FIXES [#7192](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7192)
+  * Fixed an issue where adding `AppRoleAssignedTo` assignments could send a null app role id.
+    FIXES [#7235](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7235)
 * AADTenantAppManagementPolicy
   * Fixed an issue when updating the resource.
     FIXES [#7201](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7201)
--- a/Modules/Microsoft365DSC/DSCResources/MSFT_AADServicePrincipal/MSFT_AADServicePrincipal.psm1
+++ b/Modules/Microsoft365DSC/DSCResources/MSFT_AADServicePrincipal/MSFT_AADServicePrincipal.psm1
@@ -693,7 +693,7 @@ function Set-TargetResource
                     $PrincipalIdValue = $group.Id
                 }
 
-                $appRoleId = ($newSP.AppRoles | Where-Object -FilterScript { $_.DisplayName -eq $assignment.PrincipalType }).Id
+                $appRoleId = Get-M365DSCAADServicePrincipalAppRoleId -AppRoles $newSP.AppRoles -PrincipalType $assignment.PrincipalType
                 $bodyParam = @{
                     principalId = $PrincipalIdValue
                     resourceId  = $newSP.Id
@@ -796,7 +796,7 @@ function Set-TargetResource
                             $PrincipalIdValue = $group.Id
                         }
 
-                        $appRoleId = ($appInstance.AppRoles | Where-Object -FilterScript { $_.DisplayName -eq $assignment.PrincipalType }).Id
+                        $appRoleId = Get-M365DSCAADServicePrincipalAppRoleId -AppRoles $appInstance.AppRoles -PrincipalType $assignment.PrincipalType
                         $bodyParam = @{
                             principalId = $PrincipalIdValue
                             resourceId  = $currentAADServicePrincipal.ObjectID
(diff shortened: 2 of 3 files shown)


FabienTschanz:
Wouldn‘t it be more concise if we inlined the function into those two instances where it is used instead of creating a function? If it were used more than two times yes, but for only two instances, I feel like this is a bit overkill.

NikCharlebois:
We could. I am just thinking of a scenario where we may have Agentic Identities coming down soon and may have some additional conditions to look for, it may prevent us from missing one of the 2 instances down the line.

FabienTschanz:
Got it. I‘m not familiar with Agentic Identities, can‘t say anything on that part. If there is potentially changes coming soon, we can keep it.

PR #7244, opened 2026-06-17
IntuneDeviceConfigurationPlatformScript{MacOS,Windows}: Id parameter should not be mandatory

### LINKED ISSUE #7243: IntuneDeviceConfigurationPlatformScriptWindows requires "Id" parameter
### Description of the issue

The Get-TargetResource, Set-TargetResource and Test-TargetResource functions all require that the "Id" parameter to be provided (mandatory).  This property should not be mandatory as the resource does not have an "Id" value if it does not exist.

### Microsoft 365 DSC Version

DEV

Intune

### The DSC configuration

```powershell
All of them...
```

```powershell

```

```powershell

```

ricmestre:

This PR fixes an issue where both `IntuneDeviceConfigurationPlatformScriptMacOS` and `IntuneDeviceConfigurationPlatformScriptWindows` have the `Id` parameter marked as mandatory, instead only `DisplayName` should be mandatory in this case.

- Fixes #7243 

#### Task list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 
 * IntuneDeviceConfigurationPlatformScriptMacOS
   * Removed `Id` as being a mandatory parameter
+    FIXES [#7243](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7243)
 * IntuneDeviceConfigurationPlatformScriptWindows
   * Removed `Id` as being a mandatory parameter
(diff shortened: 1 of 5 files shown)


FabienTschanz:
We need to gate the lookup by Id if it is actually provided like in the other Intune resources too. Otherwise it will always throw there (and continue, but we want to avoid those errors).

FabienTschanz (approved):
LGTM.

PR #7262, opened 2026-06-29
Initial Release - AADApplicationFederatedIdentityCredential

### LINKED ISSUE #7252: AADApplicationFederatedIdentityCredential: New resource proposal
### Description

Add the `AADApplicationFederatedIdentityCredential` resource to manage an application's federated identity credentials, supporting the full lifecycle - `Get` / `New` / `Update` / `Remove-MgApplicationFederatedIdentityCredential` (Microsoft Graph v1.0). A federated identity credential is a trust configured on the app registration that lets a token from an external OIDC issuer - matching the credential's issuer, subject, and audience - be exchanged for a Microsoft Entra token for that app, instead of using a client secret or certificate.

### Proposed properties

```powershell
AADApplicationFederatedIdentityCredential "AWSProdWorkload"
{
    AppId       = "MyWorkloadApp"
    Name        = "AWS-123456789012-Prod"
    Issuer      = "https://<account-uuid>.tokens.sts.global.api.aws"
    Subject     = "arn:aws:iam::123456789012:role/EntraTokenExchangeRole"
    Audiences   = @("api://AzureADTokenExchange")
    Description = "AWS workload identity federation"
    Ensure      = "Present"
}
```

NikCharlebois:
* Initial release of AADApplicationFederatedIdentityCredential

* Fixes #7252


Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,574 @@
...
+        -Parameters $PSBoundParameters
+    Add-M365DSCTelemetryEvent -Data $data
+    #endregion
+
+    $currentFederatedIdentityCredential = Get-TargetResource @PSBoundParameters
+    $bodyParameter = Get-M365DSCFederatedIdentityCredentialBody -Name $Name `
+        -Issuer $Issuer `
+        -Subject $Subject `
+        -Audiences $Audiences `
+        -Description $Description
```
FabienTschanz: Now that should be inlined. The function is called once and only in this place.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,574 @@
...
+
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [System.String]
+        $ApplicationObjectId,
```
FabienTschanz: Regarding multi-tenancy I suggest to either add or use the display name for an application as well.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/settings.json
```diff
@@ -0,0 +1,61 @@
...
+    }
+  },
+  "requiredModules": [
+    "Microsoft.Graph.Applications"
+  ],
+  "supportedEnvironments": [
+    "Global",
+    "USGov"
+  ],
+  "mode": "Configuration",
```
FabienTschanz, line 48: I suggest classifying this as a Data plane object instead of configuration because we enumerate all applications. Depending on the number of applications in a tenant, getting all of the credentials from the apps may take a long time.

NikCharlebois: Control Plan means that only Admins, through an admin portal can edit the settings, which is the case here. Data plane is only reserved for items that are in the grey zone between what a end-user and an admin can do. Best example is Teams Channels. while admins can manage those in the Teams Admin Center, End-users can control also those in the Teams client directly. Therefore this falls in the Data plane. Control Plane is correct for this resource.

FabienTschanz: But we also have some resources that are in the Data plane because they take an eternity to export and/or manage. Even just an export of the resource for all apps may take hours to complete. While I agree that under normal terms it is a configuration object, I'd still classify this as Data.

NikCharlebois: They shouldn't be in the data plane in this case. We addressed the long running resources a long time ago by introducing export modes Default, Lite and Full

FabienTschanz: Lite doesn't exist anymore, Default maps to Configuration and Full to Configuration + Data. But that's another discussion.

ricmestre: @NikCharlebois Hi Nik, currently AADApplication already retrieves certificates and client secrets, is it really necessary to have a separate resource for federated credentials instead of just adding this code to AADApplication? This way with a separate resource if you need to export applications it will need to go through all of them again a second time to export the credentials, additionally it also creates a discrepancy which is that applications are currently set as data plane so if you set this as control plane and I want to export the AAD workload (only control plane) it will still go through all applications even if I don't want to just to export the federated credentials.

NikCharlebois: Moving forward, we should try to create wrappers around APIs instead of trying to agglomerate multiple API endpoints into macro resources like AADApplication and AADConditionalAccessPolicy. If we want a chance at having support for all APIs and automating the process as much as possible, we should try and focus on that approach.

iamsandipshaw06:
Thanks @NikCharlebois for picking this up. This implements the lifecycle I was proposing in #7252. Three questions from reviewing the current head:

  1. `settings.json` lists four federated identity credential cmdlets that are missing from the Graph Shim used by the module loader. Should the shim be regeneratedin this PR?

  2. Could updates use a separate body containing only the properties that were provided? The current shared body includes `name` and turns omitted `issuer`, `subject`, and `audiences` into blank/null values instead of omitting them.

  3. Duplicate application display names remain an export edge case: lookup throws on multiple matches, but export removes `ApplicationObjectId`. Should exports retain it as a disambiguation hint, or should this limitation be documented?

Happy to validate the full lifecycle against an AWS OIDC -> Entra federation in a test tenant if useful.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,620 @@
...
+            }
+
+            if ($null -eq $application)
+            {
+                try
+                {
+                    [array]$application = Get-MgApplication `
+                        -Filter "DisplayName eq '$($ApplicationDisplayName -replace "'", "''")'" `
+                        -Property @('id', 'displayName') `
+                        -ErrorAction SilentlyContinue
```
FabienTschanz: Will `-ErrorAction SilentlyContinue` not swallow every exception being thrown, so the catch block is never executed?

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,620 @@
...
+            }
+
+            if ($null -eq $federatedIdentityCredential)
+            {
+                try
+                {
+                    [array]$federatedIdentityCredential = Get-MgApplicationFederatedIdentityCredential `
+                        -ApplicationId $ApplicationObjectId `
+                        -Filter "name eq '$($Name -replace "'", "''")'" `
+                        -ErrorAction SilentlyContinue
```
FabienTschanz: Same here with swallowing the exception.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,620 @@
...
+
+            $federatedIdentityCredential = $null
+            try
+            {
+                if (-not [System.String]::IsNullOrEmpty($Id))
+                {
+                    $federatedIdentityCredential = Get-MgApplicationFederatedIdentityCredential `
+                        -ApplicationId $ApplicationObjectId `
+                        -FederatedIdentityCredentialId $Id `
+                        -ErrorAction SilentlyContinue
```
FabienTschanz: Same here with swallowing the exception.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,620 @@
...
+
+            $application = $null
+            try
+            {
+                if (-not [System.String]::IsNullOrEmpty($ApplicationObjectId))
+                {
+                    $application = Get-MgApplication `
+                        -ApplicationId $ApplicationObjectId `
+                        -Property @('id', 'displayName') `
+                        -ErrorAction SilentlyContinue
```
FabienTschanz: Same here with swallowing the exception.

Modules/Microsoft365DSC/DSCResources/MSFT_AADApplicationFederatedIdentityCredential/MSFT_AADApplicationFederatedIdentityCredential.psm1
```diff
@@ -0,0 +1,620 @@
...
+                    CertificatePassword   = $CertificatePassword
+                    ManagedIdentity       = $ManagedIdentity.IsPresent
+                    ApplicationDisplayName = $application.DisplayName
+                    ApplicationObjectId   = $application.Id
+                    Name                  = $federatedIdentityCredential.Name
+                    Id                    = $federatedIdentityCredential.Id
+                    AccessTokens          = $AccessTokens
+                }
+                $Script:exportedInstance = $federatedIdentityCredential | Add-Member -MemberType NoteProperty -Name ApplicationObjectId -Value $application.Id -PassThru
+                $Script:exportedInstance = $Script:exportedInstance | Add-Member -MemberType NoteProperty -Name ApplicationDisplayName -Value $application.DisplayName -PassThru
```
FabienTschanz: `Add-Member` is a cmdlet that operates on PSObjects. Adding a note property to a hashtable doesn't make it splattable against a function, meaning that `ApplicationObjectId` and `ApplicationDisplayName` won't be available to the `Get-TargetResource` function. 

I'd suggest either adding them directly using the dot-Notation or via `$params.Add('prop', 'value')`, which is the more common way.

PR #7275, opened 2026-06-30
Fix AADGroup empty member enumeration

### LINKED ISSUE #7274: [AADGroup] Empty "Members = @()" triggers a full membership enumeration in Get-TargetResource - very slow / timing-out Test on large groups (regression in 1.26.617.1)
### Description of the issue

**Summary**

Starting in **1.26.617.1**, `MSFT_AADGroup`'s `Get-TargetResource` performs a **full enumeration of a group's membership for non-dynamic groups with 20+ members** when the configuration declares an **empty** `Members = @()` (or `GroupAsMembers = @()`). For groups with a very large membership (hundreds of thousands of members) this issues hundreds of paged Graph calls and takes a very long time, making `Test-DscConfiguration` (drift detection) extremely slow or effectively hang.

This is a regression from **1.26.527.1**, where an empty `Members` array did **not** trigger the additional full member enumeration during `Get`/`Test`.

> Note: this is specific to `Get-TargetResource` / `Test-DscConfiguration`. `Export-TargetResource` already passed a non-empty `Members = @('toextract')` placeholder in earlier versions, so the removed empty-array guard is **not** what changed export behavior. (On certain advanced filter paths, export also sets an internal fetch flag.)

**Root cause**

The guard that decides whether to run the *separate full member enumeration* in `Get-TargetResource` changed — the `-and $Members.Count -gt 0` (and the `GroupAsMembers` equivalent) check was removed:

1.26.527.1 (no full enumeration for empty Members):
```powershell
if ($Group.MembershipRuleProcessingState -ne 'On' -and `
    (($PSBoundParameters.ContainsKey('Members')        -and $Members.Count -gt 0) -or `
     ($PSBoundParameters.ContainsKey('GroupAsMembers')  -and $GroupAsMembers.Count -gt 0)))
{
    $groupMembers = $Group.Members
    if ($Group.Members.Count -eq 20 -or $Script:requireGroupMemberFetching -eq $true)
    {
        $groupMembers = Get-MgGroupMember -GroupId $Group.Id -All -Top 999
    }
    ...
}
```

1.26.617.1 (enumerates even when Members is empty):
```powershell
if ($Group.MembershipRuleProcessingState -ne 'On' -and `
    ($PSBoundParameters.ContainsKey('Members') -or `
     $PSBoundParameters.ContainsKey('GroupAsMembers')))
{
    $groupMembers = $Group.Members
    if ($Group.Members.Count -eq 20 -or $Script:requireGroupMemberFetching -eq $true)
    {
        $groupMembers = Get-MgGroupMember -GroupId $Group.Id -All -Top 999
    }
    ...
}
```

The group is always loaded with `-ExpandProperty 'members'` (an inline preview of up to 20 members) — that is unchanged. The removed guard only controlled the *separate* `Get-MgGroupMember -All` call. With the outer condition now just `ContainsKey('Members')`, declaring the property — even as an empty array — enters the fetch block; and because the inner check is `$Group.Members.Count -eq 20` (the inline preview caps at 20), **any non-dynamic group with ≥ 20 members** declared with empty `Members` runs the full enumeration during `Get`/`Test`.

**Why it is expensive (it is not capped at 999)**

`Get-MgGroupMember -All -Top 999` does not return at most 999 here. In M365DSC this call is served by `M365DSCGraphShim`, where `-Top 999` is the **page size** (`$top=999`), **not** a total cap, and `-All` follows **every** `@odata.nextLink` (`Get-M365DSCGraphShimAllPages` loops `Invoke-MgGraphRequest` until there is no next page). So the cost scales with the **full group size** — roughly N/999 sequential requests. For hundreds of thousands of members that is hundreds of paged calls and many minutes to hours.

**Impact**

Any configuration that declares `Members = @()` or `GroupAsMembers = @()` (a common way to assert a group should have no members) now pays a full-membership enumeration on every `Get-TargetResource` / `Test-DscConfiguration` for such groups (non-dynamic, ≥ 20 actual members), even though detecting drift against an empty desired set does not require enumerating the whole group. On large groups this turns a near-instant evaluation into one that can run for hours or time out.

**How to reproduce**

1. A non-dynamic group with a large actual membership (hundreds of thousands of members).
2. A config declaring it with empty `Members = @()` (see the *DSC configuration* field below).
3. Run `Test-DscConfiguration` (or call `Get-TargetResource` directly with `Members = @()`).

- **Observed:** repeated paged requests `GET /v1.0/groups/<id>/members?$top=999&$skiptoken=...` enumerating the entire group; evaluation runs for a very long time (see the *Verbose logs* field below).
- **Expected (and behavior in 1.26.527.1):** an empty desired `Members` does not trigger the full membership enumeration during `Get`/`Test`.

**Suggested fix**

The core problem is the unconditional full enumeration (`Get-MgGroupMember -All`) for the empty-`Members` case:

1. **Drift detection (Test) does not need it.** Determining that an empty desired `Members = @()` is drifted only requires knowing whether the group currently has *any* members — a bounded/count check (a single page, or `$count`) is sufficient.
2. **Reserve the full `-All` enumeration for when member identities are actually required** — i.e., `Set`/remediation (to remove existing members), or when `Get` must return the current member list.

If preserving the empty-membership-enforcement behavior is **not** required, the simplest fix is to restore the original guard so an empty desired `Members`/`GroupAsMembers` skips the fetch entirely:
```powershell
(($PSBoundParameters.ContainsKey('Members')       -and $Members.Count -gt 0) -or
 ($PSBoundParameters.ContainsKey('GroupAsMembers') -and $GroupAsMembers.Count -gt 0))
```

### Microsoft 365 DSC Version

1.26.617.1 (regression vs. 1.26.527.1)

### The DSC configuration

```powershell
Configuration Example
{
    node localhost
    {
        AADGroup 'LargeGroupEmptyMembers'
        {
            DisplayName     = 'Example Large Group'
            MailNickname    = 'examplelargegroup'
            SecurityEnabled = $true
            MailEnabled     = $false
            GroupTypes      = @()
            Members         = @()        # empty desired membership; the group itself has 100,000s of actual members
            Ensure          = 'Present'
            Credential      = $Credentials
        }
    }
}
```

```powershell

```

```powershell

```

NikCharlebois:
Avoids fetching every group member in `AADGroup` `Get-TargetResource` when `Members` or `GroupAsMembers` are explicitly declared as empty arrays. This restores the previous guard so `Test-DscConfiguration` does not trigger expensive full membership enumeration for large groups when the desired member list is empty.

Fixes #7274

#### Task list
- [x] Added an entry to the change log under the Unreleased section of the file CHANGELOG.md. Entry should say what was changed and how that affects users (if applicable), and reference the issue being resolved (if applicable).


FabienTschanz:
There was a reason behind the member evaluation. If you specify a configuration with Members = @(), and the group in the tenant contains members, then it won‘t record it as a drift. I get that the current behavior has a downside, but reverting the condition surfaces the previous issue again. This needs to be solved differently.

FabienTschanz:
I think the best way to fix this is to update the condition on line 244 of `MSFT_AADGroup.psm1` from this: 

```powershell
            if ($Group.Members.Count -eq 20 -or $Script:requireGroupMemberFetching -eq $true)
            {
                # Fetch all group members
                $groupMembers = Get-MgGroupMember -GroupId $Group.Id -All -Top 999
            }
```

to this:

```powershell
            if ((($Members.Count -gt 0 -or $GroupAsMembers.Count -gt 0) -and $Group.Members.Count -eq 20) -or $Script:requireGroupMemberFetching -eq $true)
            {
                # Fetch all group members
                $groupMembers = Get-MgGroupMember -GroupId $Group.Id -All -Top 999
            }
```

This check makes sure that we fetch all of the members if and only if the supplied DSC configuration contains at least one member in either Members or GroupAsMembers and when the member count of the group is 20. If you test with an empty Members or GroupAsMembers property, it will skip fetching all of the members, despite the group having more than 20 members.

FabienTschanz (approved):
LGTM

PR #7337, opened 2026-07-15
fixed invalid arg when run in WPS 5.1

salbeck-sit:
A recent addition includes code that uses Import-Module with parameter -SkipEditionCheck. However, this is only valid in Powershell 7, so code is added to ensure the parameter is only used with PS7

#### Task list


FabienTschanz:
I think we can even remove the `-SkipEditionCheck` parameter. That was my bad, forgot that it isn't available in WinPS. Doesn't really serve a purpose for that particular import statement here. Thanks!

FabienTschanz (approved):

PR #7387, opened 2026-07-31
Fix access package resource role scope churn when OriginId is a GUID

### LINKED ISSUE #7386: AADEntitlementManagementAccessPackage / CatalogResource: Get-TargetResource returns a group/app OriginId as its DisplayName, so Test never matches a GUID-based config
### Environment
- Microsoft365DSC 1.26.729.2 (also happens on 1.26.722.1)
- Windows PowerShell 5.1
- App-only auth (workload identity), Graph scopes EntitlementManagement.ReadWrite.All, Group.Read.All, Application.Read.All
- Resources: AADEntitlementManagementAccessPackage and AADEntitlementManagementAccessPackageCatalogResource

### What happens
If an access package role scope points at an AadGroup or AadApplication using the resource's GUID originId, Test-TargetResource always comes back as "not in desired state", and every Set removes and re-adds the role scope. After a couple of runs the package ends up with no resource roles at all ("this access package does not have any resources").

### Why
Get and Set disagree on what OriginId is.

Get (MSFT_AADEntitlementManagementAccessPackage, around lines 159-171) takes the GUID that Graph returns and swaps it for the display name:

    if ([System.Guid]::TryParse($originId, [ref]$guid)) {
        switch ($originSystem) {
            'AadApplication' { $originId = (Get-MgServicePrincipal -ServicePrincipalId $originId).DisplayName }
            'AadGroup'       { $originId = (Get-MgGroup -GroupId $originId).DisplayName }
        }
    }
    # AccessPackageResourceOriginId = $originId   (now the display name)

Set/New (around lines 403-428) does the opposite. It only resolves a display name when the value isn't already a GUID, and otherwise queries the catalog directly with originId eq '<guid>':

    if ($originSystem -in @('AadApplication','AadGroup') -and -not [System.Guid]::TryParse($originId, [ref]$guid)) {
        # resolve display name -> .Id
    }
    Get-MgBeta...AccessPackageResource -Filter "originId eq '$originId'"

So Set is fine with a GUID, but Get gives back a display name. The update logic (around lines 579-585) then compares the desired GUID against the current display name, never matches, and re-adds the scope on every run. The remove loop drops the existing (display-name) scope because it isn't in the GUID list, which is what empties the package.

The catalog resource (MSFT_AADEntitlementManagementAccessPackageCatalogResource, around lines 189-192) has the same GUID-to-display-name swap in its Get, so a GUID-based catalog resource never converges either and gets re-Set every run.

### Expected
Get should return OriginId the same way Set accepts it and the same way Graph stores it, i.e. the GUID. Set already handles GUIDs fine, and the originId Graph keeps for an AadGroup/AadApplication is the object GUID.

### Possible fix
Drop the GUID-to-display-name conversion in Get for both resources and just return originId. If you want the display name in the output for readability, the Test/update comparison needs to resolve both sides to the same form so that a GUID config and a display-name config both settle.

SebastianClaesson:
Fixes #7386

### Problem
Get-TargetResource resolves an AadGroup/AadApplication resource role scope OriginId to the object's display name. Set-TargetResource then compared the desired OriginId (commonly the raw GUID) against that display name, so a scope specified by GUID looked absent on every run and was removed and re-added on each Set. In practice this could leave the access package with no resource roles at all.

### Change
Added a small helper that resolves an OriginId to the same value Get-TargetResource returns (a GUID is resolved to the display name for AadGroup/AadApplication, any other value is returned unchanged). Set-TargetResource uses it before the add and remove comparisons, and Get-TargetResource now routes through the same helper so both paths stay consistent. Display-name-specified scopes behave exactly as before.

### Testing
Verified against a tenant: an access package with a role scope specified by GUID stays in desired state across repeated applies instead of losing its resource roles. Display-name-based configs are unaffected.

### Notes
AADEntitlementManagementAccessPackageCatalogResource also resolves OriginId to a display name in Get, which can show up as drift for GUID-based configs, though it does not have the destructive remove/re-add behaviour. Happy to address that separately if you'd like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 # 1.26.729.2
 
+* AADEntitlementManagementAccessPackage
+  * Fixed a bug where a resource role scope whose `AccessPackageResourceOriginId`
+    is specified as the object GUID was removed and re-added on every Set,
+    eventually leaving the access package with no resource roles. The desired
+    OriginId is now resolved to the same value that Get-TargetResource returns
+    before comparison.
+    FIXES [#7386](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7386)
 * M365DSCConnection
   * Fixed issue connecting to a workload using `New-M365DSCConnection`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,17 @@
 # Change log for Microsoft365DSC
 
-# 1.26.729.2
+# UNRELEASED
 
 * AADEntitlementManagementAccessPackage
-  * Fixed a bug where a resource role scope whose `AccessPackageResourceOriginId`
+  * Fixed an issue where a resource role scope whose `AccessPackageResourceOriginId`
     is specified as the object GUID was removed and re-added on every Set,
     eventually leaving the access package with no resource roles. The desired
     OriginId is now resolved to the same value that Get-TargetResource returns
     before comparison.
     FIXES [#7386](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7386)
+
+# 1.26.729.2
+
 * M365DSCConnection
   * Fixed issue connecting to a workload using `New-M365DSCConnection`.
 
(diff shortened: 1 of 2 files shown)


FabienTschanz (approved):

rajveersah:
fix access packages

PR #7397, opened 2026-08-11
Fix Intune export crash when tenant has no matching configuration policies

### LINKED ISSUE #7396: Intune export aborts with "Object reference not set to an instance of an object" when the tenant has no matching Settings Catalog configuration policies
### Description

Exporting the Intune workload fails immediately, before any resource is extracted, on a tenant that has no Settings Catalog configuration policies. The whole workload is lost — not one resource, all of them.

```
WARNING: Based on the provided Authentication parameters, the following resources cannot be extracted: IntuneDeviceEnrollmentScopeConfigurationMam,IntuneDeviceEnrollmentScopeConfigurationMdm
Partial Export file was saved at: C:\Users\...\....partial.ps1
Exception calling "Populate" with "2" argument(s): "Object reference not set to an instance of an object."
```

Nothing is written to the M365DSC error log, because the exception is thrown by the pre-pass in `Start-M365DSCConfigurationExtract` rather than by a resource.

### Root cause

`Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1:910`

```powershell
[array]$allRequestedConfigurationPolicies = Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds }
```

When the filtered pipeline emits nothing, this assigns **`$null`**, not an empty array. Casting an empty pipeline to `[array]` does not produce `@()`:

```powershell
PS> [array]$x = @() | Where-Object { $_ -eq 'nothing' }
PS> $null -eq $x
True
```

That `$null` is then passed to the cache at line 927:

```powershell
[Microsoft365DSC.Intune.ConfigurationPolicyCache]::Populate($allRequestedConfigurationPolicies, ...)
```

and `src/Microsoft365DSC.Intune/ConfigurationPolicyCache.cs` enumerates it without a null check:

```csharp
public static void Populate(IEnumerable<object> allPolicies, Func<object, string> templateIdSelector)
{
    lock (_lock)
    {
        if (_isPopulated)
            return;

        foreach (var policy in allPolicies)   // NullReferenceException when allPolicies is null
```

The exception propagates to the `catch` at line 1224, which saves the partial export and rethrows. `Pop-Location` at line 1213 is inside the `try`, so the caller's working directory is also left changed as a side effect.

### Reproduction

Any tenant whose Settings Catalog configuration policies do not match the requested template IDs will do — `GET /beta/deviceManagement/configurationPolicies` returning `0` is sufficient, which is a normal state for a new or small tenant.

```powershell
Export-M365DSCConfiguration -ApplicationId <id> -CertificateThumbprint <thumb> `
    -TenantId <tenant> -Path <path> -Workloads 'INTUNE' -Mode 'Full'
```

A tenant with zero policies is not strictly required: the trigger is only that the `Where-Object` filter produces no output, so requesting any component in `IntuneTemplateRegistry.json` whose `templateId` matches nothing reproduces it.

39 resources are in `IntuneTemplateRegistry.json`, so any export including one of them takes the affected path — this covers the security baselines, firewall, ASR, disk encryption, EDR and antivirus policies.

### Suggested fix

Guard the cache, which protects every caller:

```csharp
if (allPolicies == null)
{
    _isPopulated = true;
    return;
}
```

and optionally wrap the caller's pipeline in `@()` at `M365DSCReverse.psm1:910` so it yields an empty array.

Marking the cache populated is the correct semantic: consumers call `GetByTemplateId`, and `M365DSCIntuneUtil.psm1:528-533` already treats a `null` return as "query Graph directly", which is the right behaviour for a tenant with no policies.

### Version

- Microsoft365DSC 1.26.805.2
- PowerShell 7.6.3, Windows 11
- Authentication: service principal with certificate thumbprint

Confirmed still present on `Dev` at the time of writing (`M365DSCReverse.psm1:910` and `ConfigurationPolicyCache.cs` are unchanged).

I have a PR ready for this.

akabyshev:

Exporting the Intune workload aborts before any resource is extracted, with
`Exception calling "Populate" with "2" argument(s): "Object reference not set to an instance of an object."`,
on tenants that have no Settings Catalog configuration policies matching the requested template IDs.

`M365DSCReverse.psm1:910` casts a filtered pipeline to `[array]`. When that pipeline emits nothing the
result is `$null`, not an empty array:

```powershell
PS> [array]$x = @() | Where-Object { $_ -eq 'nothing' }
PS> $null -eq $x
True
```

The `$null` reaches `ConfigurationPolicyCache.Populate` at line 927, whose `foreach` dereferences it. As
this runs in the pre-pass of `Start-M365DSCConfigurationExtract`, the whole workload is lost rather than a
single resource, and nothing is written to the M365DSC error log.

Changes:

* `ConfigurationPolicyCache.Populate` now treats a null collection as a populated-but-empty cache. This is
  the correct semantic rather than merely defensive: `GetByTemplateId` returns `null` for a miss, and
  `M365DSCIntuneUtil.psm1:528-533` already falls back to a live filtered Graph query, which is the right
  result for a tenant with no policies.
* `M365DSCReverse.psm1:910` wraps the pipeline in `@()` so the caller produces an empty array. Belt and
  braces — either change alone fixes the crash.

Verified against the built assembly: `Populate($null, selector)` no longer throws, `GetByTemplateId`
returns `null` so consumers take the fallback path, and a non-null collection still caches as before.
Also verified end to end by exporting `IntuneFirewallPolicyWindows10`, `IntuneSecurityBaselineWindows10`
and `IntuneDeviceCategory` against a zero-policy tenant: 3/3 successful where the export previously died
before resource 1.

- Fixes #7396

#### Task list

### DIFF
--- a/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
@@ -907,9 +907,7 @@ function Start-M365DSCConfigurationExtract
                 ManagedIdentity = $ManagedIdentity
                 AccessTokens = $AccessTokens
             }
-            # @() is required: casting a pipeline that emitted nothing to [array] yields $null, not an
-            # empty array, which would then be passed to ConfigurationPolicyCache::Populate.
-            [array]$allRequestedConfigurationPolicies = @(Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds })
+            $allRequestedConfigurationPolicies = Get-M365DSCArrayFromProperty -PropertyValue (Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds })
             $batchRequests = @()
             foreach ($policy in $allRequestedConfigurationPolicies)
             {
--- a/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
+++ b/Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
@@ -907,7 +907,9 @@ function Start-M365DSCConfigurationExtract
                 ManagedIdentity = $ManagedIdentity
                 AccessTokens = $AccessTokens
             }
-            [array]$allRequestedConfigurationPolicies = Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds }
+            # @() is required: casting a pipeline that emitted nothing to [array] yields $null, not an
+            # empty array, which would then be passed to ConfigurationPolicyCache::Populate.
+            [array]$allRequestedConfigurationPolicies = @(Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds })
             $batchRequests = @()
             foreach ($policy in $allRequestedConfigurationPolicies)
             {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,7 @@
 * M365DSCReverse
   * Fixed an issue where exporting the Intune workload against a tenant with no
     matching Settings Catalog configuration policies aborted the entire export
-    with `Object reference not set to an instance of an object`. Casting an empty
-    pipeline to `[array]` yields `$null`, which was passed to
-    `ConfigurationPolicyCache.Populate`.
+    with `Object reference not set to an instance of an object`.
     FIXES [#7396](https://github.com/Microsoft365DSC/Microsoft365DSC/issues/7396)
 * M365DSCTenantInfo
   * Added support for Bleu and Delos sovereign cloud.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 # UNRELEASED
 
+* M365DSCReverse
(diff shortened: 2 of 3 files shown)


Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
```diff
@@ -907,7 +907,9 @@ function Start-M365DSCConfigurationExtract
                 ManagedIdentity = $ManagedIdentity
                 AccessTokens = $AccessTokens
             }
-            [array]$allRequestedConfigurationPolicies = Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds }
+            # @() is required: casting a pipeline that emitted nothing to [array] yields $null, not an
+            # empty array, which would then be passed to ConfigurationPolicyCache::Populate.
+            [array]$allRequestedConfigurationPolicies = @(Get-MgBetaDeviceManagementConfigurationPolicy -All | Where-Object { $_.templateReference.templateId -in $requestedConfigurationPolicyTemplateIds })
```
FabienTschanz: This can be resolved using the `Get-M365DSCArrayFromProperty` function. You feed it's `-PropertyValue` the policies, and it will create an array out of the provided value. This is cleaner than having to work with an array (which may contain a `$null` entry when you use it like that).

FabienTschanz (changes requested):

Modules/Microsoft365DSC/Modules/M365DSCReverse.psm1
akabyshev: Good call — updated to `Get-M365DSCArrayFromProperty`, thanks.

FabienTschanz (approved):

